PR #48, opened 2019-01-31
Fix indentation

vyasr:
This fixes indentation to match the updates to pycodestyle.


vyasr:
@simi2525 thanks for the suggestion!

Issue #53, opened 2019-02-08
The TORQUE scheduler is_present() method triggers subprocess.CalledProcessError


csadorf:
### Description

The TORQUE scheduler is_present() method forks to `qsub -V` to determine whether the scheduler is present. On some systems this fork will return with an exit status, because the current context does not allow for submission, e.g., on a compute node.

### To reproduce

For example, try to call `project.py status` on a Cray compute node.

csadorf:
Fixed by pull request #54.

bdice:
@csadorf Which system caused this issue? Is that system still in production? I ran into a problem with @klywang and we might want to revert PR #54 so that `is_present` returns `False` on systems where the scheduler version command fails (i.e. `bjobs -V` or `qsub --version` returning a non-zero exit code).

csadorf:
> @csadorf Which system caused this issue? Is that system still in production? @klywang ran into a problem and we might want to revert PR #54. (I'll let @klywang create a separate issue to document what's going wrong.)

This was probably on either titan or summit, but I do not remember anymore, sorry.

vyasr:
I'd assume from the combination of Torque and Cray that this was on Titan.

PR #54, opened 2019-02-08
Fix issue with TORQUE is_present() method triggering exception.

csadorf:
Fixes issue #53 .

Possible point of contention: Should we consider the scheduler to *not* be present in case that it triggers an error? I think we should say it's present even if it's not usable, because we might still be able to query the status of running jobs.

### DIFF
--- a/flow/scheduling/torque.py
+++ b/flow/scheduling/torque.py
@@ -142,6 +142,8 @@ class TorqueScheduler(Scheduler):
         "Return True if it appears that a TORQUE scheduler is available within the environment."
         try:
             subprocess.check_output(['qsub', '--version'], stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError:
+            return True
         except (IOError, OSError):
             return False
         else:


bdice:
Alternative approach: you could just use the Python equivalent of `which` instead of actually calling the command. It should work on Python 2.4 and newer. 😄

```python
    @classmethod
    def is_present(cls):
        "Return True if it appears that a TORQUE scheduler is available within the environment."
        import distutils.spawn
        return distutils.spawn.find_executable('qsub') is not None
```

csadorf:
> Alternative approach: you could just use the Python equivalent of `which` instead of actually calling the command. It should work on Python 2.4 and newer. 😄
> 
> ```python
>     @classmethod
>     def is_present(cls):
>         "Return True if it appears that a TORQUE scheduler is available within the environment."
>         import distutils.spawn
>         return distutils.spawn.find_executable('qsub') is not None
> ```

We should probably use this technique for all schedulers then?

bdice:
@csadorf Yes, consistency across schedulers would be good.

`distutils.spawn.find_executable` returns `None` if the command it's searching for is an alias. I don't know if that breaks anything, I think I've seen some clusters alias their submission command so that it worked (as a hack) with a different scheduler... like aliasing `qsub` to perform `sbatch`. 😞 I don't know if that affects any clusters we work on.

There is an alternative that works for Python 3.3+, `shutil.which`. It has the same behavior with aliases. https://docs.python.org/dev/library/shutil.html#shutil.which

csadorf:
> @csadorf Yes, consistency across schedulers would be good.
> 
> `distutils.spawn.find_executable` returns `None` if the command it's searching for is an alias. I don't know if that breaks anything, I think I've seen some clusters alias their submission command so that it worked (as a hack) with a different scheduler... like aliasing `qsub` to perform `sbatch`. 😞 I don't know if that affects any clusters we work on.
> 
> There is an alternative that works for Python 3.3+, `shutil.which`. It has the same behavior with aliases. https://docs.python.org/dev/library/shutil.html#shutil.which

I'm not sure how to deal with this, but unless there are any problems with that, I'm willing to ignore that for the time being.

csadorf:
@bdice I would like to just merge this and not change the behavior of all schedulers.

PR #56, opened 2019-02-13
Fix/refactor status updates

csadorf:
I've refactored a lot of code to achieve a better separation of legacy-related code paths and non-legacy related code paths. The `project` and `environment` modules should be much cleaner now.

That means that I've moved more legacy code into the already existing `legacy` module, as well introduced the `legacy_templating` module which serves a similar purpose. A lot of deprecation functions implemented for environments are decorated with `deprecated_since_06` to trigger a `DeprecationWarning` as well as inform developers about their deprecation status.

I also refactored the status update related functions. The status cache is now stored in the project document, not in the individual job documents. That should be safer and faster and prepares us for the aggregated-operations revision.

It might be a good idea to test status and pretend submission on a few real projects.


csadorf:
@vyasr @bdice I just pushed one more commit, but this is ready for review.

flow/project.py
```diff
@@ -281,14 +266,12 @@ def __init__(self, name, job, cmd, directives=None, np=None):
...
-            warnings.warn(
-                "The np argument for the JobOperation constructor is deprecated.",
-                DeprecationWarning)
-            assert directives.setdefault('np', np) == np
-        else:
-            # Future: We can remove everything but the else: clause.
-            directives.setdefault(
-                'np', directives.get('nranks', 1) * directives.get('omp_num_threads', 1))
+            raise RuntimeError(
+                "The np argument for the JobOperatoin constructor has been deprecated "
```
bdice: JobOperatoin ➡️ JobOperation, "as of version 0.6" below

flow/legacy.py
```diff
@@ -43,10 +43,10 @@ def wrapper(self, *args, **kwargs):
         # Raise an exception or warn based on the detected API version.
         if api_version is None:
             raise RuntimeError("Unable to determine legacy API use.")
-        elif api_version < 3:
-            raise RuntimeError(
+        elif api_version < 4:
+            warnings.warn(
                 "This FlowProject implementation uses deprecated API "
-                "version 0.{}.x Please downgrade your signac-flow installation "
+                "version 0.{}.x. Please downgrade your signac-flow installation "
```
bdice: Are these APIs labeled by the _earliest_ version to support that syntax, or the _last_ version to support that syntax? I would rather see the _last_ version somewhere, so that users would know what to install to get maximal features/bugfixes while retaining compatibility.

csadorf:
> All of these changes seem right to me, but I can't say I'm familiar enough with the `flow` internals to weigh in more strongly. I am eager to see the `flow` codebase become a little simpler. The net impact of this pull request is an increase in the number of lines - largely because most (all?) functional deprecated code was moved and wrapped in legacy-warning APIs. I would like to see support being cut quickly just to make the code less complicated, as these features have been deprecated for a long while.

Yes, I agree and this patch is a step in that direction. While the net increase might be positive, the legacy code paths are now clearly separated and that should make it much easier to just remove all of it.

I suggest we move fast on this. Let's release 0.7 ASAP to give users a a chance to adjust and then we can clear up the code base immediately moving forward.

flow/legacy.py
csadorf: I'm not sure that I understand the question. The APIs are labeled by the version that they were current. I think I am actually going to elevate this warning to an error.

flow/project.py
```diff
@@ -281,14 +266,12 @@ def __init__(self, name, job, cmd, directives=None, np=None):
...
-                "The np argument for the JobOperation constructor is deprecated.",
-                DeprecationWarning)
-            assert directives.setdefault('np', np) == np
-        else:
-            # Future: We can remove everything but the else: clause.
-            directives.setdefault(
-                'np', directives.get('nranks', 1) * directives.get('omp_num_threads', 1))
+            raise RuntimeError(
+                "The np argument for the JobOperation constructor has been deprecated "
+                "as of version 0.6 and been removed as of version 0.7!")
```
vyasr: I'm confused why this is an error, whereas most of the "will be removed in 0.7" (e.g. the `deprecated_since_06` decorator) is just a warning.

flow/project.py
```diff
@@ -1002,53 +856,20 @@ def scheduler_jobs(self, scheduler):
...
-            A nested dictionary (job_id, op_name, scheduler jobs)
-        """
-        sjobs_map = defaultdict(dict)
-        for job_id, op, sjob in self._map_scheduler_jobs(scheduler_jobs):
-            sjobs = sjobs_map[job_id].setdefault(op, list())
-            sjobs.append(sjob)
-        return sjobs_map
+        """Function was removed as of version 0.7."""
+        raise RuntimeError("The FlowProject.map_scheduler_jobs() function has "
+                           "been removed as of version 0.7!")
```
vyasr: Same issue as before; I thought our plan was to release a 0.6.5 version or so with these deprecation warnings uniformly present, then remove everything. Otherwise our legacy templating system fixes should be changed to indicate that those things will be kept until 0.8 or so.

csadorf:
We'll wait until pull request #57 has been merged and then update this PR to match the accepted deprecation policy accordingly.

csadorf:
Here are my next steps:

 - [ ] Go through all deprecations and identify which one are considered *core API*.
 - [ ] Go through the deprecation handling and ensure it complies with the [deprecation policy](https://docs.signac.io/projects/flow/en/latest/support.html#deprecation-policy).

PR #57, opened 2019-02-14
Add deprecation policy to dev docs and add link to changes doc page.

csadorf:
I added a detailed deprecation policy to the developer documentation and a link to the changes page as part of the user documentation.

**Let me know if you want me to build this on readthedocs!**

### DIFF
--- a/doc/support.rst
+++ b/doc/support.rst
@@ -157,13 +157,15 @@ then completely removed in any following minor or major release.
 
 *A feature is considered to be part of the core API if it is likely to be used by the majority of existing projects.*
 
-For example, a feature which is deprecated in version *0.x*, will trigger
+A feature which is deprecated in version *0.x*, will trigger
 warnings for all releases with release number *0.x.\**, and will be removed in
 version *0.y.0*.
 A feature, which is deprecated in version *0.x* and which is considered core
 API will trigger warnings for versions *0.x.\** and *0.y.\**, limited backwards
 compatibility will be maintained throughout versions *0.y.\**, and the feature
 will be removed in version *0.z.0*.
+
+**For example: A feature deprecated in version 0.6, will be removed in version 0.7, unless it is considered core API, in which case, some backwards compatibility is maintained in version 0.7, and it is removed in version 0.8.**
 
 After the release of version *1.0*, we apply the following deprecation policy:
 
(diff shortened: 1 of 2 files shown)


csadorf:
@glotzerlab/signac-maintainers @mikemhenry I would like to get approval from all maintainers for this PR if possible.

mikemhenry:
You might need add everyone as a reviewer @csadorf

doc/support.rst
```diff
@@ -138,3 +138,38 @@ For example:
...
+
+For example, a feature which is deprecated in version *0.x*, will trigger
+warnings for all releases with release number *0.x.\**, and will be removed in
+version *0.y.0*.
+A feature, which is deprecated in version *0.x* and which is considered core
+API will trigger warnings for versions *0.x.\** and *0.y.\**, limited backwards
+compatibility will be maintained throughout versions *0.y.\**, and the feature
+will be removed in version *0.z.0*.
+
+After the release of version *1.0*, we apply the following deprecation policy:
```
vyasr: I think it's a little strange to address this here. I would prefer to just change the deprecation policy when we hit 1.0, since everything above is talking about how our API is not stable yet but we try to maintain backwards-compatibility. However, it's not a dealbreaker if you just want to leave it in for now.

doc/support.rst
```diff
@@ -138,3 +138,38 @@ For example:
...
+
+*A feature is considered to be part of the core API if it is likely to be used by the majority of existing projects.*
+
+For example, a feature which is deprecated in version *0.x*, will trigger
+warnings for all releases with release number *0.x.\**, and will be removed in
+version *0.y.0*.
+A feature, which is deprecated in version *0.x* and which is considered core
+API will trigger warnings for versions *0.x.\** and *0.y.\**, limited backwards
+compatibility will be maintained throughout versions *0.y.\**, and the feature
+will be removed in version *0.z.0*.
```
vyasr: The x, y, z order is not obvious.

csadorf:
@vyasr I've updated per your suggestions, but left x,yz in place and instead added a concrete example. The post-final policy is removed, we will just update that once that becomes relevant.

vyasr:
Sounds good to me. Everything looks fine to me now, so merge at your discretion.

PR #63, opened 2019-02-22
Fix submission of uneven jobs on summit

jglaser:
When the number of resources requested is not divisible by the number of resources per node, this fix fills the minimum number of nodes possible on summit.

## Description
The current formula using gcd can create jobs with nodes that are underutilized (e.g., only 5 out of 6 GPUs used). This PR fixes that and falls back onto the formula only if it CPU core/GPU requirements are incommensurable.

## Motivation and Context
Make sensible use of Summit's hardware. May potentially require updating unit tests (not done yet).

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

This pull request is based on
- [x] *develop*, ~because it introduces a new feature.~

## Checklist:

If necessary:


flow/environments/incite.py
```diff
@@ -21,7 +21,7 @@ class SummitEnvironment(DefaultLSFEnvironment):
         @directives(ngpu=3) # 3 GPUs
         @directives(np=3) # 3 CPU cores
         @directives(rs_tasks=3) # 3 tasks per resource set
-        @directives(extra_jsrun_args='-smpiargs="-gpu"') # extra jsrun arguments
+        @directives(extra_jsrun_args='--smpiargs="-gpu"') # extra jsrun arguments
```
csadorf: How is this related to the fix?

jglaser: > How is this related to the fix?

It is only related in that it pertains to the Summit template. The double dash is necessary for jsrun, the previous example contained a mistake.

csadorf:
I looked at the [test output](https://circleci.com/gh/glotzerlab/signac-flow/201?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) and noticed that it seems that this change has some unintended side-effects. Specifically, all the thread-parallelized operations are executed incorrectly.

csadorf:
@jglaser Did you get a chance to look at the issue I pointed out above? I am unable to accept the patch in its current form unless we are able to avoid the regression. @bdice Could you briefly weigh in on this?

jglaser:
I'll try to have a look later.

jglaser:
> I'll try to have a look later.

I fixed the openmp job parameters, but it turned out the parameters for `parallel_op` were incorrect (-c 3 instead of -n 3), so I updated the reference scripts.

jglaser:
I also removed the now the ascent environment, as that host ceased to exist

jglaser:
> Unfortunately I need to revise my previous assessment. I generated a diff of the template data and a new review document and found one issue (files attached):
> [diff.txt](https://github.com/glotzerlab/signac-flow/files/2923325/diff.txt)
> [SummitEnvironment.docx](https://github.com/glotzerlab/signac-flow/files/2923326/SummitEnvironment.docx)
> 
> Specifically, I believe that the `mpi_gpu_op` test case is wrong. The new template requests 5 cores per rank, when we actually only need one core.
> 
> ```
>   # mpi_gpu_op(e6cbac22c5887a52771be793228ff1a9)
> - jsrun -n 1 -a 5 -c 1 -g 2 -d packed -b rs /usr/local/bin/python generate_template_reference_data.py exec mpi_gpu_op e6cbac22c5887a52771be793228ff1a9
> ?                    ^
> + jsrun -n 1 -a 5 -c 5 -g 2 -d packed -b rs /usr/local/bin/python generate_template_reference_data.py exec mpi_gpu_op e6cbac22c5887a52771be793228ff1a9
> ```

When requesting MPI tasks (-a 5) in a single resource set you also need to make sure that you have the matching number of cores allocated (-c 5). So the second (current) version should be correct for `nranks=5` and `ngpu=2`.

csadorf:
> > Unfortunately I need to revise my previous assessment. I generated a diff of the template data and a new review document and found one issue (files attached):
> > [diff.txt](https://github.com/glotzerlab/signac-flow/files/2923325/diff.txt)
> > [SummitEnvironment.docx](https://github.com/glotzerlab/signac-flow/files/2923326/SummitEnvironment.docx)
> > Specifically, I believe that the `mpi_gpu_op` test case is wrong. The new template requests 5 cores per rank, when we actually only need one core.
> > ```
> >   # mpi_gpu_op(e6cbac22c5887a52771be793228ff1a9)
> > - jsrun -n 1 -a 5 -c 1 -g 2 -d packed -b rs /usr/local/bin/python generate_template_reference_data.py exec mpi_gpu_op e6cbac22c5887a52771be793228ff1a9
> > ?                    ^
> > + jsrun -n 1 -a 5 -c 5 -g 2 -d packed -b rs /usr/local/bin/python generate_template_reference_data.py exec mpi_gpu_op e6cbac22c5887a52771be793228ff1a9
> > ```
> 
> When requesting MPI tasks (-a 5) in a single resource set you also need to make sure that you have the matching number of cores allocated (-c 5). So the second (current) version should be correct for `nranks=5` and `ngpu=2`. See e.g. here
> 
> https://jsrunvisualizer.olcf.ornl.gov
> 
> and here
> https://www.olcf.ornl.gov/for-users/system-user-guides/summit/summit-user-guide/#running-jobs
> 
> (e.g. Multiple tasks, single GPU per RS)

Got it, thank you very much for clearing that up for me.

Issue #58, opened 2019-02-14
single string names input won't be auto detected in project.submit()


zhou-pj:
### Description

When submitting jobs with python interface using project.submit(), and the names input is only one single string rather than list, submit won't auto detect single string and won't submit any job.

### To reproduce

project.submit(names='operation_name',...) 
won't submit any jobs as compare to

project.submit(names=['operation_name'],...) 

### System configuration

 - Operating System [stampede2]:
 - Version of Python [3.7]:
 - Version of signac [0.9.5]:
 - Version of signac-flow [0.6.4]:

csadorf:
@zhou-pj I've created a patch that detects the bad `names` argument and raises an exception: See pull request #66.

csadorf:
Resolved with merge of #66 .

PR #66, opened 2019-02-26
Fix/issue 58

csadorf:
Fixes issue #58.

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -1815,7 +1815,7 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
         if isinstance(names, six.string_types):
             raise ValueError(
                 "The 'names' argument must be a sequence of strings, however you "
-                "provided a single string.")
+                "provided a single string: {}.".format(names))
         if env is None:
             env = self._environment
         if walltime is not None:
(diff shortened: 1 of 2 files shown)


flow/project.py
```diff
@@ -1812,6 +1812,10 @@ def submit(self, bundle_size=1, jobs=None, names=None, num=None, parallel=False,
         # Regular argument checks and expansion
         if jobs is None:
             jobs = self  # select all jobs
+        if isinstance(names, six.string_types):
+            raise ValueError(
+                "The 'names' argument must be a sequence of strings, however you "
+                "provided a single string.")
```
mikemhenry: LGTM, however, I think adding the string that the user provided could add some clarity? Something like  
```suggestion
                "provided a single string: {}.".format(names))
```
I think it could help a user see more clearly the input they are providing is wrong. 

~Also, is there a reason we say sequence?~  Never mind, that is what we call it in the documentation :) Yay consistency!

csadorf:
@mikemhenry I've committed your suggested change, anything else I should change?

mikemhenry:
Sorry about that, I should have made that comment/review as 'once resolved approve'

Issue #59, opened 2019-02-14
Bridges Template missing --gres=gpu:p100:2 with GPU-shared partition


evanmiller326:
### Description

When trying to submit a job to the GPU-shared partition on Bridges with Signac Flow version 0.6.4 results in a failure ` sbatch: error: No gpu count given, specify --gres=gpu `. 

### To reproduce

Within the template bridges.sh where the resources are allocated based on the partition chosen:

```
{% elif partition == 'GPU-shared' %}
#SBATCH -N {{ nn|default(1, true)|check_utilization(gpu_tasks, 1, threshold, 'GPU') }}
#SBATCH --ntasks-per-node=16
{% elif 'shared' in partition %}
#SBATCH -N {{ nn|default(1, true) }}
#SBATCH --ntasks-per-node={{ cpu_tasks }}
{% else %}
```
This section needs ` #SBATCH --gres=gpu:p100:2` 

### Error output

If possible, copy any terminal outputs or attach screenshots that provide additional information on the problem.

``` 
sbatch: error: No gpu count given, specify --gres=gpu
             See https://www.psc.edu/bridges/user-guide
sbatch: error: Batch job submission failed: Invalid generic resource (gres) specification
Submission error: sbatch error:
make: *** [run] Error 1 
```

### System configuration

Please complete the following information:

```
Linux-3.10.0-693.21.1.el7.x86_64-x86_64-with-centos-7.4.1708-Core
3.7.1 (default, Dec 14 2018, 19:28:38)
[GCC 7.3.0]
0.9.5
0.6.4
```

mikemhenry:
Looks like a regression from here https://github.com/glotzerlab/signac-flow/commit/5b02ecf8f55453675f47decfbdd5a862090b2bb7#diff-309c6ee2196cea793ac60cc84f55be35L35

Thanks for the report @evanmiller326 :tada:

csadorf:
Fixed with merge of PR #69 .

PR #69, opened 2019-03-01
Add missing --gres=gpu.. line to bridges template.

csadorf:
Fixes issue #59


csadorf:
> This maybe more like question then comment since I have not used bridges before. I check on bridges for P100 you can only get either 1 or 2, but I wonder why here it is always set to 2 since the default #SBATCH -N is 1 and every other thing is divided by 2 compare to the 'GPU-small' section above. Or do we need something like {{ (gpu_tasks, 2)|min }} here for when there is only one gpu task?

Good point! This should probably be `#SBATCH --gres=gpu:p100:1`. This is also confirmed when I look at this [example script ](https://www.psc.edu/bridges/user-guide/sample-batch-scripts#gpu-shared)from the user guide.

csadorf:
I just reached out to the Bridges support for clarification.

csadorf:
I have clarified this with the Bridges admins. In summary:
1. The number of requested GPUs should be **one or multitudes of two** when requesting on the `GPU-shared` partition.
2. We should use `#SBATCH --gres=gpu:p100:1` when we request **one** GPU.
3. We should use `#SBATCH --gres=gpu:p100:2` when we request a **multitude of 2** GPUs.
4. It is generally preferable to use the `GPU` partition when requesting more than one GPU.

I suggest that we update the template as follows:

* [ ] Enforce that the number of requested GPUs is one or a multitude of 2.
* [ ] Add the `#SBATCH --gres=gpu:p100:X` line, where `X=1` when requesting **one** GPU, otherwise `X=2`.
* [ ] Print a warning suggesting that the user should preferably request from the GPU partition when requesting more than one GPU.

csadorf:
> Print a warning suggesting that the user should preferably request from the GPU partition when requesting more than one GPU.

We have no mechanism to issue warnings from within templates, I'll skip this therefore.

PR #71, opened 2019-03-01
Build dag

vyasr:
## Description
This PR adds the ability to detect the operation graph implied by the sequence of pre- and post-conditions applied to operations associated with a FlowProject.

## Motivation and Context
For debugging and graphical purposes, it's nice to be able to see the full operation graph. Additionally, this is a first step towards enabling more automated execution of dependencies for operations.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

This pull request is based on

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -692,16 +692,16 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
     def detect_operation_graph(self):
         R"""Determine the directed acyclic graph defined by operation pre- and
         post- conditions."""
-        num_ops = len(self._operations)
+        num_ops = len(self.operations)
         mat = [[0 for _ in range(num_ops)] for _ in range(num_ops)]
-        ops = list(self._operations.items())
+        ops = list(self.operations.items())
         for i, (name1, op1) in enumerate(ops):
             for j, (name2, op2) in enumerate(ops[i:]):
                 if set(op1._postconds).intersection(set(op2._prereqs)):
                     mat[i][j+i] = 1
                 elif set(op1._prereqs).intersection(set(op2._postconds)):
                     mat[j+i][i] = 1
-        return ops, mat
+        return mat
 
     pre = _pre
     """Decorator to add a pre-condition function for an operation function.
--- a/flow/project.py
+++ b/flow/project.py
@@ -175,7 +175,7 @@ class _condition(object):
         return cls._get_func('not', condition, lambda job: not condition(job))
 
     @classmethod
-    def _multi(cls, funcs):
+    def all_(cls, funcs):
         R"""Generate a condition function equivalent to a sequence of
         conditions functions. In order to make it identifiable to the graph
         building algorithm, we need to tag this function with all the
@@ -198,12 +198,12 @@ class _pre(_condition):
     @classmethod
     def copy_from(cls, other_func):
         "True if and only if all pre conditions of other function are met."
-        return cls._multi(getattr(other_func, '_flow_pre', list()))
+        return cls.all_(getattr(other_func, '_flow_pre', list()))
 
     @classmethod
     def after(cls, other_func):
         "True if and only if all post conditions of other function are met."
-        return cls._multi(getattr(other_func, '_flow_post', list()))
+        return cls.all_(getattr(other_func, '_flow_post', list()))
 
 
 class _post(_condition):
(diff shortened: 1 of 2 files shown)


flow/project.py
```diff
@@ -104,20 +104,63 @@
 
 class _condition(object):
 
-    def __init__(self, condition):
+    # All condition functions created using the _condition class are stored in this
+    # global registry variable, allowing direct equivalence comparison between
+    # condition objects.
+    _CONDITION_FUNCTIONS = dict()
```
csadorf: While this variable is technically global, I believe that it's convention to still write them lower-caps, because it is not on the module-level.

csadorf:
@vyasr Could we maybe identify lambda functions based on their `__code__` object? Two anonymous functions with the exact same code will have the same `__code__.co_code` object as far as I know.

Proof of concept:
```python
a = lambda x: x**2
b = lambda x: x**2

assert a.__code__.co_code == b.__code__.co_code
```

Of course, if a user uses different variable names or chooses to express the condition differently, e.g. with `x*x` instead of `x**2`, this will fail, but maybe better than nothing?

flow/project.py
```diff
@@ -104,20 +104,63 @@
...
+
+    def __init__(self, condition, _internal_lambda=False):
+        """The _internal_lambda argument is used to indicate that a condition
+        function is part of the built-in registry of lambdas and can be
+        identified as such during graph detection."""
+        conditions = type(self)._CONDITION_FUNCTIONS.setdefault(
+            'conditions', list())
```
csadorf: Searching for the string 'conditions' on this page returns nothing. When are we actually accessing this list?

flow/project.py
```diff
@@ -104,20 +104,63 @@
...
-    def __init__(self, condition):
+    # All condition functions created using the _condition class are stored in this
+    # global registry variable, allowing direct equivalence comparison between
+    # condition objects.
+    _CONDITION_FUNCTIONS = dict()
+
+    def __init__(self, condition, _internal_lambda=False):
+        """The _internal_lambda argument is used to indicate that a condition
+        function is part of the built-in registry of lambdas and can be
+        identified as such during graph detection."""
```
csadorf: I find the `_internal_lambda` parameter and handling very confusing. It took me a while to understand this and I am still not quite sure that I really do. 

This is really only used in combination with `_get_func()`, correct?

flow/project.py
vyasr: Yes, I think you are correct.

flow/project.py
vyasr: Yes, you are correct. The reason I introduced it is because in general we decided not to try and use any sort of code introspection to determine equality of lambda functions, but all of our pre-made conditions are implemented as lambdas. I use this flag later to distinguish between the lambdas we create internally and the lambdas that users pass in.

flow/project.py
vyasr: The list is never explicitly accessed outside the constructor of `_condition`. The purpose of this list is to retain a complete set of all condition functions that have been provided. The conditional on the next line serves to keep track of which user-passed conditions are lambda functions.

vyasr:
I should be clear that this PR really should be more like a draft PR in terms of its implementation, it just got made before that feature was added to Github. I'm totally open to changing the internals if we can find a different way to do things.

The original goal of this PR was to make it possible to define a complex workflow, request that a specific operation mid-workflow be executed, and then have flow automatically determine which other operations need to run in order to execute this one by following the graph backwards. There are other ancillary benefits, such as visualizing the graph to ensure that everything is defined as expected, but IIRC the automatic sequence detection and execution was the initial motivator.

flow/project.py
csadorf: So if we never actually access this dictionary with the keyword 'conditions', why is it not simply a list?

flow/project.py
csadorf: Maybe we should provide a thin-wrapper that users can use to wrap their anonymous functions that is slightly less intrusive:
```
def with_name(name, func):
    setattr(func, '_flow_func_name', name)
    return func

@Project.pre(with_name('myfunc', lambda job: ... ))
# ...
```

vyasr:
I haven't had a chance to really work on this again, but I'd like to spend some time to see if there's any way to implement the changes to condition functions in a simpler way. I'll ping this thread once I have time to give it a shot.

vyasr:
Closing pending further work.

PR #72, opened 2019-03-02
Fix/eos submission

jglaser:
This fixes two issues when trying to submit MPI jobs on EOS.

1. EOS uses a qsub wrapper (a perl script), which expects to run inside a tty.
2. the # of ranks per node was wrong (it is 32, not 24)

## Description
ad 1. to honor the shebang (#1) line for perl of the qsub wrapper, apparently only subprocess.Popen works, but check_output does not. To additionally create a tty, one has to use python's pty lib.

## Motivation and Context
Trying to submit on OLCF eos.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

This pull request is based on
- [x] *master*, because it is a bug fix or an update to the documentation.

## Checklist:

If necessary:

### DIFF
--- a/flow/scheduling/torque.py
+++ b/flow/scheduling/torque.py
@@ -144,7 +144,8 @@ class TorqueScheduler(Scheduler):
             # some qsub wrappers don't run on no-tty
             import pty
             master, slave = pty.openpty()
-            subprocess.Popen(['qsub', '--version'],stdin=slave, stdout=slave, stderr=slave, shell=True)
+            subprocess.Popen(['qsub', '--version'],stdin=slave, stdout=slave, stderr=slave,
+                shell=True)
         except (IOError, OSError):
             return False
         else:
--- a/flow/scheduling/torque.py
+++ b/flow/scheduling/torque.py
@@ -144,7 +144,7 @@ class TorqueScheduler(Scheduler):
             # some qsub wrappers don't run on no-tty
             import pty
             master, slave = pty.openpty()
-            subprocess.Popen(['qsub', '--version'],stdin=slave, stdout=slave, stderr=slave,
+            subprocess.Popen(['qsub', '--version'], stdin=slave, stdout=slave, stderr=slave,
                              shell=True)
         except (IOError, OSError):
             return False
--- a/flow/scheduling/torque.py
+++ b/flow/scheduling/torque.py
@@ -141,7 +141,10 @@ class TorqueScheduler(Scheduler):
     def is_present(cls):
         "Return True if it appears that a TORQUE scheduler is available within the environment."
         try:
-            subprocess.check_output(['qsub', '--version'], stderr=subprocess.STDOUT)
+            # some qsub wrappers don't run on no-tty
+            import pty
+            master, slave = pty.openpty()
+            p = subprocess.Popen(['qsub', '--version'],stdin=slave, stdout=slave, stderr=slave, shell=True)
         except (IOError, OSError):
             return False
         else:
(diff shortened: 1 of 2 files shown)


flow/scheduling/torque.py
```diff
@@ -141,7 +141,11 @@ def submit(self, script, after=None, pretend=False, hold=False, flags=None, *arg
     def is_present(cls):
         "Return True if it appears that a TORQUE scheduler is available within the environment."
         try:
-            subprocess.check_output(['qsub', '--version'], stderr=subprocess.STDOUT)
+            # some qsub wrappers don't run on no-tty
+            import pty
+            master, slave = pty.openpty()
+            subprocess.Popen(['qsub', '--version'], stdin=slave, stdout=slave, stderr=slave,
+                             shell=True)
```
csadorf: It appears that this function is not reliably raising either an `IOError` or `OSError` to indicate the scheduler is *not* present.

For example, I executed this command on a machine that I know has no Torque scheduling system: 
```
$ python -c "from flow.scheduling import TorqueScheduler; print(TorqueScheduler.is_present())"
True
```
I'm pretty sure that this is also the reason why unit tests are failing.

flow/templates/eos.sh
```diff
@@ -7,7 +7,7 @@
 {% if operations|calc_tasks('ngpu', true, true) and not force %}
 {% raise "GPUs were requested but are unsupported by Eos!" %}
 {% endif %}
-{% set nn = nn|default(cpu_tasks|calc_num_nodes(24), true) %}
+{% set nn = nn|default(cpu_tasks|calc_num_nodes(32), true) %}
```
csadorf: Thank you for catching this.

csadorf:
@jglaser I've scheduled this PR for the v0.7, however I've identified some issues with the patch which would need to be resolved prior to merging.

Do you need any assistance with this? Maybe we just split this so that the problematic changes to the `torque.py` module can be addressed separately?

jglaser:
> @jglaser I've scheduled this PR for the v0.7, however I've identified some issues with the patch which would need to be resolved prior to merging.
> 
> Do you need any assistance with this? Maybe we just split this so that the problematic changes to the `torque.py` module can be addressed separately?

I will try to get to it soon, perhaps over the weekend. Feel free to fix the 24->32 issue separately otherwise.

csadorf:
> > @jglaser I've scheduled this PR for the v0.7, however I've identified some issues with the patch which would need to be resolved prior to merging.
> > Do you need any assistance with this? Maybe we just split this so that the problematic changes to the `torque.py` module can be addressed separately?
> 
> I will try to get to it soon, perhaps over the weekend. Feel free to fix the 24->32 issue separately otherwise.

Ok, sounds good. I'll move this to v0.8 and fix the 24->32 issue as part of #69 .

csadorf:
Sorry, I got confused with the dates, there is still more than enough time to make this part of v0.7.

csadorf:
@jglaser I'm planning on releasing 0.7.0 either today or tomorrow? Should we reschedule this for 0.8.0?

jglaser:
For some reason I *was* able to submit on eos with the new `feature/enable-aggregate-operations` branch, which includes develop. Perhaps this branch is now superfluous?

csadorf:
> For some reason I _was_ able to submit on eos with the new `feature/enable-aggregate-operations` branch, which includes develop. Perhaps this branch is now superfluous?

It might be. I can leave it open a bit longer if you like. Otherwise I'll decline it.

csadorf:
I'll close this for now.

PR #77, opened 2019-03-08
Fix/eos incorrect node size

csadorf:
The eos node size was set to 24 instead of the correct number: 32.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -49,6 +49,7 @@ We recommend to go through the tutorial on signac-docs_ to learn how to best tak
 - Fixes issues with the Summit environment resource set calculation for parallel operations under specific conditions.
 - Remove the Ascent environment (host decomissioned).
 - Fixes issue where callable cmd-directives were not evaluated.
+- Fix the node size specified in the template for the ORNL Eos system.
 
 [0.6.4] -- 2018-12-28
 ---------------------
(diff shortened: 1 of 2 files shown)


mikemhenry:
Just one note 
> To match the charge factor for Titan, the usage factor for Eos will be set at 30 times nodes requested. The actual number of cores per nodes is 16 without Intel Hyper Threading (default) or 32 with Intel Hyper Threading enabled (-j2 aprun option). 

It doesn't look like the script actually does any check/ensure that the `-j2` option is used, will that be an issue?

csadorf:
> Just one note
> 
> > To match the charge factor for Titan, the usage factor for Eos will be set at 30 times nodes requested. The actual number of cores per nodes is 16 without Intel Hyper Threading (default) or 32 with Intel Hyper Threading enabled (-j2 aprun option).
> 
> It doesn't look like the script actually does any check/ensure that the `-j2` option is used, will that be an issue?

Good point! I just did some research on this.

The default is `-j 0` which means "use the system default". The system default on eos is `-j 2`. So no, I don't think that this is an issue.

From: https://www.olcf.ornl.gov/for-users/system-user-guides/eos/running-jobs/

> For Eos: The -j parameter controls Hyper Threading. The valid values for -j are 0 (use the system default), 1 (turn Hyper Threading off), and 2 (turn Hyper Threading on; this is the system default).

And:

> Code Example 1: (32) MPI tasks, (1) task per “core”, with Hyper Threading
> ```
> # Request 1 node.
> #PBS -l nodes=1# Tell aprun to use hyper threading.
> 
> # Implicitly using HT (default)
> aprun -n 32 ./a.out
> 
> # Explicitly using HT
> aprun -n 32 -j2 ./a.out
> ```

PR #79, opened 2019-03-08
Misc/automate release management

csadorf:
Various updates to automate the release management.


.zenodo.json
```diff
@@ -1,44 +1,67 @@
...
+            "orcid": "0000-0002-7197-0085"
         }
-    ]
-}
+    ],
+    "description": "The signac-flow tool provides the basic components to set up simple to complex workflows for projects as part of the signac framework. That includes the definition of data pipelines, execution of data space operations and the submission of operations to high-performance super computers.",
+    "keywords": [
+        "python",
+        "reproducibility",
+        "sharability",
```
bdice: typo: shareability, but I also don't think that shareability applies to signac-flow as much as signac. I would consider "automation" or "high performance computing" or similar.

.zenodo.json
```diff
@@ -1,44 +1,67 @@
...
+        "workflow",
+        "scientific computing"
+    ],
+    "license": {
+        "id": "http://www.opensource.org/licenses/BSD-3-Clause"
```
bdice: I would switch this to https://opensource.org/licenses/BSD-3-Clause

CITATION.cff
```diff
@@ -0,0 +1,48 @@
...
+    orcid: "https://orcid.org/0000-0002-7197-0085"
+keywords:
+  - python
+  - "data management"
+  - reproducibility
+  - sharability
+  - workflow
+  - "scientific computing"
+license: "http://www.opensource.org/licenses/BSD-3-Clause"
+message: "Development and deployment supported by MICCoM, as part of the Computational Materials Sciences Pro- gram funded by the U.S. Department of Energy, Office of Science, Basic Energy Sciences, Materials Sciences and Engineering Division, under Subcontract No. 6F-30844. Project conceptualization and implementation supported by the National Science Foundation, Award # DMR 1409620."
```
bdice: Fix `Pro- gram` to `Program` and see if there is a reasonable way to wrap this.

doc/support.rst
```diff
@@ -35,13 +35,13 @@ for example with `venv <https://docs.python.org/3/library/venv.html>`_:
 
     ~ $ python -m venv ~/envs/signac-flow-dev
     ~ $ source ~/envs/signac-flow-dev/bin/activate
-    (signac-flow-dev) ~ $ pip install six flake8
+    (signac-flow-dev) ~ $ pip install -r requirements-dev.txt
```
bdice: I'm hesitant about recommending the text file when it has exact version pinnings, instead of `>=`. I understand that those exact requirements should be used for our sanity during tests, but let's avoid forcing contributors to potentially downgrade unless necessary. Perhaps make a separate requirements file for the tests?

.zenodo.json
csadorf: Thank's for the catch! Are you open to the argument that signac-flow enables sharing of workflows? I imagine that down the road sharing of project-implementations becomes more common.

doc/support.rst
csadorf: I agree with that sentiment and I definitely see the issue. Maybe using a [constraint file](https://pip.pypa.io/en/stable/user_guide/#constraints-files) is a way to mitigate this?

bdice: I don't see how a constraint files solves the problem... I'd consider making a `requirements-test.txt` that pins `==` exact versions and a `requirements-dev.txt` that uses `>=` pinnings.

.zenodo.json
csadorf: I'm actually going to leave this for now, because I copied that from the zenodo metadata and I'm not sure that zenodo is otherwise going to recognize it properly...

csadorf:
@bdice I've applied most of the requested changes.

csadorf:
> Sorry, I accidentally started commenting on outdated files a moment ago. Ignore those comments. (Also please click "Resolve Conversation" on review comments after handling them.)

I usually leave them open if I think that it is the referee's call whether the discussion is resolved.

> I'm not sure if our handling of `requirements.txt` files is correct. Do devs installing with `requirements-dev.txt` not need jinja and cloudpickle from `requirements.txt`? I would think that dev requirements are everything needed to run all features and tests, PLUS other stuff. If I'm interpreting this wrong, please correct me.

The developer would install the package with `pip install -e .`, which will install the other "standard" requirements.

> Also the requirements are listed in different orders in `requirements-dev.txt` and `requirements-test.txt`, which was a little weird.

I'll fix that.

PR #68, opened 2019-02-26
Add legend for scheduler status symbols.

csadorf:
Fixes issue #39.

PR #80, opened 2019-03-09
Revert move of legacy functions from environments into separate module.

csadorf:
This commit reverts the previous refactorization where a lot of the
legacy methods of environments were moved into the legacy_templating
module. This refactorization turned out to lead to issues in combination
with environments derived from these classes.


bdice:
This fixed the problem I had with recursion when testing PR https://github.com/glotzerlab/signac-flow/pull/68.

Issue #19, opened 2018-03-05
The project status view should have an option to automatically show all varied parameters


csadorf:
**[Original report](https://bitbucket.org/glotzer/signac-flow/issue/19) by Carl Simon Adorf (Bitbucket: [csadorf](https://bitbucket.org/csadorf), GitHub: [csadorf](https://github.com/csadorf)).**

----------------------------------------

There should be an option that allows the user to automatically show all those parameters in the status overview that are varied (non-const) over the data space.

So that they don't have to be specified via the `-p`/ `--parameters` argument.

csadorf:
Please make sure to coordinate with @zhou-pj on this, since he will be working #67 which will touch the same code segment.

I think it should be possible to work on both issues concurrently since this issue is probably more concerned with the command line API and the logic of gathering those parameters, but some coordination will probably be advantageous to avoid effort duplication.

PR #87, opened 2019-03-15
Feature/all params -- add command line option to print all statepoint parameters in a status call

rosecers:
## Description
L1045: Added "all_parameters" argument to print_status()
L1075-1078: Added description of "all_parameters" to print_status()
L1173-1175: When "all_parameters" is true and no parameters have been specified, assign all statepoint parameters to variable parameter. If parameters have also been specified, overrides "all_parameters" call.
L2072-2076: Added "--all-parameters" or "-l" options to command line interface.

## Motivation and Context
#https://github.com/glotzerlab/signac-flow/issues/19

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [X] New feature

This pull request is based on

## Checklist:

- [X] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md).
- [X] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac-flow/blob/master/ContributorAgreement.md).
- [X] My code follows the [code style guideline](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md#code-style) of this project.
- [X] The changes introduced by this pull request are covered by existing or newly introduced tests.

If necessary:
- [N/A] I have updated the API documentation as part of the package doc-strings.
- [N/A] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
- [X] I have updated the [changelog](https://github.com/glotzerlab/signac-flow/blob/master/changelog.txt).

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -1166,8 +1166,8 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
                 return x
 
         if parameters is not None and len(parameters) == 0:
-            sp = self.open_job(id=status['job_id']).statepoint()
-            parameters = list(sp.keys())
+            parameters = list(sorted({key for job in jobs for key in job.sp.keys()
+                                      if len(set([job.sp[key] for job in jobs])) > 1}))
 
         if detailed:
             rows_status = []
--- a/flow/project.py
+++ b/flow/project.py
@@ -1042,7 +1042,7 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
 
     @_support_legacy_api
     def print_status(self, jobs=None, overview=True, overview_max_lines=None,
-                     detailed=False, parameters=None, all_parameters=False,
+                     detailed=False, parameters=None,
                      skip_active=False, param_max_width=None,
                      expand=False, all_ops=False, only_incomplete=False, dump_json=False,
                      unroll=True, compact=False, pretty=False,
@@ -1072,10 +1072,6 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
             Print the value of the specified parameters.
         :type parameters:
             list of str
-        :param all_parameters:
-            Print the value of the all statepoint parameters.
-        :type parameters:
-            bool
         :param skip_active:
             Only print jobs that are currently inactive.
         :type skip_active:
@@ -1094,7 +1090,6 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
             err = sys.stderr
         if jobs is None:
             jobs = self     # all jobs
-        # print(all_parameters, parameters)
 
         tmp = self._fetch_status(jobs, err, ignore_errors, no_parallelize)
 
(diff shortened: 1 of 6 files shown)


flow/project.py
```diff
@@ -1042,7 +1042,8 @@ def _print_progress(x):
 
     @_support_legacy_api
     def print_status(self, jobs=None, overview=True, overview_max_lines=None,
-                     detailed=False, parameters=None, skip_active=False, param_max_width=None,
+                     detailed=False, parameters=None, all_parameters=False,
```
csadorf: Instead of adding an additional argument here, I'd suggest to use a special constant as argument to `parameters`. In this way we avoid breaking backwards-compatibility and the ambiguity when users provide both the `all_parameters` and the `parameters `argument.

So my suggestion is to add the following line somewhere at the top of the module:
```
PRINT_STATUS_ALL_PARAMETERS = True
```
And then check for that in line 1173: `if parameters is PRINT_STATUS_ALL_PARAMETERS:`.

flow/project.py
```diff
@@ -1089,6 +1094,7 @@ def print_status(self, jobs=None, overview=True, overview_max_lines=None,
             err = sys.stderr
         if jobs is None:
             jobs = self     # all jobs
+        # print(all_parameters, parameters)
```
csadorf: This left-over line should be removed.

flow/project.py
```diff
@@ -1071,6 +1072,10 @@ def print_status(self, jobs=None, overview=True, overview_max_lines=None,
             Print the value of the specified parameters.
         :type parameters:
             list of str
+        :param all_parameters:
+            Print the value of the all statepoint parameters.
```
csadorf: typo: "of all statepoint"

flow/project.py
```diff
@@ -2058,6 +2069,11 @@ def _add_print_status_args(cls, parser):
             nargs='*',
             help="Display select parameters of the job's "
                  "statepoint with the detailed view.")
+        view_group.add_argument(
+            '-l', '--all-parameters',
+            action='store_true',
+            help="Display all parameters of the job's "
+                  "statepoint with the detailed view.")
```
csadorf: I would replace `job's` with `jobs'` to indicate that we are showing the parameters for all jobs, not just of one.

rosecers:
We've updated the project.py script with the requested changes. Now, when the "-p" parameter is supplied with no strings (i.e. parameters is not None and len(parameters)==0), all parameters are included.

rosecers:
Updated!

csadorf:
> Updated!

Great! Circle-CI was misconfigured to not build on forks. That should be enabled now, but I have trouble triggering it. Can you push an empty commit to your branch to trigger the CI build?

This should do it:
```
git commit --allow-empty -m "trigger circle-ci"
git push
```

rosecers:
Done!

csadorf:
The failing tests are unrelated to your code, I'll address that.

PR #89, opened 2019-03-15
Feature/select operations with regex

csadorf:
## Description

This patch enables the selection of operations for run/script/submit
with a regular expression. For this one provides an argument that starts
with a slash ('/').

## Motivation and Context

When a user has specified a lot of operations (e.g. programmatically), it may be impractical to provide a long list of operations. In this case being able to select operations *by pattern* would be highly advantageous.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

This pull request is based on

## Checklist:

If necessary:


csadorf:
@glotzerlab/signac-maintainers Could I get some feedback on this draft?

flow/project.py
```diff
@@ -1582,14 +1583,18 @@ def _generate_operations(self, cmd, jobs, requires=None):
...
+        if operation_names is None:
+            names = patterns = None
+        else:
+            patterns = [name.lstrip('/') for name in operation_names if name.startswith('/')]
+            names = [name for name in operation_names if not name.startswith('/')]
 
-        for op in self.next_operations(*jobs):
-            if operation_names and op.name not in operation_names:
-                continue
-            if not self.eligible_for_submission(op):
```
vyasr: I don't think we want to lose the eligibility check here because the function is also used for determining operations to run and operations to generate scripts for (in addition to in submission).

flow/project.py
```diff
@@ -1582,14 +1583,18 @@ def _generate_operations(self, cmd, jobs, requires=None):
 
     def _get_pending_operations(self, jobs, operation_names=None):
         "Get all pending operations for the given selection."
-        operation_names = None if operation_names is None else set(operation_names)
+        if operation_names is None:
+            names = patterns = None
+        else:
+            patterns = [name.lstrip('/') for name in operation_names if name.startswith('/')]
+            names = [name for name in operation_names if not name.startswith('/')]
```
vyasr: For consistency with the broad scope of regex engines, should we both start _and end_ with slashes?

csadorf: Is that how it typically works? You can search in `vim` with only a trailing slash.

bdice: I _think_ it's an intended coincidence that `/` triggers the search command in `vim`, not a way of defining a regex.

On another note, most regular expression engines support any character as a regex "pattern delimiter." [1,2] Is it possible to interpret all operation names as regular expressions, and let "exact string matches with the name of an operation" be a special case of that?

[1] https://en.wikipedia.org/wiki/Regular_expression#Delimiters
[2] http://php.net/manual/en/regexp.reference.delimiters.php

csadorf: Looking at those links, using a delimiter on both ends is probably the best way to go. And slash is commonly used, so not the worst choice.

> Is it possible to interpret all operation names as regular expressions, and let "exact string matches with the name of an operation" be a special case of that?

I considered that, but it would definitely break backwards compatibility. If you had two operations *foo* and *foobar* you would now need to do something like this:
```./project.py run -o foo\$```

Extra akward, because we need to escape the `$`. If we believe that this is case is rare and trumps the use of delimiters, then we should consider it.

Comment (bdice) [reply flow/project.py]: I think there's a way you can use regexes as exact matches or something like that. Such that `foo` matches only `foo` and `foo.*` matches `foo` and `foobar`. This is definitely possible with one of the functions in `re`, let me know if you want me to figure out a code snippet for this behavior.

Comment (csadorf) [reply flow/project.py]: > I think there's a way you can use regexes as exact matches or something like that. Such that `foo` matches only `foo` and `foo.*` matches `foo` and `foobar`. This is definitely possible with one of the functions in `re`, let me know if you want me to figure out a code snippet for this behavior.

In case you are referring to `re.search()` vs `re.match()` you would be out of luck. `foo` is a match for `foo` and `foobar` in both cases. But it's possible that I'm missing something.

Comment (bdice) [reply flow/project.py]: https://docs.python.org/3/library/re.html#re.fullmatch
`re.fullmatch`, and it's Python 3.4+ only. 😉

```python
>>> re.fullmatch('foo', 'foo')
<_sre.SRE_Match object; span=(0, 3), match='foo'>
>>> re.fullmatch('foo', 'foobar')
>>> # no match found
```

flow/project.py
csadorf: Yes, exactly. And so it makes more sense to check the "eligible_for_submission()" in the `submit` function, see here: https://github.com/glotzerlab/signac-flow/blob/e2d3cf18a51a52e3f86a157d8df2e401f1ace9ad/flow/project.py#L1819

csadorf: Yes, exactly. And so it makes more sense to check the "eligible_for_submission()" in the `submit()` function, see here: https://github.com/glotzerlab/signac-flow/blob/e2d3cf18a51a52e3f86a157d8df2e401f1ace9ad/flow/project.py#L1819

flow/project.py
csadorf: Ok, since we have not dropped Python 2.7 support we would probably need to use this [recipe](https://stackoverflow.com/a/30212799).

I am open to this approach.

csadorf: Done. Works great I think.

vyasr: vim is designed to minimize keystrokes, so yes I'd assume that the keystroke choice was intentional. But in general bracketing the expression is the standard for regexes; I think ed/sed/grep are the closest to canonical sources for that.

Regarding having options for delimiters I don't think that's important. Not all regex flavors support that, so we don't have to. I like the fullmatch though!

flow/project.py
vyasr: I got confused reading the diff here and didn't realize that `next_operations` was being used, so I was concerned that we were never checking `op.eligible`. I wanted to make sure that we weren't losing that check while removing the `eligible_for_submission` check, so I agree that we're good.

csadorf:
Looks like we're good with the concept, I'll add unit tests etc and then we can ship this.

flow/project.py
```diff
@@ -1520,6 +1521,12 @@ def run(self, jobs=None, names=None, pretend=False, np=None, timeout=None, num=N
         if num and num < 0:
             num = None
 
+        # The 'names' argument must be a sequence, not a string.
+        if isinstance(names, six.string_types):
+            raise ValueError(
+                "The names argument of FlowProject.run() must be a sequence of strings, "
+                "not a string.")
```
vyasr: What's the reason that we can't just handle this conversion internally?

tests/test_project.py
```diff
@@ -540,6 +546,24 @@ def test_run_with_selection(self):
...
+                with self.assertRaises(ValueError):
+                    # The names argument must be a sequence of strings, not a string.
+                    project.run(names='op1')
+                project.run(names=['non-existent-op'])
+                self.assertFalse(any(job.isfile('world.txt') for job in even_jobs))
+                self.assertFalse(any(job.doc.get('test') for job in project))
+                project.run(names=['op1', 'non-existent-op'])
+                self.assertTrue(all(job.isfile('world.txt') for job in even_jobs))
+                self.assertFalse(any(job.doc.get('test') for job in project))
+                project.run(names=['op[^3]', 'non-existent-op'])
```
vyasr: Can you also add an assertion that `dynamic` is not in the doc (to ensure `op3` didn't execute)?

flow/project.py
csadorf: For the sake of a well-defined API. This could go either way...

tests/test_project.py
csadorf: Done.

Issue #25, opened 2018-05-02
Show a budget (CPU-hours, etc) before submitting jobs.


csadorf:
**[Original report](https://bitbucket.org/glotzer/signac-flow/issue/25) by Bradley Dice (Bitbucket: [bdice](https://bitbucket.org/bdice), GitHub: [bdice](https://github.com/bdice)).**

----------------------------------------

csadorf:
The simplest and immediate solution for this is probably to print the walltime directly from the base template.

Long-term, we should move away from the `--walltime` CLI argument that is currently implemented on a per-environment basis and make the walltime a first-class directive, that can be specified on a per-operation basis. The required walltime could then be summed up much more easily.

PR #90, opened 2019-03-15
Feature/walltime budget

rosecers:
## Description
L2367-2410: Added a function to calculate the budget of CPU hours.

## Motivation and Context
#25 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [X] New feature

This pull request is based on

## Checklist:

- [X] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md).
- [X] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac-flow/blob/master/ContributorAgreement.md).
- [X] My code follows the [code style guideline](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md#code-style) of this project.
- [X] The changes introduced by this pull request are covered by existing or newly introduced tests.

If necessary:
- [N/A] I have updated the API documentation as part of the package doc-strings.
- [N/A] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
- [X] I have updated the [changelog](https://github.com/glotzerlab/signac-flow/blob/master/changelog.txt).

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -18,6 +18,15 @@
             "affiliation": "University of Michigan",
             "name": "Timothy Moore",
             "orcid": "0000-0002-5709-7259"
+        },
+        {
+            "affiliation": "University of Michigan",
+            "name": "Rose Cersonsky",
+            "orcid": "0000-0003-4515-3441"
+        },
+        {
+            "affiliation": "University of Michigan",
+            "name": "Yuan Zhou"
         }
     ],
     "creators": [
--- a/contributors.yaml
+++ b/contributors.yaml
@@ -49,4 +49,13 @@ contributors:
     given-names: Timothy
     affiliation: "University of Michigan"
     orcid: "https://orcid.org/0000-0002-5709-7259"
+  -
+    family-names: Cersonsky 
+    given-names: Rose
+    affiliation: "University of Michigan"
+    orcid: "http://orcid.org/0000-0003-4515-3441"
+  -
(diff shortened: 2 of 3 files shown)


csadorf:
@rosecers I've rebased this branch on develop to only include the commits directly related to this pull request. Please follow these steps prior to adding additional commits to this branch:
```
git fetch --all
git checkout feature/walltime-budget
git reset --hard origin/feature/walltime-budget
```

csadorf:
I believe this is more or less fine as is, the only question I have is how we integrate this with a potential `walltime` directive for operations moving forward. Currently this requires storing the walltime in the project document.

We could use a hierarchical approach, where we look up the walltime from the job document, then the project document, and then finally from a directive. @glotzerlab/signac-developers Thoughts?

vyasr:
I don't think this necessarily conflicts with walltime directives, since we could just use a more descriptive name for this key (e.g. `expected_walltime` or `walltime_budget`). Having said that, I think that the approach you suggested makes sense; the directive is how much will be requested upon submission, so it should be the default expected value for estimation, but the user can override that by placing values in a job document.

csadorf:
I am closing this for now until we have figure out how to properly define the walltime as part of the directives.

PR #91, opened 2019-03-15
Feature/issue 85 full launcher support

zhou-pj:
## Description
implemented full launcher support for single processor job in stampede2 and make it default on stampede2 when bundle argument is present

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [X] New feature

This pull request is based on

## Checklist:

If necessary:


flow/templates/stampede2.sh
```diff
@@ -20,6 +32,27 @@
 {% endblock %}
 
 {% block body %}
+{% if not use_mpi %}
+{% if parallel%}
+{{("Warning! Bundle submission without mpi on stampede2 is using launcher and --parallel is ignored")|print_warning}}
```
csadorf: Could you use the following formulation?

> Bundled submission without MPI on Stampede2 is using launcher; the `--parallel` option is therefore ignored.

Please note that I intentionally removed the leading "Warning!".

flow/util/template_filters.py
```diff
@@ -156,6 +156,16 @@ def calc_num_nodes(np, ppn=1, threshold=0, name=None):
...
 
 
+def print_warning(warning):
+    """Print warning message within jinja2 template
+
+    :param:
+        The warning message as a string
+    """
+    print(warning)
+    return ''
```
csadorf: I assume this is necessary to make jinja2 happy?

zhou-pj: you mean the return '' part? yes with that jinja2 will put an empty line on the output file when the function is called and will not break anything

zhou-pj:
I am not entirely sure about the test output. I think Vyas and I decided that the launcher should be used only when --bundle is present and there is no mpi or tbb. If the implementation is doing so then the MPI/OpenMP/hybrid operations should not go down the launcher path. I might need more info on how the test if performed to see where went wrong.

zhou-pj:
The wrong behavior that launcher is used in MPI tests is because in jinjia2 variables initialized with set only live within the loop or block. This is corrected with namespace() function in jinjia2 setting variable with global access.

Once this is reviewed and approved I can run the tests/generate_template_reference_data.py to regenerate the test reference for launcher

flow/templates/stampede2.sh
```diff
@@ -20,6 +31,7 @@
 {% endblock %}
 
 {% block body %}
+{% if ns.use_mpi %}
```
csadorf: Can we replace this with `{% ns.use_mpi or operations.length < 2 %}` so that we only use launcher when we have actually more than one operation?

flow/project.py
```diff
@@ -247,6 +247,7 @@ def __init__(self, name, job, cmd, directives=None, np=None):
         directives.setdefault('ngpu', 0)
         directives.setdefault('nranks', 0)
         directives.setdefault('omp_num_threads', 0)
+        directives.setdefault('factor', 1)
```
csadorf: Can we rename this to `oversubscribe`?

@vyasr Would that be a good name?

zhou-pj: This factor is currently designed to be a number between (0,1) indicating the fraction of processor time needed for one single job, so maybe naming something like processor_fraction? Or if we want to use oversubscribe, would it make more sense to make the value >1 and use 1/oversubscribe to calculate resources requested instead? @csadorf @vyasr What do you think?

csadorf: I'm not sure either, I just think that `factor` is a bit too generic. How does the launcher-tool help page describe it?

flow/templates/stampede2.sh
zhou-pj: I think if we only have one operation, launcher would still be compatible but using mpi would possibly break if the job does not support mpi. On another note, on stampede2 if use_mpi is false and there is only one operation, this is not a good use of the resources and signac will stop the submission unless --force is added right?

flow/project.py
zhou-pj: the launcher does not directly use this value but only takes in num_process, which is currently calculated by num_operation * factor. How about I change it to processor_fraction or make it 1/ oversubscribe instead?

flow/templates/stampede2.sh
csadorf: > I think if we only have one operation, launcher would still be compatible but using mpi would possibly break if the job does not support mpi.

Sorry, I meant `{% ns.use_mpi and operations.length < 2 %}`. I just don't want to use launcher for a single operation.

> On another note, on stampede2 if use_mpi is false and there is only one operation, this is not a good use of the resources and signac will stop the submission unless --force is added right?

You could have a single operation that is "manually" parallelized without MPI.

flow/project.py
vyasr: I still think the fractional value makes the most sense. I agree that `factor` is too generic, though; `processor_fraction` is a more descriptive name that I think would be appropriate.

As Pengji said, the Launcher assumes that you're writing the job script with a specify number of nodes/tasks and then you manually construct the list of operations to run. This quantity is necessary for us in order to calculate the appropriate number of nodes/tasks dynamically based on the list of operations.

flow/templates/stampede2.sh
vyasr: Hmm that logic doesn't seem quite right. That would cause us to go the launcher route if we were submitting multiple MPI operations, right? I think your original suggestion with `or` is correct. Pengji raises a valid point, but I don't think it's a concern because `mpi_prefix` is only getting set within this block if `operations.directives.nranks` is set, so `ibrun` won't be inserted if the job is not requesting multiple ranks.

I think we should rename the variable from `use_mpi` to `use_launcher` or so, because it's actually a catch-all for all the possible parallel behaviors that launcher won't support.

flow/project.py
csadorf: @zhou-pj Could you post the output of `module help launcher` on the stampede2 system? I don't have access to it.

zhou-pj: The output of module help launcher is very limited, but here is the README.md file for launcher on stampede2:
# Launcher
[Build Status](https://travis-ci.org/marshalllerner/launcher)
[status](http://joss.theoj.org/papers/7b5df63cd8a40f557d66051695d300a7)

Launcher is a utility for performing simple, data parallel, high throughput computing (HTC) workflows on clusters, massively parallel processor (MPP) systems, workgroups of computers, and personal machines.

## Installing Launcher
Launcher does not need to be compiled. Unpack the tarball or clone the repository in the desired directory. Then, set `LAUNCHER_DIR` to point to that location. Python 2.7 or greater and hwloc are required for full functionality. See INSTALL for more information.

## Verifying Installation

Included in the download is a file called "quickstart" found in the folder "tests". In order to verify installation, open the command line and find the launcher file, and then type "cd tests" and then press the enter key. If the quickstart file is in the correct place, there is no need for arguments, so type "./quickstart". However, if the Launcher directory is found somewhere else, type "./quickstart <Launcher directory>". The script will run in the terminal and if there are no errors in the process, the last line will say "Launcher: Done. Job exited without errors".

## Quickstart

* Set `LAUNCHER_JOB_FILE` to point to your job file. Example job files are provided in extras/examples.
* Be sure that `LAUNCHER_DIR` is set to the directory containing the launcher source files (user-installed ONLY. Not required if using system installed version of launcher).
* From the command-line or within your jobscript, run:`$LAUNCHER_DIR/paramrun`

## Available Environment Variables

You should set the following environment variables:

* `$LAUNCHER_JOB_FILE` is the file containing the jobs to run in your parametric submission.
* `$LAUNCHER_WORKDIR` is the directory where the launcher will execute. All relative paths will resolve to this directory.

The launcher defines the following environment variables for each job that is started:

* `$LAUNCHER_NPROCS` contains the number of processes running simultaneously in your parametric submission.
* `$LAUNCHER_NHOSTS` contains the number of hosts running simultaneously in your parametric submission.
* `$LAUNCHER_PPN` contains the number of processes per node.
* `$LAUNCHER_NJOBS` contains the number of jobs in your job file.
* `$LAUNCHER_TSK_ID` is the particular processing core that the job is running on, from 0 to `$LAUNCHER_NPROCS-1`.
* `$LAUNCHER_JID` represents the particular job instance currently running. `$LAUNCHER_JID` is numbered from 1 to `$LAUNCHER_NJOBS`.

Example: If you want to redirect stdout to a file containing the unique ID of each line, you can specify the following in the paramlist file: ```a.out > out.o$LAUNCHER_JID```

If this particular execution instance of a.out was the first line in the job file, the output would be placed in the file "out.o1".

Note: you can also use the launcher to run a sequence of serial jobs when you have more jobs to run than the requested number of processors.  

## Task Scheduling Behavior

The launcher has three available behaviors for scheduling jobs, available by setting the environment variable `$LAUNCHER_SCHED`: (descriptions below assume k = task, p = num. procs, n = num. jobs)

* dynamic (default) - each task k executes first available unclaimed line
* interleaved - each task k executes every (k+p)th line
* block - each task k executes lines [ k(n/p)+1, (k+1)(n/p) ]

## Using Launcher on Multi-/Many-core Processors
Launcher uses the hwloc utility to determine layout of cores on the node. If hwloc is installed on your system and the commands are in the default `PATH`, Launcher will use this to partition the cores on node between the tasks. You can enable task binding by setting `LAUNCHER_BIND=1` before calling `paramrun`.

## Using Launcher with Intel Xeon Phi (KNC) Co-processor Cards

Launcher has the ability to execute appropriately compiled executables natively on first generation Intel Xeon Phi (KNC) cards.

Available Environment Variables for Intel Xeon Phi execution:

* `$LAUNCHER_NPHI` is the number of Intel Xeon Phi cards per node. This is set to zero (0) by default. Acceptable values are '1' and '2'.
* `$LAUNCHER_PHI_PPN` is the number of processes per Intel Xeon Phi card.
* `$LAUNCHER_PHI_JOB_FILE` is the file containing the jobs to run on the Intel Xeon Phi cards.

## Job Submission

Copy the example job submission script `launcher.<sched>` to your working directory to use as a starting point for interfacing with the desired batch system. Note that this script provides some simple error checking prior to the actual submission to aid in diagnosing missing executables and misconfiguration.

The launcher/extras/batch-scripts directory contains several example submission scripts:
  * SGE:   launcher.sge
  * SLURM: launcher.slurm

## Referencing Launcher
If you are using Launcher, please remember to make a reference to it when publishing results. The file `paper/paper.bib` contains the BibTeX-formatted citation list. Please reference entry `Wilson:2014:LSF:2616498.2616533` (i.e., in LaTeX: `\cite{Wilson:2014:LSF:2616498.2616534}`).

vyasr: One implication is that we should probably include a note to cite Launcher in our documentation or so if you use this environment. Similar to whatever manner we use to chose to indicate that we should be cited (alternatively we can adopt something like HOOMD's citation model).

flow/templates/stampede2.sh
```diff
@@ -1,14 +1,30 @@
 {# Templated in accordance with: https://portal.tacc.utexas.edu/user-guides/stampede2 #}
 {% extends "slurm.sh" %}
+{% set ns = namespace(use_launcher=True) %}
+{% if operations|length() < 2 %}
+{% set ns.use_launcher = False %}
+{% endif %}
```
csadorf: I believe this should be an `{% else %}`, right?

zhou-pj: I think both logic gets the same result, now I am just checking each criteria one by one, it will do extra check even when operation.length=1 but given that I think it won't have significant slow down

csadorf: Ok, we can leave as is if you like.

Issue #92, opened 2019-03-18
PBS parse failure


mattwthompson:
### Description

When the output of a job status query (`squeue`, `qstat`, etc.) is completely empty instead of an empty node (`<Data></Data>`), the parser inside of `_fetch` fails and errors out.

### To reproduce

This happens on one of our clusters at Vanderbilt (running PBS 4.2.9) when no other users are running jobs, but the bug should reproducible whenever this behavior occurs:

```shell
$ echo $USER
mwt
$ qstat -fx -u $USER
$ qstat
```
Compared to the desired behavior (when somebody, not necessarily me, is running a job):
```
$ qstat -fx -u mwt # should be equivalent to qstat -fx -u $USER
<Data></Data>
```

I'm not sure why this happens, but it can be worked around via submitting a job that does anything, including sleeping:
```
$ qstat -fx -u $USER
$ qsub sleep.pbs 
39352.master.cl.vanderbilt.edu
$ qstat -fx -u $USER
[...]

PR #93, opened 2019-03-18
Store an empty node and warn when qstat returns an empty string

mattwthompson:
## Description

In at least one rare case, the call to `qstat` can return an empty string (`""`) instead of an empty node (`<Data></Data>`) which currently leads to a parser error. This attempts to catch that error and handle it gracefully.

## Motivation and Context

Attempts to fix the behavior reported in detail in #92. It's not clear to me how this sort of direct interaction with a scheduler is handled by tests (`grep -R squeue tests/` gave me nothing) so I have not added supporting tests.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

This pull request is based on
- [x] *master*, because it is a bug fix or an update to the documentation. (really just because I branch from `master` out of habit)

## Checklist:

If necessary:

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -18,6 +18,11 @@
             "affiliation": "University of Michigan",
             "name": "Timothy Moore",
             "orcid": "0000-0002-5709-7259"
+        },
+        {
+            "affiliation": "Vanderbilt University",
+            "name": "Matthew  W. Thompson",
+            "orcid": "0000-0002-1460-3983"
         }
     ],
     "creators": [
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -20,11 +20,6 @@
             "orcid": "0000-0002-5709-7259"
         },
         {
-            "affiliation": "Vanderbilt University",
-            "name": "Matthew  W. Thompson",
-            "orcid": "0000-0002-1460-3983"
-        },
-        {
             "affiliation": "University of Michigan",
             "name": "Rose Cersonsky",
             "orcid": "0000-0003-4515-3441"
@@ -32,6 +27,11 @@
         {
             "affiliation": "University of Michigan",
             "name": "Yuan Zhou"
+        },
+        {
+            "affiliation": "Vanderbilt University",
+            "name": "Matthew  W. Thompson",
+            "orcid": "0000-0002-1460-3983"
         }
     ],
     "creators": [
--- a/changelog.txt
+++ b/changelog.txt
@@ -8,6 +8,7 @@ next
 ----
 
 - Add function to automatically print all varying state point parameters in the detailed status view triggered by providing option `-p/--parameters` without arguments (#19, #87).
+- Fix issue with the TORQUE scheduler that occured when there was no job scheduled at all on the system (for any user) (#92, #93).
 
 Version 0.7
 ===========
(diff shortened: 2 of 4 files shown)


bdice:
@mattwthompson Were you able to test if this works? I am not sure that this will do the right thing but I have a suggested replacement at the bottom.

If you run a subprocess command that returns no output, the resulting `io.BytesIO` object is not an empty string, it's a file-like object with an empty buffer. You have to call `readlines()` to know if it has any content, which returns a list of line objects.
```python
>>> import subprocess, io
>>> result = io.BytesIO(subprocess.check_output(['touch', 'tmpfile']))
>>> result == ""
False
>>> result.readlines()
[]
```
Also, the syntax for XML parsing doesn't accept a string, but a file-like object such as `io.BytesIO`.
```python
>>> import xml.etree.ElementTree as ET
>>> ET.parse(source="<Data></Data>")
  File "<stdin>", line 1, in <module>
  File ".../lib/python3.6/xml/etree/ElementTree.py", line 1196, in parse
    tree.parse(source, parser)
  File ".../lib/python3.6/xml/etree/ElementTree.py", line 586, in parse
    source = open(source, "rb")
FileNotFoundError: [Errno 2] No such file or directory: '<Data></Data>'
```

Try something like this, which checks to see if the buffer is empty before attempting to parse:
```python
    if len(result.getbuffer()) == 0:
        warnings.warn(
            "No scheduler jobs, from any user(s), were detected. "
            "This may be the result of a misconfiguration in the "
            "environment.", UserWarning)
        result = io.BytesIO(b"<Data></Data>")
```

csadorf:
@mattwthompson I just saw @bdice comment after approving the PR. Could you comment on that? The patch might need some adjustments after all.

mattwthompson:
Good catch, it did not occur to me that I could just test this directly with my own setup. As you suspected, my approach did not work (my earlier changes just produce the same error as reported previously) but your suggested patch did the trick (note the new `UserWarning` buried in the rest of the mess!)

```
(mosdef) [mwt@master sfg-on-quartz]$ python project.py status --debug
/raid6/homes/mwt/software/anaconda/envs/mosdef/lib/python3.5/site-packages/MDAnalysis/due.py:88: UserWarning: No module named 'mock'
  warnings.warn(str(err))
/raid6/homes/mwt/software/signac-flow/flow/environment.py:352: DeprecatedWarning: add_args is deprecated as of 0.6 and will be removed in 0.8. This function is part of the legacy templating system.
  super(DefaultTorqueEnvironment, cls).add_args(parser)
/raid6/homes/mwt/software/signac-flow/flow/scheduling/torque.py:41: UserWarning: No scheduler jobs, from any user(s), were detected. This may be the result of a misconfiguration in the environment.
  "environment.", UserWarning)
Query scheduler...
/raid6/homes/mwt/software/signac-flow/flow/util/tqdm/_tqdm.py:65: DeprecationWarning: sys.getcheckinterval() and sys.setcheckinterval() are deprecated.  Use sys.setswitchinterval() instead.
  sys.setcheckinterval(100)
Collect job status info: 100%|████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 33554.43it/s]
# Overview:
Total # of jobs: 5

label               ratio
------------------  --------------------------------------------------
system_initialized  |########################################| 100.00%
nvt_done            |########################################| 100.00%
minimized           |########################################| 100.00%
```

PR #94, opened 2019-03-18
Feature/optimize status

vyasr:
## Description
This PR accelerates the status update by loading the cached status information only once rather than once per job. Additionally, the status data is immediately converted to a dictionary to avoid expensive calls to the various `_SyncedDict` functions that maintain the integrity of the underlying file. This conversion should be safe since this part of the status fetching is a read-only operation; the status information is written by the `_fetch_scheduler_status` method, but this change only affects the downstream `get_job_status` function calls.

There is one catch: because I'm passing a large dictionary as an argument to the `get_job_status` function there is additional overhead associated with serializing the data for parallelization. In fact, updating in parallel is actually slower than serial updates with the new implementation. As a result, the performance improvement of the new method is far more pronounced in serial than in parallel. There is probably a data size where the old method outperforms the new one if you benchmark parallel status updates only; I'd expect that to be about 50000 jobs based on the set of tests I've run so far. However, the new version running in serial might still be the fastest. I'm going to run a test on 1e5 data points once, just to be sure.

As a benchmark, I constructed test data spaces of 1000 and 10000 jobs and defined a FlowProject with a single operation with a simple file existence post-condition. I've attached the results in a text file, but the summary is that when status updates are performed in serial (passing the `--no-parallelize` option) the performance is now improved by a factor of almost 100. When parallelism is allowed, this advantage shrinks considerably but is still at least a factor of 5.

## Motivation and Context
For large data spaces (>1000 jobs) the current status update is far too slow. In particular, the collection of individual job status information takes extremely long because each individual job is accessing the `_status` key in the project document, which is an extremely large file for such large data spaces. 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

This pull request is based on

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -11,7 +11,7 @@ next
 Changed
 ++++++
 
-- The performance of status updates has been improved by a factor of up to 1000 for large data spaces by only pulling from the status cache once rather than once per operation (#94).
+- The performance of status updates has been significantly improved (up to a factor of 1,000 for large data spaces) by applying a more efficient caching strategy (#94).
 - Add clear environment notification when submitting job scripts.
 
 Version 0.7
--- a/changelog.txt
+++ b/changelog.txt
@@ -10,7 +10,7 @@ next
 Changed
 ++++++
 
-- Status updates only pull from the cache once rather than once per operation.
+- The performance of status updates has been improved by a factor of up to 1000 for large data spaces by only pulling from the status cache once rather than once per operation.
 
 
 Version 0.7
--- a/changelog.txt
+++ b/changelog.txt
@@ -4,6 +4,15 @@ Changes
 
 The **signac-flow** package follows `semantic versioning <https://semver.org/>`_.
 
+next
+----
+
+Changed
+++++++
+
+- Status updates only pull from the cache once rather than once per operation.
+
+
 Version 0.7
 ===========
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -11,7 +11,7 @@ next
 Changed
 ++++++
 
-- The performance of status updates has been improved by a factor of up to 1000 for large data spaces by only pulling from the status cache once rather than once per operation.
+- The performance of status updates has been improved by a factor of up to 1000 for large data spaces by only pulling from the status cache once rather than once per operation (#94).
 - Add clear environment notification when submitting job scripts.
 
 Version 0.7
--- a/flow/project.py
+++ b/flow/project.py
@@ -826,9 +826,10 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
                 expanded = JobOperation.expand_id(name)
                 yield expanded['job_id'], expanded['operation-name'], sjob
 
-    def _get_operations_status(self, job):
+    def _get_operations_status(self, job, cached_status=None):
         "Return a dict with information about job-operations for this job."
-        cached_status = self.document.get('_status', dict())
+        if cached_status is None:
+            cached_status = self.document.get('_status', dict())._as_dict()
         for job_op in self._job_operations([job], False):
             flow_op = self.operations[job_op.name]
             completed = flow_op.complete(job)
@@ -840,12 +841,12 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
                 'completed': completed,
             }
 
-    def get_job_status(self, job, ignore_errors=False):
+    def get_job_status(self, job, ignore_errors=False, cached_status=None):
         "Return a dict with detailed information about the status of a job."
         result = dict()
         result['job_id'] = str(job)
         try:
-            result['operations'] = OrderedDict(self._get_operations_status(job))
+            result['operations'] = OrderedDict(self._get_operations_status(job, cached_status))
             result['_operations_error'] = None
         except Exception as error:
             msg = "Error while getting operations status for job '{}': '{}'.".format(job, error)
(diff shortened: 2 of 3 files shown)


vyasr:
Changelog is not updated yet, I'll hold off until we're sure this is desirable.

flow/project.py
```diff
@@ -826,9 +826,16 @@ def _map_scheduler_jobs(scheduler_jobs):
                 expanded = JobOperation.expand_id(name)
                 yield expanded['job_id'], expanded['operation-name'], sjob
 
-    def _get_operations_status(self, job):
+    def _get_operations_status(self, job, cached_status=None):
```
csadorf: I don't think it's necessary to pass this argument along here. We can always assume that the *current* status is represented by `self.document._status`.

vyasr: The `cached_status = self.document.get('_status', dict())` in this function is actually the most expensive part, so passing it through is the main point of the optimization.

flow/project.py
```diff
@@ -826,9 +826,16 @@ def _map_scheduler_jobs(scheduler_jobs):
...
         "Return a dict with information about job-operations for this job."
-        cached_status = self.document.get('_status', dict())
+        if cached_status is None:
```
csadorf: To reduce complexity please convert `cached_status` to a required argument for this internal function.

flow/project.py
```diff
@@ -907,7 +914,9 @@ def _fetch_scheduler_status(self, jobs=None, file=None, ignore_errors=False):
             scheduler_info = {sjob.name(): sjob.status() for sjob in self.scheduler_jobs(scheduler)}
             status = dict()
             print(self._tr("Query scheduler..."), file=file)
-            for op in self.next_operations(*jobs):
+            for op in tqdm(self.next_operations(*jobs),
```
csadorf: Is the additional progress bar insightful for the user or did you primarily use it during debugging?

flow/project.py
```diff
@@ -934,7 +943,15 @@ def _print_progress(x):
                     err.flush()
                 yield _
 
-        _get_job_status = functools.partial(self.get_job_status, ignore_errors=ignore_errors)
+        cached_status = self.document.get('_status', dict())
```
csadorf: To reduce complexity, I propose the following change:
```
try:
    cached_status = self.document['_status']._as_dict()
except KeyError:
    cached_status = dict()
```

vyasr: NIce. I was trying to find a more idiomatic way to write this but didn't move away from calling `get. This is better.

flow/project.py
vyasr: I created it for debugging, but I actually do think that it's helpful. However, the utility is dependent on the data space being quite large; for <1000 jobs this step is typically instant. I left it in intentionally, but if you are concerned that it's too verbose have a look at the sample output and we can make a call.

csadorf: We can leave it in if you think that it is helpful not a distraction.

vyasr:
Looks like I was wrong. Even for 100,000 jobs, the new version is substantially faster. Without investigating the exact time spent in each function too deeply, it looks like reading the status from the project document that many times in the loop is prohibitively expensive and the gains from parallelism in the old version get washed out. The nice thing is that when the data space is this large parallelism starts to help with the new version as well, so if the space gets even larger it's possible that using parallelism would be worthwhile.

As it stands, I'm not sure if you want to change the default behavior or not given backwards compatibility concerns. We could add a `--parallel` flag and deprecate the old flag, since it looks like in all realistic cases parallelism is now wasted on this function.

[Benchmarks.pdf](https://github.com/glotzerlab/signac-flow/files/2979871/Benchmarks.pdf)

csadorf:
> As it stands, I'm not sure if you want to change the default behavior or not given backwards compatibility concerns. We could add a `--parallel` flag and deprecate the old flag, since it looks like in all realistic cases parallelism is now wasted on this function.

I believe that parallelism will generally still be beneficial under normal circumstances. You have only one super cheap condition function in your benchmark, which might exaggerate the shift of the bottleneck towards serialization.

I suggest that instead of abandoning parallelism, we optimize it for the new requirements.

vyasr:
That's a fair point. For now, let's not change anything then. As I make use of this in my production project, I can evaluate further optimizations in context.

vyasr:
@csadorf I added a test and updated the changelog to reflect this change. Once tests pass this should be ready to merge.

csadorf:
Consider to execute the test with / and without project caching and skip the former on the older signac versions that do not support caching.

Looks like there is some Py27 incompatibility as well.

tests/test_project.py
```diff
@@ -146,6 +146,41 @@ def mock_project(self):
         return project
 
 
+@unittest.skipIf(six.PY2, 'Only check performance on Python 3')
+class ProjectStatusTest(BaseProjectTest):
```
csadorf: Are there ay other tests that we could potentially move into this test case? Otherwise I'd suggest to rename to `ProjectStatusPerformanceTest`.

vyasr: Yes I had that originally and accidentally changed that. I'll switch it back.

vyasr:
OK all set.

changelog.txt
```diff
@@ -4,6 +4,15 @@ Changes
 
 The **signac-flow** package follows `semantic versioning <https://semver.org/>`_.
 
+next
+----
+
+Changed
+++++++
+
+- Status updates only pull from the cache once rather than once per operation.
```
csadorf: Could you rephrase along the lines of "The performance of X has been significantly improved (factor X) by optimizing Y"?

Btw, no need to skip CI tests.

csadorf: And add the PR # in parentheses.

csadorf:
Cool! Almost done!

PR #101, opened 2019-04-02
Fix summit multigpu

jglaser:
## Description
Multi-GPU jobs on summit (i.e. one MPI process with multiple GPUs) are incorrectly handled. This PR is supposed to fix that.

## Motivation and Context
Current behavior for e.g. ngpu=3*nranks is to calculate using the greatest common denominator, ending up with partially filled nodes. E.g.

```
jsrun -n 38 -a 1 -c 1 -g 5 -d packed -b r
```
is clearly not desirable (it leaves one GPU idle, and the number of MPI processes is incorrect for 75 jobs with 3 GPUs each). Without a way to specify the number of GPUs per resource set explicitly, we need to do a smarter guess.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

This pull request is based on

`feature/enable-aggregate-operations` which should be merged first

## Checklist:

perhaps need to add a new test

If necessary:


jglaser:
Simon and I discussed this off-line, and we believe we have figured out the general way to do this. The assumption is that either ngpu=nrank*g or np=ngpu*n, where n and g are integers. Here is what we intend to implement: 

IMG_1425

vyasr:
@jglaser have you had a chance to implement the proposed fix for the calculation?

jglaser:
> @jglaser have you had a chance to implement the proposed fix for the calculation?

I did now. I had to update the test reference data to always assign the `np` directive to the same resource set, so `parallel_op` should generate

```
jsrun -n 1 -a 1 -c 3 ...
```
instead of
```
jsrun -n 3 -a 1 -c 1
```

My understanding is that `np`, as opposed to `ntasks` refers to parallelism per MPI task, such as multithreading, and therefore needs to be part of the same resource set.

csadorf:
I am very sorry, I won't be able to thoroughly review and test this before the end of this week, but it still on my agenda.

Issue #103, opened 2019-04-29
Capture scheduler error messages and print to user


justinGilmer:
### Feature description

When the user runs `submit` for a system with a scheduler management system, any error messages from the scheduler command itself should be captured and displayed to the user.

### Proposed solution

If `subprocess.py` can capture the STDOUT and STDERR of the command it calls, and returns that to the user when calling `submit` that would drastically improve the debugging experience. 

### Additional context

Just today I was submitting some jobs to a `torque` system, however I did not realize that I had over-allocated the maximum hours I could allocate. 

The error message mentioned that there was an error with `qsub` and the temporary file that was the batch script. It also mentioned that it returned error code 255, unfortunately that is a generic error code. If `subprocess` can capture the STDOUT and STDERR of `qsub` or the slurm equivalent, that would be excellent in debugging.

vyasr:
I agree that this would be a good idea. I'll try to have a look at it.

justinGilmer:
Here is the selection of code from [torque.py](https://github.com/glotzerlab/signac-flow/blob/fc7916fae3d9e10bf9828a4b296ce93321e86e61/flow/scheduling/torque.py#L141) where I think the check should be done
```python3
        if pretend:
            print("# Submit command: {}".format(' '.join(submit_cmd)))
            print(script)
            print()
        else:
            with tempfile.NamedTemporaryFile() as tmp_submit_script:
                tmp_submit_script.write(str(script).encode('utf-8'))
                tmp_submit_script.flush()
                output = subprocess.check_output(
                    submit_cmd + [tmp_submit_script.name])
            jobsid = output.decode('utf-8').strip()
return jobsid
```

What I think should be changed:
```python3
        if pretend:
            print("# Submit command: {}".format(' '.join(submit_cmd)))
            print(script)
            print()
        else:
            with tempfile.NamedTemporaryFile() as tmp_submit_script:
                tmp_submit_script.write(str(script).encode('utf-8'))
                tmp_submit_script.flush()
				try:
                	output = subprocess.check_output(
                    	submit_cmd + [tmp_submit_script.name])
					jobsid = output.decode('utf-8').strip()
				except subprocess.CalledProcessError as e:
					print(e.output)
return jobsid
```
All that I changed was wrapping the `subprocesss` call in a try,except block.

I think all that is required is capturing the exception raised and then printing that out. However, I feel like this will be a lot of code replicating for each scheduler file.

csadorf:
[...]

PR #104, opened 2019-05-03
Initial exception handling of scheduler errors

justinGilmer:
## Description

In `torque.py`, when the call is made using the `subprocess` module,
if the scheduler returns an error that message is not passed on to the
user. The error code is, but usually that is not enough to debug what
could be the cause of that issue.

This PR attempts to catch that exception and provide it to the user.

## Motivation and Context
Debugging scheduler submissions can be difficult when only provided the error code.
This change will provide the users the full output from the `subprocess` module
as a way to more easily determine what in their script was incorrect.

This is related to issue #103 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

This pull request is based on
- [x] *master*, because it is a bug fix or an update to the documentation.

## Checklist:

If necessary:

### DIFF
--- a/flow/scheduling/torque.py
+++ b/flow/scheduling/torque.py
@@ -16,6 +16,7 @@ import xml.etree.ElementTree as ET
 
 from .base import Scheduler
 from .base import ClusterJob, JobStatus
+from ..errors import SubmitError
 
 
 logger = logging.getLogger(__name__)
@@ -146,9 +147,12 @@ class TorqueScheduler(Scheduler):
             with tempfile.NamedTemporaryFile() as tmp_submit_script:
                 tmp_submit_script.write(str(script).encode('utf-8'))
                 tmp_submit_script.flush()
-                output = subprocess.check_output(
-                    submit_cmd + [tmp_submit_script.name])
-            jobsid = output.decode('utf-8').strip()
+                try:
+                    output = subprocess.check_output(
+                        submit_cmd + [tmp_submit_script.name])
+                    jobsid = output.decode('utf-8').strip()
+                except subprocess.CalledProcessError as e:
+                    raise SubmitError("qsub error: {}".format(e.output()))
             return jobsid
 
     @classmethod
--- a/changelog.txt
+++ b/changelog.txt
@@ -8,6 +8,7 @@ next
 ----
 
 - Add full launcher support for job submission on XSEDE Stampede2 for large parallel single processor jobs (#85, #91).
+- Show submission error messages in combination with a TORQUE scheduler (#103, #104).
 
 Version 0.7
 ===========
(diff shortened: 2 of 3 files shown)


justinGilmer:
Not entirely sure where these additional commits came from? Assuming those are just the changes that are on master now?

csadorf:
> Not entirely sure where these additional commits came from? Assuming those are just the changes that are on master now?

We recently switched the branching model to "OneFlow", so all development goes directly into master (see [here](https://docs.signac.io/en/latest/community.html#contributions) for more info). That's no problem though, I just switched the base of the PR.

flow/scheduling/torque.py
```diff
@@ -146,9 +146,12 @@ def submit(self, script, after=None, pretend=False, hold=False, flags=None, *arg
...
                 tmp_submit_script.flush()
-                output = subprocess.check_output(
-                    submit_cmd + [tmp_submit_script.name])
-            jobsid = output.decode('utf-8').strip()
+                try:
+                    output = subprocess.check_output(
+                        submit_cmd + [tmp_submit_script.name])
+                    jobsid = output.decode('utf-8').strip()
+                except subprocess.CalledProcessError as e:
+                    print(e.output())
```
csadorf: Reviewing this, I realized that we already implement the desired behavior for the SLURM scheduler, just not for TORQUE. Thank you very much for fixing this!

Can you update this line to raise a `SubmitError` similar to how we do this for the SLURM scheduler?
https://github.com/glotzerlab/signac-flow/blob/0685611bfa370ab47166e80449e6de6aaece443f/flow/scheduling/slurm.py#L138

In this case it should say "qsub" instead of "sbatch".

justinGilmer:
Ah ok that makes sense. I just raised an issue about that. #105

justinGilmer:
This is still a very early draft of this PR, but before I get too deep into the changes, how exactly would you want to test for this? I guess if `simplescheduler` produces an error message and we check for that output, that should be a decent test?

csadorf:
> This is still a very early draft of this PR, but before I get too deep into the changes, how exactly would you want to test for this? I guess if `simplescheduler` produces an error message and we check for that output, that should be a decent test?

I'm ok with not testing this as part of the integrated testing, but relying on a "manual" tests. @vyasr Any suggestions?

justinGilmer:
I am a bit unfamiliar with rebasing, normally just merge and squash when necessary.

Let me know if that rebase is what you wanted!

csadorf:
> I am a bit unfamiliar with rebasing, normally just merge and squash when necessary.
> 
> Let me know if that rebase is what you wanted!

Looks fine to me!

vyasr:
Sorry I was unresponsive this week, my plan was to have a look over the weekend but it looks like  @justinGilmer beat me to it. Thanks for the contribution! @csadorf I agree that manual testing in this case is sufficient.

Issue #67, opened 2019-02-26
Use jinja2 templating for the status output formatting


csadorf:
### Feature description

We should switch to using `jinja2` to format the status output. This would not only simplify the implementation, but would also allow users to design their own status output for individual projects.

### Proposed solution

Just like submission templates, there would be a default status template which can be replaced, extended by placing a `status.txt` file in a project's `templates/` directory.

zhou-pj:
Start working on this issue now. Just want to make sure no one is also working on it.

tcmoore3:
Resolved with PR #111

PR #109, opened 2019-05-06
Show legend in all detailed-views.

csadorf:
Actually show legend in all detailed-status views.

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -1328,8 +1328,8 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
                                 sched_stat = sched_stat.decode('utf-8')
                             msg = u"{} {} [{}]".format(frame, name, sched_stat)
                             _print_unicode(msg)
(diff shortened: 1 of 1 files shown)


bdice:
@csadorf I might have misunderstood the problem. I thought `[S]` was not showing up because this line was removed. https://github.com/glotzerlab/signac-flow/pull/68/files#diff-896014ecae7b1581f5b837d124cdb4c9L771

csadorf:
> @csadorf I might have misunderstood the problem. I thought `[S]` was not showing up because this line was removed. https://github.com/glotzerlab/signac-flow/pull/68/files#diff-896014ecae7b1581f5b837d124cdb4c9L771

I think there is some confusion here, because we use varies tables with various symbols. I don't have the mental capacity to sort this out right now.

csadorf:
It might be worth it to wait until @zhou-pj has completed his work to address issue #67 .

vyasr:
Yes, I was on the same page as Bradley regarding the problem here. @zhou-pj is making good progress, though, so I agree that we can maybe just punt on fixing this until he completes his work

Issue #107, opened 2019-05-05
Feature proposal: Integrated profiling


csadorf:
### Feature description

I suggest that we implement an integrated profiling tool to make it easier to optimize the operations-graph and speed up function such as `status`.

### Proposed solution

I suggest that we add a `--profile` option to the `FlowProject` `status` command that executes the status update, but then provides the user with profiling information about which condition functions are responsible for the majority of the runtime.

### Additional context

Even without this feature, users can profile their status function like this:

```
python -m cProfile -s time project.py status --no-parallelize
```

However, it is kind of tedious to need to remember this command and the output is not always intuitive. I'd prefer to analyze the profiling results and provide a more meaningful report on which condition function is responsible for major delay and why.

In addition, the standard profiler will not provide any meaningful result unless the `--no-parallelize` option is provided, however there are profilers such as [yappi](https://github.com/sumerc/yappi) that would work even in the parallelized status compilation.

Finally, we could do some additional "meta"-profiling such as analyzing the number of required `os.stat` calls and their performance and then provide some recommendations on a reasonable data space size for that specific file system.

csadorf:
@justinGilmer Do you think that would be useful?

justinGilmer:
@csadorf Very much so! I normally use tools like [vprof](https://github.com/nvdv/vprof) when I am trying to profile, but having a dedicated profiling option that will parse the information and give a nicer summary will be very welcome!

csadorf:
Great! Concerning vprof: I had looked at various profiling tools, but didn't come across this one yet. Looks really cool! Unfortunately we might not be able to use it for this specific case, because it has npm requirements.

justinGilmer:
Yeah I do not think vprof is the best option here, its quite useful for personal projects though!

I do think that parsing some profiler output and presenting that to the user would be super helpful, im not entirely sure what the best option is for that, however.

csadorf:
Resolved  with merge of #110 .

PR #110, opened 2019-05-07
Implement optional statistical profiling for status command.

csadorf:
Implements integrated statistical profiler for status updates to help identify code paths that cause a slow status update, such as expensive condition functions.

Addresses issue #107 .

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -2489,11 +2489,11 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
             if not args['profile'] and delta_t > warn_threshold >= 0:
                 print(
                     "WARNING: "
-                    "The status compilation took more than {}ms per job. Consider to "
+                    "The status compilation took more than {}s per job. Consider to "
                     "use `--profile` to determine bottlenecks within your project "
                     "workflow definition.\n"
                     "Execute `signac config set flow.{} VALUE` to specify the "
-                    "warning threshold in ms. Use -1 to completely suppress this "
+                    "warning threshold in seconds. Use -1 to completely suppress this "
                     "warning."
                     .format(warn_threshold, config_key), file=sys.stderr)
 
--- a/flow/project.py
+++ b/flow/project.py
@@ -2486,6 +2486,7 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
             delta_t = (time.time() - start) / len(jobs)
             config_key = 'status_performance_warn_threshold'
             warn_threshold = flow_config.get_config_value(config_key)
+            warn_threshold = 0.2 if warn_threshold is None else warn_threshold  # signac 0.9.0
             if not args['profile'] and delta_t > warn_threshold >= 0:
                 print(
                     "WARNING: "
--- a/flow/project.py
+++ b/flow/project.py
@@ -1217,11 +1217,10 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
 
             # Unrestricted
             total_impact = 0
-            total_num_hits = sum([hit[2] for fn, ft in prof.merged_file_dict.items()
-                                  if fn not in fn_filter for hit in ft.iterHits()])
             hits = [hit for fn, ft in prof.merged_file_dict.items()
                     if fn not in fn_filter for hit in ft.iterHits()]
             sorted_hits = reversed(sorted(hits, key=lambda hit: hit[2]))
+            total_num_hits = sum([hit[2] for hit in hits])
 
             profiling_results = [
                 '# Profiling:\n',
--- a/flow/project.py
+++ b/flow/project.py
@@ -22,6 +22,7 @@ import re
 import logging
 import warnings
 import argparse
+import time
 import datetime
 import json
 import inspect
@@ -2469,6 +2470,7 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
         if args.pop('full'):
             args['detailed'] = args['all_ops'] = True
 
+        start = time.time()
         try:
             self.print_status(jobs=jobs, **args)
         except NoSchedulerError:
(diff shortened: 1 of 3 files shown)


csadorf:
@justinGilmer This is my first take at implementing this function. I implemented the profiling with [pprofile](https://github.com/vpelletier/pprofile) and use a multithreading-compatible statistical profiler.

I think the major challenge here is to determine what to show and how. I'm generally sorting by the number of hits and filter out a few results which will always be somewhat dominant (related to process and thread management), but that do not provide any further insight in how to optimize the code.

I am limiting the results to the hits that are responsible for 80% of total hits (following the Pareto principle) and further limit the results to the top 10 (whichever is smaller).

I think a major drawback right now is that it's not telling us specifically about what lines within the `project.py` file should be optimized.

csadorf:
I added some more information for the file that defines the `FlowProject` class, usually called `project.py`.

flow/project.py
```diff
@@ -1087,7 +1089,86 @@ def print_status(self, jobs=None, overview=True, overview_max_lines=None,
...
+            total_num_hits = sum([hit[2] for fn, ft in prof.merged_file_dict.items()
+                                  if fn not in fn_filter for hit in ft.iterHits()])
+            hits = [hit for fn, ft in prof.merged_file_dict.items()
+                    if fn not in fn_filter for hit in ft.iterHits()]
+            sorted_hits = reversed(sorted(hits, key=lambda hit: hit[2]))
+
+            profiling_results = [
+                '# Profiling:\n',
+                'Rank Impact Code object',
+                '---- ------ -----------']
```
bdice: Could/should we jinja-ify this?

csadorf: Yes, I'll rebase this once #111 is merged.

csadorf:
Per request (@bdice ), here is a screen-grab from my example-project:
```bash
$ python project.py status --show-traceback --profile
Using environment configuration: UnknownEnvironment
Collect job status info: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 9/9 [00:03<00:00,  1.36it/s]
# Overview:
Total # of jobs: 9

label    ratio
-------  -------
[no labels to show]

# Profiling:

Rank Impact Code object
---- ------ -----------
   1    46% /Users/csadorf/my_project/util.py:5:expensive_function
   2    26% /Users/csadorf/local/signac-flow/flow/util/tqdm/_tqdm.py:87:run
   3    13% project.py:20:expensive_condition

Hits by line for 'project.py':
-------------------------------
project.py:21 (49%):
    0   20: def expensive_condition(job):
  650   21:     sleep(FACTOR * (0.5 + random.random() / 2))
    0   22:     return True

project.py:26 (29%):
    0   25: def second_expensive_condition(job):
  389   26:     sleep(FACTOR * (0.5 + random.random() / 2))
    0   27:     return True

project.py:38 (20%):
    0   36: def another_expensive_condition(job):
    9   37:     with open(os.devnull, 'w') as file:
  275   38:         x = [print('.', end='', file=file) for _ in range(int(random.random() * 1e6))]
    0   39:     return True

Total runtime: 3s
Warning: Profiler ran only for a short time, results may be highly inaccurate.
```

justinGilmer:
So far this is excellent! ill toy around with this later this week with some of my larger systems!

csadorf:
@justinGilmer Did you get a chance to test this? I'm planning on finishing this up soon.

justinGilmer:
@csadorf It appears to be working!

However, I am not getting a `hits by line` summary at the bottom.

I guess it does support though, that my major issue was the amount of labels I have for my project. Since ~`42%` of the impact is from `isfile`.

However, it is not very intuitive looking at the image below. I was expecting to see the `isfile` method contained in `signac`.

image

csadorf:
> However, I am not getting a `hits by line` summary at the bottom.
> 
That might mean that your implementation in `src/project.py` is not actually limiting the status generation or it's simply a bug. Could you add some kind of "expensive function" to your `src/project.py` to verify that?

> I guess it does support though, that my major issue was the amount of labels I have for my project. Since ~`42%` of the impact is from `isfile`.

> However, it is not very intuitive looking at the image below. I was expecting to see the `isfile` method contained in `signac`.

The upper overview is showing which function directly is causing the slow-down. One would need to optimize that specific function to speed up the status. It's not the `Job.isfile()` function that presents the bottleneck, it's actually the underlying built-in function that would need to be optimized. Do you have a suggestion how we could present that better?

bdice:
I read through the code changes. I would approve this PR as-is but the proposed jinja rewrite would merit a second pass of review. Consider this comment my approval, if you decide to merge it as-is.

csadorf:
> I read through the code changes. I would approve this PR as-is but the proposed jinja rewrite would merit a second pass of review. Consider this comment my approval, if you decide to merge it as-is.

@bdice Rebased this on `master`, which included the jinja rewrite.

flow/project.py
```diff
@@ -1174,7 +1176,87 @@ def highlight(s, eligible):
...
+            ]
+
+            with prof(single=False):
+                tmp = self._fetch_status(jobs, err, ignore_errors, no_parallelize)
+
+            prof._mergeFileTiming()
+
+            # Unrestricted
+            total_impact = 0
+            total_num_hits = sum([hit[2] for fn, ft in prof.merged_file_dict.items()
```
b-butler: I think it would be cleaner to define hits first and then do something like
total_num_hits = `sum([hit[2] for hit in hits])`

flow/project.py
```diff
@@ -2519,6 +2605,12 @@ class MyProject(FlowProject):
             'status',
             parents=[base_parser])
         self._add_print_status_args(parser_status)
+        parser_status.add_argument(
```
b-butler: Adding a way to specify that we should not exclude `signac-flow`'s own code would be helpful for developers and users who may want to avoid certain functions that could be more costly for whatever reason. This would also help us find pressure points in the code.

flow/project.py
```diff
@@ -1174,7 +1176,87 @@ def highlight(s, eligible):
...
+                profiling_results.append(
+                    "{rank:>4} {impact:>6.0%} {code.co_filename}:"
+                    "{code.co_firstlineno}:{code.co_name}".format(
+                        rank=i+1, impact=impact, code=code))
+                if i > 10 or total_impact > 0.8:
+                    break
+
+            # Restricted to the 'self'-module:
+
+            self_module = inspect.getmodule(self)
```
b-butler: We should consider adding `signac-flow`'s own code to the profile output.

csadorf:
@b-butler I've implemented your change request.

csadorf:
> LGTM sorry for the late review.

@b-butler No worries. I've added an automated warning that recommends users to use `--profile` when the status compilation takes quite long. The current (configurable) threshold is set to more than 0.2 seconds.

csadorf:
@bdice This PR should be good to go now.

csadorf:
> Thanks for the reminder to review, @csadorf -- looks great! 0.2 seconds seems like a very small computation time for the warning. I would think 2 (or even 5) seconds is more appropriate, but I don't think that's a blocking issue. We can change it later if we find it to be annoying.

This value is *per job*. You really think that 5mins for the update would be appropriate for a project with 60 jobs?

csadorf:
There is some issue with the signac 0.9.0 compatibility. Trying to fix it right now.

bdice:
@csadorf I missed that the time was per-job. That's sensible. 👍 Maybe the configuration key name should imply that?

csadorf:
It says "per job" in the message. The key gets too long IMO.

PR #111, opened 2019-05-07
Issue 67 use jinja2 templating to print status

zhou-pj:
## Description

Use jinja2 templating to print status

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


zhou-pj:
Currently not all the existing options are properly implemented in jinja template yet. But I have make majority of them working and I think it would be a good place to create a temporary pull request for some code review and suggestions. @csadorf @vyasr 
Also might be helpful to discuss all the current options and discuss if there is anything we want to change or improve before I implement all of them.

Currently are the formatting logic are implemented in a single jinja template, except the --expand option is implemented in an extended template.

Current options implemented:
-d, --detailed   
-a, --all-operations
--only-incomplete-operations
-e, --expand
--pretty
--full
--no-overview

not implemented:
-1, --one-line
--stack

csadorf:
Great! I'm going to test that out as soon as possible!

csadorf:
I pushed one fix for a bug when there are no labels defined, but I still get this error for my
[test project](https://github.com/glotzerlab/signac-flow/files/3159713/project.py.zip):
```
  File "project.py", line 66, in <module>
    Project().main()
  File "/Users/csadorf/local/signac-flow/flow/project.py", line 2496, in main
    _exit_or_raise()
  File "/Users/csadorf/local/signac-flow/flow/project.py", line 2464, in main
    args.func(args)
  File "/Users/csadorf/local/signac-flow/flow/project.py", line 2144, in _main_status
    self.print_status(jobs=jobs, **args)
  File "/Users/csadorf/local/signac-flow/flow/project.py", line 1131, in print_status
    print(template.render(** context), file=file)
  File "/Users/csadorf/miniconda3/envs/signac-flow-dev/lib/python3.7/site-packages/jinja2/asyncsupport.py", line 76, in render
    return original_render(self, *args, **kwargs)
  File "/Users/csadorf/miniconda3/envs/signac-flow-dev/lib/python3.7/site-packages/jinja2/environment.py", line 1008, in render
    return self.environment.handle_exception(exc_info, True)
  File "/Users/csadorf/miniconda3/envs/signac-flow-dev/lib/python3.7/site-packages/jinja2/environment.py", line 780, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/csadorf/miniconda3/envs/signac-flow-dev/lib/python3.7/site-packages/jinja2/_compat.py", line 37, in reraise
    raise value.with_traceback(tb)
  File "/Users/csadorf/local/signac-flow/flow/templates/print_status.sh", line 22, in top-level template code
    {% block body %}
  File "/Users/csadorf/local/signac-flow/flow/templates/print_status.sh", line 27, in block "body"
    {{ format_progress_bar | format('label','ratio')}}
  File "/Users/csadorf/miniconda3/envs/signac-flow-dev/lib/python3.7/site-packages/jinja2/filters.py", line 685, in do_format
    return soft_unicode(value) % (kwargs or args)
TypeError: not all arguments converted during string formatting
```

zhou-pj:
@csadorf  Thanks for pointing that out. I realize I forgot to test the most simple case and the bug is because I put on variable for overview inside the {if detail} section. The changes I pushed should now fix the issue.

csadorf:
The status view works great now for all the projects I tested it with. I think it would be good now to streamline this patch a little bit, especially with an emphasis on making the template script as readable as possible.

For this, I'm going to provide some feedback and add a few suggestions to the code.

flow/project.py
```diff
@@ -1117,68 +1040,30 @@ def _incomplete(s):
             print(json.dumps(statuses, indent=4), file=file)
             return
 
+        jobs_count = len(statuses)
```
csadorf: Could you try to break up the gathering of all the various template parameters into logical sections and use comments to guide the reader? This line for instance, seems to come a bit out of nowhere.

flow/project.py
```diff
@@ -1190,153 +1075,64 @@ def get(k, m):
...
-                print(next(status_table_lines), file=file)
-                print(next(status_table_lines), file=file)
-
-                def _print_unicode(value):
-                    "Python 2/3 compatibility layer."
-                    if six.PY2:
-                        print(value.encode('utf-8'), file=file)
-                    else:
-                        print(value, file=file)
+        if template is None:
```
csadorf: I think this section (lines 1087-1095) should come right after line 1042, because it helps the reader understand what is happening next "gathering of the context for these templates".

flow/project.py
```diff
@@ -1190,153 +1075,64 @@ def get(k, m):
...
-                legend = u'Legend: ' + u' '.join(u'{}:{}'.format(v, k) for k, v in symbols.items())
-                _print_unicode(legend)
-            print(' '.join('[{}]:{}'.format(v, k) for k, v in self.ALIASES.items()))
-
-        # Show any abbreviations used
-        if abbreviate.table:
-            print('\n', self._tr("Abbreviations used:"), file=file)
-            for a in sorted(abbreviate.table):
-                print('{}: {}'.format(a, abbreviate.table[a]), file=file)
+        bar_length = 50
```
csadorf: I'd suggest to use `width` instead of `length` here. I think that would convey the meaning in the context of formatting better.

flow/project.py
```diff
@@ -1190,153 +1075,64 @@ def get(k, m):
...
+        id_length = 32
+        label_length = 5
+        total_label_length = 0
+        operation_length = 0
+        operations_count = 0
```
csadorf: I'd propose the terms `column_width_X` instead of `X_length` and `num_X` instead of `X_count` to better convey the meaning in the context of on-screen formatting. In this case I'd argue that being as explicit is possible is preferable, we can always shorten later if we believe that the names are too verbose.

flow/templates/print_status.sh
```diff
@@ -0,0 +1,55 @@
...
+{% set ns.format_parameters = ns.format_parameters + ("%%-%ss  " | format(parameters_length[loop.index-1],)) %}
+{% endfor %}
+{% set para_head = ns.format_parameters | format(*parameters,) %}
+{% endif %}
+{% set format_head = format_job_id + format_operation_title + '%s' + format_label %}
+{% set format_row = format_job_id + format_operation + '%s' + format_label %}
+{% endif %}
+{% block body %}
+{% if overview %}
+# Overview:
```
csadorf: Can we try to move the natural blocks of the status view into corresponding jinja blocks?

I'd roughly expect the following block structure:
```jinja
{% block overview %}
{% block progress %}
...
{% endblock %}
...
{% endblock %}
{% block detailed %}
...
{% endblock %}
# etc.
```
That would make it *much* easier for users to override, or extend those individual blocks without the need to replace the whole template.

flow/templates/print_status.sh
```diff
@@ -0,0 +1,55 @@
+{# The following variables are available to all scripts. #}
```
csadorf: Since the status view is not a script, I propose to use the file suffix `.jinja` or `.txt` instead of `.sh`.

flow/templates/print_status.sh
```diff
@@ -0,0 +1,55 @@
...
+{% endfor %}
+{% set para_head = ns.format_parameters | format(*parameters,) %}
+{% endif %}
+{% set format_head = format_job_id + format_operation_title + '%s' + format_label %}
+{% set format_row = format_job_id + format_operation + '%s' + format_label %}
+{% endif %}
+{% block body %}
+{% if overview %}
+# Overview:
+Total # of jobs: {{'%s' |format(jobs_count)}}
```
csadorf: I think it would be more natural to use the `len` filter here: `{{ jobs | len %}}`.

That should work, right?

flow/templates/print_status.sh
```diff
@@ -0,0 +1,55 @@
...
+{% if detailed %}
+
+# Detailed View:
+{{ format_head | format('job_id', 'operation', para_head, 'labels',) }}
+{{ format_head | format('-'*id_length, '-'*(operation_length+4), ns.dash, '-'*total_label_length,) }}
+{% for job in jobs %}
+{% if parameters %}
+{% set para_output = ns.format_parameters | format(*job['parameters'].values()) %}
+{% endif %}
+{% set ns.first_row = True %}
```
csadorf: Jinja has some tricks to handle this kind of stuff. For example, you can replace this variable and use `loop.first`, see here: http://jinja.pocoo.org/docs/2.10/templates/#for

zhou-pj:
Thanks for you suggestions! I've made changes addressing all of them. Let me know if the changes are on the correct direction and I can keep working on the few format options left.

zhou-pj:
I have now implemented all the options in status view. I tried to mimic the exact format with all the options in current state but had to make some minor changes for easier jinja2 implementation or improved table alignment. We can discuss those changes and options logics for possible improvements once the current code looks good.

csadorf:
Awesome! I'll try to provide a second thorough review by the end of today.

flow/templates/print_status.jinja
```diff
@@ -0,0 +1,64 @@
...
+{% set ns.format_parameters = ns.format_parameters + ("%%-%ss  " | format(column_width_parameters[loop.index-1],)) %}
+{% endfor %}
+{% set para_head = ns.format_parameters | format(*parameters,) %}
+{% endif %}
+{% set format_head = format_job_id + format_operation_title + '%s' + format_label %}
+{% endif %}
+{% block overview %}
+{% if overview %}
+# Overview:
+Total # of jobs: {{ '%s' |format(jobs | length()) }}
```
csadorf: I don't think you need the parenthesis for `length()` when you don't pass any arguments.

flow/templates/print_status_stack.jinja
```diff
@@ -0,0 +1,21 @@
...
+{% for key,value in job['operations'].items() %}
+{% if (alias[value['scheduler_status']] != 'U' or value['eligible']) or all_ops%}
+{{ format_operation | bold_font(value['eligible'],pretty) | format(value | get_operation_status(pretty),key,'['+alias[value['scheduler_status']]+']',) }}
+{% endif %}
+{% endfor %}
+{% endfor %}
+{% if pretty %}
+Legend: ○:ineligible ●:eligible ▹:active ▸:running ✔:completed
+{% else %}
+Legend: -:ineligible +:eligible *:active >:running X:completed
```
csadorf: I think it would be better to have a `{{ legend }}` variable that we pre-populate.

flow/templates/print_status_stack.jinja
```diff
@@ -0,0 +1,21 @@
...
+{% set para_output = ns.format_parameters | format(*job['parameters'].values()) %}
+{% endif %}
+{{ format_job_id | format(job['job_id'],) }}{{para_output}}{{ '%s' | format(job['labels']|join(', ')) }}
+{% for key,value in job['operations'].items() %}
+{% if (alias[value['scheduler_status']] != 'U' or value['eligible']) or all_ops%}
+{{ format_operation | bold_font(value['eligible'],pretty) | format(value | get_operation_status(pretty),key,'['+alias[value['scheduler_status']]+']',) }}
+{% endif %}
+{% endfor %}
+{% endfor %}
+{% if pretty %}
```
csadorf: I propose to move the pretty logic out of the template and pass different symbols to the template instead.

flow/templates/print_status.jinja
```diff
@@ -0,0 +1,64 @@
+{# The following variables are available to all scripts. #}
+{% if overview %}
+{% set format_progress_bar = ("%%-%ss  %%-%ss" | format(column_width_label,column_width_bar,)) %}
```
bdice: Minor change - `flake8` requires _spaces after commas_ in the Python code, and I think it would be good to style this template's logic in the same way. Also I would drop the last comma if it's not required.

flow/templates/print_status.jinja
```diff
@@ -0,0 +1,64 @@
...
+{% set alias = scheduler_status_code %}
+{% set column_width_operation = [column_width_operation,5] | max %}
+{% set format_job_id = ("%%-%ss  " | format(column_width_id,)) %}
+{% set format_operation = ("%%-%ss %%3s  " | format(column_width_operation,)) %}
+{% set format_operation_title = ("%%-%ss  " | format(column_width_operation+4,)) %}
+{% set format_label = ("%%-%ss" | format(column_width_total_label,)) %}
+{% set ns = namespace(dash='', format_parameters='') %}
+{% if parameters %}
+{% for para in parameters %}
+{% set ns.dash = ns.dash + ("%s  " | format('-' * column_width_parameters[loop.index-1])) %}
```
bdice: Use `loop.index0` instead of `loop.index-1`. Documented here: http://jinja.pocoo.org/docs/2.10/templates/#for

flow/templates/print_status.jinja
```diff
@@ -0,0 +1,64 @@
...
+{% set ns.format_parameters = ns.format_parameters + ("%%-%ss  " | format(column_width_parameters[loop.index-1],)) %}
```
bdice: Can you write `a += b` instead of `a = a + b`?

flow/templates/print_status.jinja
bdice: I would also put a space on both sides of the pipe `|`, everywhere in this template.

flow/templates/print_status.jinja
```diff
@@ -0,0 +1,64 @@
...
+{% set format_head = format_job_id + format_operation_title + '%s' + format_label %}
+{% endif %}
+{% block overview %}
+{% if overview %}
+# Overview:
+Total # of jobs: {{ '%s' |format(jobs | length()) }}
+
+{% block progress %}
+{{ format_progress_bar | format('label','ratio') }}
+{{ format_progress_bar | format('-'*column_width_label,'-'*column_width_bar) }}
```
bdice: I'd put spaces on both sides of the `*` for readability.

flow/templates/print_status.jinja
```diff
@@ -0,0 +1,64 @@
...
+{% for label in progress_sorted %}
+{{ format_progress_bar | format(label[0], label[1] | draw_progressbar((jobs | length()))) }}
```
bdice: You might have too many parentheses here - try `draw_progressbar(jobs | length)`.

flow/templates/print_status.jinja
```diff
@@ -0,0 +1,64 @@
...
+{{format_job_id | format('',)}}{{ format_operation | bold_font(value['eligible'],pretty) | format(key,'['+alias[value['scheduler_status']]+']',) }}{{para_output}}{{ '%s' | format(job['labels']|join(', ')) }}
+{% endif %}
+{% endif %}
+{% endfor %}
+{% endfor %}
+{% endif %}
+{% endblock %}
+{% block legend %}
+{% if detailed %}
+[U]:unknown [R]:registered [Q]:queued [A]:active [I]:inactive [!]:requires_attention
```
bdice: Is there a way to get these programmatically? If there's a map of letters to descriptions elsewhere (I think there might be, but I'm not sure), let's avoid recreating it by hand here.

flow/templates/print_status_stack.jinja
bdice: ^ similar to what I proposed above. Perhaps there's a convenient way to combine these and just replace the symbols? I might be wrong - it doesn't look like there is a one-to-one mapping with the set above, so maybe I'm confused about what this is doing.

flow/templates/print_status.sh
zhou-pj: Sorry I did made the change but had to revert that later and forgot to post it here. With the current implementation we still need the ns.first_row since this is not print differently for the first operation within the loop but the first qualified one in the loop. But based on what you suggested today if we have a seperate filter logic before the table printing then we may then use loop.first.

flow/templates/print_status.jinja
zhou-pj: I tried but I think jinja2 does not support this

flow/project.py
```diff
@@ -1117,68 +1093,32 @@ def _incomplete(s):
...
-                    print(self._tr("Lines omitted:"), lines_skipped, file=file)
-
-        # Generate detailed view:
-        def _select_op(doc):
-            active = JobStatus(doc['scheduler_status']) > JobStatus.unknown
-            if skip_active and active:
-                return False
-            if not all_ops and not (active or doc['eligible']):
-                return False
-            return True
```
zhou-pj: How is skip_active option passed to the function? the status command don't have this option and I am confused what will trigger skip_active

csadorf: The `--skip-active` option is active, but not shown in the help function, probably because it is supposed to be deprecated, but we forgot to put a deprecation warning into the code.
https://github.com/glotzerlab/signac-flow/blob/f0b696c11a41700ae0bc50f608d5a7a1b10a442e/flow/project.py#L2021-L2024

flow/templates/print_status.jinja
zhou-pj: Now legend are formatted in python and passed to template. Does that look better?

flow/templates/print_status_stack.jinja
zhou-pj: I now premade the legend string in python and moved the if pretty in python and pass the proper legend dictionary to template.

flow/templates/print_status.jinja
```diff
@@ -0,0 +1,59 @@
...
+{% else %}
+{{field_job_id | format('')}}{{ field_operation | bold_font(value['eligible'], pretty) | format(key, '['+scheduler_status_code[value['scheduler_status']]+']') }}{{para_output}}{{ '%s' | format(job['labels'] | join(', ')) }}
+{% endif %}
+{% endfor %}
+{% endfor %}
+{% endif %}
+{% endblock %}
+{% block legend %}
+{% if detailed %}
+{{ status_legend }}
```
csadorf: I'd just move that right above L53, no need to put this into a separate block.

flow/project.py
```diff
@@ -1190,153 +1135,83 @@ def get(k, m):
...
-            print(' '.join('[{}]:{}'.format(v, k) for k, v in self.ALIASES.items()))
-
-        # Show any abbreviations used
-        if abbreviate.table:
-            print('\n', self._tr("Abbreviations used:"), file=file)
-            for a in sorted(abbreviate.table):
-                print('{}: {}'.format(a, abbreviate.table[a]), file=file)
+        print(template.render(**context), file=file)
+
+        return None
```
csadorf: The two lines above are not necessary.

flow/util/template_filters.py
```diff
@@ -172,6 +173,81 @@ def print_warning(msg):
     return ''
 
 
+def bold_font(s, eligible, pretty=False):
```
csadorf: I suggest we rename this filter-function to `strong()` and remove any logic from it. The filter should always return text that is presented "strongly" (bold). The logic on _when_ to use the filter should be in the template.

flow/util/template_filters.py
```diff
@@ -172,6 +173,81 @@ def print_warning(msg):
...
+        int
+    """
+
+    assert value >= 0 and total > 0
+    ratio = ' %0.2f%%' % (100 * value / total)
+    n = int(value / total * width)
+    return '|' + ''.join(['#'] * n) + ''.join(['-'] * (width - n)) + '|' + ratio
+
+
+def job_filter(job_op, scheduler_status_code, all_ops):
```
csadorf: This filter-function should be defined locally in the `print_status()` function.

flow/util/template_filters.py
```diff
@@ -172,6 +173,81 @@ def print_warning(msg):
...
+    """
+
+    if scheduler_status_code[job_op['scheduler_status']] != 'U' or job_op['eligible'] or all_ops:
+        return True
+    else:
+        return False
+
+
+def get_operation_status(operation_info, symbols):
+    """Visualize progess with a progress bar.
```
csadorf: This doc-string summary does not seem to describe the purpose of this function.

flow/project.py
zhou-pj: I am not sure about this, which two line do you mean?

csadorf: Lines 1213 and 1214.

flow/project.py
```diff
@@ -541,6 +539,7 @@ def _setup_template_environment(self):
         template_environment = jinja2.Environment(
             loader=jinja2.ChoiceLoader(load_envs),
             trim_blocks=True,
+            lstrip_blocks=True,
```
zhou-pj: I found that adding this could allow jinja2 to ignore indent in all logic codes. I think this could work for the status print template but I want to check with you @csadorf if that won't affect the job submission script as this change is global.

flow/templates/print_status.jinja
```diff
@@ -0,0 +1,56 @@
...
+        {% endif %}
+        {% for key, value in job['operations'].items() if value | job_filter(scheduler_status_code, all_ops) %}
+            {% if loop.first %}
+{{ field_job_id | format(job['job_id']) }}{{ field_operation | highlight(value['eligible']) | format(key, '['+scheduler_status_code[value['scheduler_status']]+']') }}{{para_output}}{{ '%s' | format(job['labels'] | join(', ')) }}
+            {% else %}
+{{ field_job_id | format('')}}{{ field_operation | highlight(value['eligible']) | format(key, '['+scheduler_status_code[value['scheduler_status']]+']') }}{{para_output}}{{ '%s' | format(job['labels'] | join(', ')) }}
+            {% endif %}
+        {% endfor %}
+    {% endfor %}
+{{ status_legend }}
```
zhou-pj: I currently tried adding indent in only this template so we could see if it looks good and I can change all once we finalized a proper format. Currently only logic codes are indented({% %}), (output lines ({{ }}) are not, I had some issue make them work properly)

zhou-pj: @bdice Would like your input on what's the best way to properly format the template code.

flow/project.py
csadorf: I'd prefer to not the change the global template context to avoid breaking backwards compatibility.

zhou-pj: I removed this line

bdice:
@csadorf @zhou-pj If the templates are difficult to indent, don't worry about it. I've found it useful to have indentation in my templates once I played with whitespace for a while, but it's probably not important enough to spend further time on it for now.

zhou-pj:
I have undone the indent and added some minor improvements to the template. Let me know if there is other suggestions on the template I should change except for the indentation. Thanks!

PR #112, opened 2019-05-10
Make progressbar on a per-job basis

### LINKED ISSUE #108: More intuitive progress bar
### Feature description
The progress bar in `flow` should reflect the progress of each statepoint, not the total operations it has to collect and run.

### Proposed solution
The progress bar should be a function of the number of statepoints and the bar should only increment once all the labels/statuses have been checked and applied to that statepoint.

### Additional context

When showing the progress bar when running any flow command `status`, etc. on versions of `flow` >= 0.7.X, the progress bar will overflow past the # of statepoints due to label checking, etc. This is different compared to previous behavior, as the label checking appeared to be a part of the progress bar (it appeared that all the label checking was done per statepoint before incrementing the progress bar.)

I think this should be the preferred way to present this information to the user, since the progress bar loses meaning if it overflows.

This was an issue that I ran into with my systems which had a lot of labels, the progress bar would fill in a few seconds, but then it would spend an additional 10-30 seconds still counting iterations for the various labels, etc. This was quite confusing and made it difficult for me to determine what the issue was.

vyasr:
This pull request fixes the progress bar drawn when fetching status information.

## Description
As of the last release of flow, I added an additional progressbar to the scheduler status fetching because that step can be slow when there are large numbers of operations. However, I had set the total on that progressbar to be the number of jobs, not the number of job-operations, so the available progressbar was exhausted before the iterations completed. This PR fixes that problem by counting up to the number of jobs.  @justinGilmer let me know if this fixes your problem.

There is a catch to this; because the iterable is a generator that filters out ineligible operations, we don't know the total number *a priori*. To fix this issue, I replaced the original logic with two nested loops, one over jobs and one over the operations for each job. I'm not sure what a better solution to this would be, but I'm open to suggestions.

Related, the history for the last set of changes is a little confusing. I see two different commit entries for the PR merge that added this progressbar, e590d70de073bc4e466baf03ab749c60f3c0c02a and 564417c0303c4418adea73fc945e1e260367b899. In 564417c0, the relevant progressbar line uses `next_operations`, which is what I recall putting there and what is shown in Pull Request #94, but e590d70d matches the current master, which directly calls `_job_operations(jobs=jobs, only_eligible=False)`. I'm not sure how or why this change was made, but I don't think it's necessary; we shouldn't need to update the scheduler status for ineligible jobs. Any idea what could have caused this? I'd like to fix, but since I don't recall making this change perhaps someone else had a reason for doing so.

## Motivation and Context
Closes #108

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -831,7 +831,7 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
 
     def _get_operations_status(self, job, cached_status):
         "Return a dict with information about job-operations for this job."
-        for job_op in self._job_operations([job], False):
+        for job_op in self._job_operations(job, False):
             flow_op = self.operations[job_op.name]
             completed = flow_op.complete(job)
             eligible = False if completed else flow_op.eligible(job)
@@ -917,7 +917,7 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
             for job in tqdm(jobs,
                             desc="Fetching operation status",
                             total=len(jobs), file=file):
-                for op in self._job_operations(jobs=[job], only_eligible=False):
+                for op in self._job_operations(job, only_eligible=False):
                     status[op.get_id()] = int(scheduler_info.get(op.get_id(), JobStatus.unknown))
             self.document._status.update(status)
         except NoSchedulerError:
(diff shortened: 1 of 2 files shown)


flow/project.py
```diff
@@ -914,10 +914,11 @@ def _fetch_scheduler_status(self, jobs=None, file=None, ignore_errors=False):
...
             status = dict()
             print(self._tr("Query scheduler..."), file=file)
-            for op in tqdm(self._job_operations(jobs=jobs, only_eligible=False),
-                           desc="Fetching operation status",
-                           total=len(jobs), file=file):
-                status[op.get_id()] = int(scheduler_info.get(op.get_id(), JobStatus.unknown))
+            for job in tqdm(jobs,
+                            desc="Fetching operation status",
+                            total=len(jobs), file=file):
+                for op in self._job_operations(jobs=[job], only_eligible=False):
```
csadorf: Can you refactor the `_job_operations()` function to accept a single `job` argument instead of a sequence of jobs? I believe that the function accepts a sequence right now to prepare us for aggregation-operations, but the loop-structure as implemented right now simply represents an anti-pattern.

No need to accommodate for backwards-compatibility, since this is a purely internal function.

vyasr: Agreed, I was considering that as a possible solution and i think it's the right call.

PR #117, opened 2019-06-12
Fix issue where active operations are resubmitted.

tcmoore3:
Operations are resubmitted although already active on the scheduler.

This fix patches a bug introduced with pull request #89.


csadorf:
@glotzerlab/signac-developers Would any of you be available to implement a unit test for this bug fix on short notice? Otherwise I propose to merge as is to fix this as soon as possible.

PR #118, opened 2019-06-19
Fix default np directive when callable directives are provided

vyasr:
## Description
I fixed a bug that occurs when the `nranks` or `omp_num_threads` directives are specified using callables. I'm open to suggestions on how to make this less verbose. One alternative is:

```bash
if callable(nranks) or callable(omp_num_threads):
    np_callable = lambda job: (nranks(job) if callable(nranks) else nranks) * (omp_num_threads(job) if callable(omp_num_threads) else omp_num_threads)
    directives.setdefault('np', np_callable )
else:
    directives.setdefault('np', nranks*omp_num_threads )
```

but I find that pretty unappealing because it makes the lambda function complicated enough that I don't find it more readable. If others strongly prefer it then we could do that, though.

## Motivation and Context
The `np` directive is set based on the `nranks` and `omp_num_threads` directives. However, if either of these are callables, the current default setting for `np` will not work; we need to define it as a callable as well. Without this fix, these submissions fail.

## Types of Changes

- [x] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -244,13 +244,17 @@ class JobOperation(object):
         keys_set_by_user = set(directives.keys())
 
         nranks = directives.get('nranks', 1)
-        omp_num_threads = directives.get('omp_num_threads', 1)
+        nthreads = directives.get('omp_num_threads', 1)
 
-        if callable(nranks) or callable(omp_num_threads):
-            np_callable = lambda job: (nranks(job) if callable(nranks) else nranks) * (
-                    omp_num_threads(job) if callable(omp_num_threads) else omp_num_threads)
+        if callable(nranks) or callable(nthreads):
+            def np_callable(job):
+                nr = nranks(job) if callable(nranks) else nranks
+                nt = nthreads(job) if callable(nthreads) else nthreads
+                return nr*nt
+
+            directives.setdefault('np', np_callable)
         else:
-            directives.setdefault('np', nranks*omp_num_threads )
+            directives.setdefault('np', nranks*nthreads)
 
         directives.setdefault('ngpu', 0)
         directives.setdefault('nranks', 0)
--- a/flow/project.py
+++ b/flow/project.py
@@ -245,19 +245,12 @@ class JobOperation(object):
 
         nranks = directives.get('nranks', 1)
         omp_num_threads = directives.get('omp_num_threads', 1)
-        if callable(nranks):
-            if callable(omp_num_threads):
-                directives.setdefault(
-                    'np', lambda job: nranks(job)*omp_num_threads(job) )
-            else:
-                directives.setdefault(
-                    'np', lambda job: nranks(job)*omp_num_threads )
-        elif callable(omp_num_threads):
-            directives.setdefault(
-                'np', lambda job: nranks*omp_num_threads(job) )
+
+        if callable(nranks) or callable(omp_num_threads):
+            np_callable = lambda job: (nranks(job) if callable(nranks) else nranks) * (
+                    omp_num_threads(job) if callable(omp_num_threads) else omp_num_threads)
         else:
-            directives.setdefault(
-                'np', lambda job: nranks*omp_num_threads )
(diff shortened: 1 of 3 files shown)


bdice:
@vyasr I find the alternative you suggested (one big lambda) more readable. Even though the lambda function is more complicated by itself, it's significantly less logic in total.

Testing probably requires updates to the template testing framework - I'm not sure if we're testing callable directives at all currently, I couldn't find any such tests.

flow/project.py
```diff
@@ -243,8 +243,22 @@ def __init__(self, name, job, cmd, directives=None, np=None):
...
+                    'np', lambda job: nranks(job)*omp_num_threads(job))
+            else:
+                directives.setdefault(
+                    'np', lambda job: nranks(job)*omp_num_threads)
+        elif callable(omp_num_threads):
+            directives.setdefault(
+                'np', lambda job: nranks*omp_num_threads(job))
+        else:
+            directives.setdefault(
+                'np', lambda job: nranks*omp_num_threads)
```
csadorf: Why does it have to be a callable in that case?

csadorf:
I have one question about this patch, otherwise it's good.

flow/project.py
vyasr: It doesn't, I can change that. What are your thoughts regarding the options above? If you agree with Bradley I'll make that change as well, otherwise I'll just change this else clause to a non-callable case.

tcmoore3:
I'm chiming in since I'm also running into this issue now. Is there a reason you can't just `evaluate()` `directives['nranks']` and `directives['omp_num_threads']` before setting `np`?

tcmoore3:
Also, as far as testing is concerned, we wrote a test for callable directives [here](https://github.com/glotzerlab/signac-flow/blob/master/tests/test_project.py#L397).

tcmoore3:
> I'm chiming in since I'm also running into this issue now. Is there a reason you can't just `evaluate()` `directives['nranks']` and `directives['omp_num_threads']` before setting `np`?

Ah, that won't work since the default values for `nranks` and `omp_num_threads` are both 0, so if you don't pass one of those `np` also gets set to 0.

flow/project.py
vyasr: @csadorf @mikemhenry thoughts?

vyasr:
@tcmoore3 so what you're suggesting is to use `evaluate` for `np` to reassign to the `directives` dictionary before the finally dict comprehension that generates the directives? I think that would work too, I don't have a preference for which solution to use.

flow/project.py
csadorf: I'd prefer the alternative, because I don't like the amount of nested if-clauses here.

tcmoore3:
@vyasr Yes exactly. Just to avoid the nested ifs.

flow/project.py
vyasr: Sounds good, I'll switch to what I pasted above. That version also fixes the callable issue.

flow/project.py
```diff
@@ -243,8 +243,19 @@ def __init__(self, name, job, cmd, directives=None, np=None):
...
+
+        if callable(nranks) or callable(nthreads):
+            def np_callable(job):
+                nr = nranks(job) if callable(nranks) else nranks
+                nt = nthreads(job) if callable(nthreads) else nthreads
+                return nr*nt
+
+            directives.setdefault('np', np_callable)
+        else:
+            directives.setdefault('np', nranks*nthreads)
```
csadorf: I'm surprised that the linter did not complain about the missing spaces around `*`.

csadorf:
@tcmoore3 I've approved this PR, but I'd appreciate if you could also review and test this patch.

vyasr:
Added documentation in glotzerlab/signac-docs#34 and updated the changelog.

csadorf:
@vyasr Thank you for fixing this! 
@tcmoore3 Thank you for reviewing and adding additional unit tests!

PR #120, opened 2019-07-02
Enable copy_from and after conditions with multiple arguments.

csadorf:
## Description

Enable the provision of multiple operation functions as argument to `pre.after()` and `pre/post.copy_from()` 

## Motivation and Context

When generating operation functions programmatically, it is much more convenient to store those in a list and then we can use those lists as argument to the condition functions. Otherwise we cannot use the decorator syntax.

### Minimal example:
```python
def setup_foo(i):

    @Project.operation(f'foo-{i}')
    @Project.post.true(f'{foo_{i}')
    def foo(job):
        print(f"{i}th foo!")
        job.doc[f'foo_{i}'] = True

    return foo

foos = [setup_foo(i) for i in range(3)]

@Project.operation
@Project.pre.after(*foos)
def bar(job):
    print('all foos ran!')
```

**Without this feature** we would need to awkwardly call `Project.pre.after` as a function manually:
```python
for foo in foos:
    Project.pre.after(foo)(bar)
```

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -11,6 +11,7 @@ next
 - Show submission error messages in combination with a TORQUE scheduler (#103, #104).
 - Fix issue that caused the "Fetching operation status" progressbar to be inaccurate (#108).
 - Rewrite print status function using Jinja2 templates.
+- Allow the provision of multiple operation-functions to the `pre.after` condition (#120).
 
 Version 0.7
 ===========
--- a/changelog.txt
+++ b/changelog.txt
@@ -11,7 +11,7 @@ next
 - Show submission error messages in combination with a TORQUE scheduler (#103, #104).
 - Fix issue that caused the "Fetching operation status" progressbar to be inaccurate (#108).
 - Rewrite print status function using Jinja2 templates.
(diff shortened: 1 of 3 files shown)


csadorf:
@mikemhenry @vyasr Any chance you could review this PR soon? Or should I try to find someone else?

vyasr:
I can review, I started going through all the issues/PRs associated with me yesterday and will get around to this one sometime today.

csadorf:
> Nice feature. One note, depending on the sequence of the workflow you could end up evaluating conditions multiple times if they're conditions for multiple conditions passed to these decorators. However, not sure that there's a whole lot to be done about that, and it shouldn't matter except for massive projects.

I agree and I've created issue #125 to keep track of that.

csadorf:
> Is it possible to add a unit test for this feature?

I've added a unit test to specifically test this feature.

mikemhenry:
LGTM :tada:

csadorf:
Thank's for the help @vyasr @mikemhenry @b-butler !

PR #121, opened 2019-07-03
Optionally shuffle operations prior to execution pass.

csadorf:
## Description

Optionally shuffle operations prior to execution pass with the `run --shuffle` command line flag or `shuffle` argument to `FlowProject.run()`.

## Motivation and Context

Shuffling may serve multiple purposes:

 * Speed up execution for parallelized runs, where operations that
   operate on the same job may block eachother.
 * Simplify debugging since errors that may occur later in the workflow
   are more likely to occur a bit earlier.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -1491,6 +1491,10 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
             Show a progress bar during execution.
         :type progess:
             bool
+        :param shuffle:
+            Shuffle operations prior to each execution pass.
+        :type shuffle:
+            bool
         """
         # If no jobs argument is provided, we run operations for all jobs.
         if jobs is None:
--- a/flow/project.py
+++ b/flow/project.py
@@ -1439,7 +1439,7 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
             fork(cmd=operation.cmd, timeout=timeout)
 
     def run(self, jobs=None, names=None, pretend=False, np=None, timeout=None, num=None,
-            num_passes=1, progress=False, shuffle=False):
+            num_passes=1, progress=False, order=None):
         """Execute all pending operations for the given selection.
 
         This function will run in an infinite loop until all pending operations
@@ -1491,11 +1491,16 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
             Show a progress bar during execution.
         :type progess:
             bool
-        :param shuffle:
-            Shuffle operations prior to each execution pass.
-        :type shuffle:
-            bool
+        :param order:
+            Specify the order of operations, possible values are:
+                * None or 'none' (no reordering, operations are grouped by job)
+                * 'round-robin' (order operations in round-robin fashion by job)
+                * 'random'    (shuffle the execution order randomly)
+                * callable specify a key-callable by which to order operations
+            The default value is None.
         """
+        from itertools import groupby
+        from .util.misc import roundrobin
         # If no jobs argument is provided, we run operations for all jobs.
         if jobs is None:
             jobs = self
@@ -1570,8 +1575,23 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
                     del messages[:]     # clear
             if not operations:
                 break   # No more pending operations or execution limits reached.
-            if shuffle:
+
+            # Optionally re-order operations for execution if order argument is provided:
+            if callable(order):
+                operations = list(sorted(operations, key=order))
+            elif order == 'round-robin':
+                groups = [list(group)
+                          for _, group in groupby(operations, key=lambda op: op.job)]
+                operations = list(roundrobin(*groups))
+            elif order == 'random':
                 random.shuffle(operations)
+            elif order is None or order == 'none':
+                pass  # no reordering
+            else:
+                raise ValueError(
+                    "Invalid value for 'order' argument, valid arguments are "
+                    "None, 'none', 'round-robin', or 'random'.")
+
             logger.info(
                 "Executing {} operation(s) (Pass # {:02d})...".format(len(operations), i_pass))
             self.run_operations(operations, pretend=pretend,
@@ -2373,7 +2393,7 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
                                 jobs=jobs, names=args.operation_name, pretend=args.pretend,
                                 np=args.parallel, timeout=args.timeout, num=args.num,
                                 num_passes=args.num_passes, progress=args.progress,
-                                shuffle=args.shuffle)
+                                order=args.order)
 
         if args.switch_to_project_root:
             with add_cwd_to_environment_pythonpath():
@@ -2584,9 +2604,11 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
             help="Specify the number of cores to parallelize to. Defaults to all available "
                  "processing units if argument is ommitted.")
         execution_group.add_argument(
-            '-s', '--shuffle',
-            action='store_true',
-            help="Shuffle the order of operations prior to each execution pass.")
+            '--order',
+            type=str,
+            choices=['none', 'round-robin', 'random'],
+            default=None,
+            help="Specify the execution order of operations for each execution pass.")
         parser_run.set_defaults(func=self._main_run)
 
         parser_script = subparsers.add_parser(
--- a/flow/project.py
+++ b/flow/project.py
@@ -1499,7 +1499,7 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
                 * 'by-job' (operations are grouped by job)
                 * 'cyclic' (order operations cyclic by job)
                 * 'random' (shuffle the execution order randomly)
-                * callable (a callable returning a a comparison key for an
+                * callable (a callable returning a comparison key for an
                             operation used to sort operations)
 
             The default value is `none`, which is equivalent to `by-job` in the current
--- a/flow/project.py
+++ b/flow/project.py
@@ -31,6 +31,7 @@ from collections import defaultdict
 from collections import OrderedDict
 from itertools import islice
 from itertools import count
+from itertools import groupby
 from hashlib import sha1
 from multiprocessing import Pool
 from multiprocessing import cpu_count
@@ -57,6 +58,7 @@ from .errors import TemplateError
 from .util.tqdm import tqdm
 from .util.misc import _positive_int
 from .util.misc import _mkdir_p
+from .util.misc import roundrobin
 from .util import template_filters as tf
 from .util.misc import add_cwd_to_environment_pythonpath
 from .util.misc import switch_to_directory
@@ -1512,8 +1514,6 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
         :type order:
             str, callable, or NoneType
         """
-        from itertools import groupby
-        from .util.misc import roundrobin
         # If no jobs argument is provided, we run operations for all jobs.
         if jobs is None:
             jobs = self
--- a/flow/project.py
+++ b/flow/project.py
@@ -1493,7 +1493,8 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
             bool
         :param order:
             Specify the order of operations, possible values are:
-                * None or 'none' (no reordering, operations are grouped by job)
+                * None or 'none' (no reordering, do not assume any order)
+                * 'by-job'  (operations are grouped by job)
                 * 'round-robin' (order operations in round-robin fashion by job)
                 * 'random'    (shuffle the execution order randomly)
                 * callable specify a key-callable by which to order operations
@@ -1585,12 +1586,12 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
                 operations = list(roundrobin(*groups))
             elif order == 'random':
                 random.shuffle(operations)
-            elif order is None or order == 'none':
-                pass  # no reordering
+            elif order is None or order in ('none', 'by-job'):
+                pass  # by-job is the default order
             else:
                 raise ValueError(
                     "Invalid value for 'order' argument, valid arguments are "
-                    "None, 'none', 'round-robin', or 'random'.")
+                    "None, 'none', 'by-job', 'round-robin', or 'random'.")
 
             logger.info(
                 "Executing {} operation(s) (Pass # {:02d})...".format(len(operations), i_pass))
@@ -2606,7 +2607,7 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
         execution_group.add_argument(
             '--order',
             type=str,
-            choices=['none', 'round-robin', 'random'],
+            choices=['none', 'by-job', 'round-robin', 'random'],
             default=None,
             help="Specify the execution order of operations for each execution pass.")
         parser_run.set_defaults(func=self._main_run)
--- a/flow/project.py
+++ b/flow/project.py
@@ -1498,8 +1498,9 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
                 * 'none' or None (no specific order)
                 * 'by-job' (operations are grouped by job)
                 * 'cyclic' (order operations cyclic by job)
-                * 'random' (explicitly shuffle the execution order randomly)
-                * callable (specify a key-callable used to sort operations)
+                * 'random' (shuffle the execution order randomly)
+                * callable (a callable returning a a comparison key for an
+                            operation used to sort operations)
 
             The default value is `none`, which is equivalent to `by-job` in the current
             implementation.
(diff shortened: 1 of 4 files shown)


csadorf:
Alternatively we could just shuffle by default, what do you think?

b-butler:
With groups coming up, I am not sure of the best choice. Shuffling by default seems like a good option with the caveat it may require that more use cases pass a `--num-passes=2` option to the group.

vyasr:
I'm fine with using this randomization, but we should probably also change the default operation order to sort things by operation to maximize the chances that different operations on the same job will not be run in parallel. That change should also minimize any parallel file access issues that occur when multiple operations acting on the same job execute simultaneously.

b-butler:
@vyasr I think that is a good idea as well.

csadorf:
I've adjusted the implementation to actually be a bit more flexible, I think the only remaining question is whether we want to change the default behavior. Possible downside is just the overhead for rearrangement, but I'd opt for making 'round-robin' the default.

flow/project.py
```diff
@@ -1491,11 +1491,17 @@ def run(self, jobs=None, names=None, pretend=False, np=None, timeout=None, num=N
...
+        :param order:
+            Specify the order of operations, possible values are:
+                * None or 'none' (no reordering, do not assume any order)
+                * 'by-job'  (operations are grouped by job)
+                * 'round-robin' (order operations in round-robin fashion by job)
+                * 'random'    (shuffle the execution order randomly)
+                * callable specify a key-callable by which to order operations
+            The default value is None.
         """
+        from itertools import groupby
```
b-butler: I am not sure why the imports are going here. If the purpose is conditional importing, shouldn't it be further down in the if statements? Otherwise, I think it should go to the top with the other imports as is typical python form.

flow/project.py
```diff
@@ -1439,7 +1439,7 @@ def _fork(self, operation, timeout=None):
             fork(cmd=operation.cmd, timeout=timeout)
 
     def run(self, jobs=None, names=None, pretend=False, np=None, timeout=None, num=None,
-            num_passes=1, progress=False, shuffle=False):
+            num_passes=1, progress=False, order=None):
```
b-butler: You say in a comment that job ordering is done by default. Unless that changes, I think the default value should be `'by_job'`. Strings are not mutable in python so this would be a fine default value, and would more clear specify to users in the documentation that default ordering is `by-job`.

flow/project.py
```diff
@@ -1570,8 +1576,23 @@ def select(operation):
...
+            if callable(order):
+                operations = list(sorted(operations, key=order))
+            elif order == 'round-robin':
+                groups = [list(group)
+                          for _, group in groupby(operations, key=lambda op: op.job)]
+                operations = list(roundrobin(*groups))
+            elif order == 'random':
                 random.shuffle(operations)
+            elif order is None or order in ('none', 'by-job'):
+                pass  # by-job is the default order
```
b-butler: See comment above.

flow/project.py
```diff
@@ -2584,9 +2605,11 @@ class MyProject(FlowProject):
             help="Specify the number of cores to parallelize to. Defaults to all available "
                  "processing units if argument is ommitted.")
         execution_group.add_argument(
-            '-s', '--shuffle',
-            action='store_true',
-            help="Shuffle the order of operations prior to each execution pass.")
+            '--order',
+            type=str,
+            choices=['none', 'by-job', 'round-robin', 'random'],
+            default=None,
```
b-butler: I think default should be `'by-job'` for previously specified reasons.

flow/project.py
csadorf: Making the default value 'by-job' instead of `None` means that we are guaranteeing a specific order. If ever wanted to change that, then we need to go through a whole deprecation cycle. It's much safer to make no guarantees about the order by default to avoid future incompatibilities. The fact that `none` and `by-job` happen to be the same one for our specific implementation is therefore essentially of no relevance with respect to the default value to this argument.

There is one more reason why I'd prefer to have `None` be the default: Our execution model is generally based on the idea of a state-machine, where the definition order and execution order of operations should in principle not matter. The only reason I am introducing this option in the first place is to allow to speed up the execution under specific circumstances.

csadorf:
> I've adjusted the implementation to actually be a bit more flexible, I think the only remaining question is whether we want to change the default behavior. Possible downside is just the overhead for rearrangement, but I'd opt for making 'round-robin' the default.

Having thought about this a bit longer, I am now advising to leave the default oder at `none`, i.e., no specific order, reasons are outlined in my comment to @b-butler abover.

flow/project.py
```diff
@@ -1490,7 +1491,17 @@ def run(self, jobs=None, names=None, pretend=False, np=None, timeout=None, num=N
             Show a progress bar during execution.
         :type progess:
             bool
+        :param order:
```
bdice: Does it get a type docstring, `str`?

flow/project.py
```diff
@@ -1490,7 +1491,17 @@ def run(self, jobs=None, names=None, pretend=False, np=None, timeout=None, num=N
...
+            Specify the order of operations, possible values are:
+                * None or 'none' (no reordering, do not assume any order)
```
bdice: I'd prefer to specify what this does instead of what it does not do. Something like "uses order as returned by (some function)" or some definition from the FlowProject. I'm not sure how this works internally, but I find the current docstring unclear and would want to search the source code for clarification of what "no assumed order" actually means.

flow/project.py
```diff
@@ -1490,7 +1491,17 @@ def run(self, jobs=None, names=None, pretend=False, np=None, timeout=None, num=N
...
+                * 'by-job'  (operations are grouped by job)
```
bdice: Spacing of these lines is weird - it looks like they were meant to be aligned at one point. I recommend keeping it consistent at one space between the string `'by-job'` and the description `(operations are grouped by job)`, for all elements of this list.

flow/project.py
```diff
@@ -1490,7 +1491,17 @@ def run(self, jobs=None, names=None, pretend=False, np=None, timeout=None, num=N
...
+                * 'round-robin' (order operations in round-robin fashion by job)
```
bdice: Alternatively this could be called 'cyclic' -- to me, 'round-robin' _could_ suggest a finite time slot offered to each operation, as described here: https://en.wikipedia.org/wiki/Round-robin_scheduling

It's clear enough as-is, just wanted to offer this as a suggestion in case you like it better.

flow/project.py
```diff
@@ -1490,7 +1491,17 @@ def run(self, jobs=None, names=None, pretend=False, np=None, timeout=None, num=N
...
+                * 'random'    (shuffle the execution order randomly)
+                * callable specify a key-callable by which to order operations
```
bdice: Would prefer `* callable (specify a key-callable used to sort operations)`

flow/project.py
```diff
@@ -1565,6 +1576,23 @@ def select(operation):
...
+                          for _, group in groupby(operations, key=lambda op: op.job)]
+                operations = list(roundrobin(*groups))
+            elif order == 'random':
+                random.shuffle(operations)
+            elif order is None or order in ('none', 'by-job'):
+                pass  # by-job is the default order
+            else:
+                raise ValueError(
+                    "Invalid value for 'order' argument, valid arguments are "
+                    "None, 'none', 'by-job', 'round-robin', or 'random'.")
```
bdice: or callable?

flow/project.py
```diff
@@ -1565,6 +1576,23 @@ def select(operation):
...
+            # Optionally re-order operations for execution if order argument is provided:
+            if callable(order):
+                operations = list(sorted(operations, key=order))
+            elif order == 'round-robin':
+                groups = [list(group)
+                          for _, group in groupby(operations, key=lambda op: op.job)]
+                operations = list(roundrobin(*groups))
+            elif order == 'random':
+                random.shuffle(operations)
+            elif order is None or order in ('none', 'by-job'):
```
bdice: Why allow both `None` and `'none'`? Seems unnecessary and increases the API scope.

flow/project.py
```diff
@@ -2575,6 +2604,12 @@ class MyProject(FlowProject):
             const='-1',
             help="Specify the number of cores to parallelize to. Defaults to all available "
                  "processing units if argument is ommitted.")
```
bdice: fix typo: `omitted`

flow/util/misc.py
```diff
@@ -199,3 +200,19 @@ def keys_used(self):
...
     """Emulate python-3.4 re.fullmatch()."""
     return re.match("(?:" + regex + r")\Z", string, flags=flags)
+
+
+def roundrobin(*iterables):
+    # From: https://docs.python.org/3/library/itertools.html#itertools-recipes
+    "roundrobin('ABC', 'D', 'EF') --> A D E B F C"
+    # Recipe credited to George Sakkis
+    num_active = len(iterables)
+    nexts = cycle(iter(it).__next__ for it in iterables)
```
bdice: See? It's cyclic. (See previous comment) 😉

flow/util/misc.py
```diff
@@ -199,3 +200,19 @@ def keys_used(self):
 def fullmatch(regex, string, flags=0):
     """Emulate python-3.4 re.fullmatch()."""
     return re.match("(?:" + regex + r")\Z", string, flags=flags)
+
+
+def roundrobin(*iterables):
+    # From: https://docs.python.org/3/library/itertools.html#itertools-recipes
+    "roundrobin('ABC', 'D', 'EF') --> A D E B F C"
```
bdice: Would prefer a `# comment` instead of a string literal.

flow/project.py
csadorf: Like I explained in my response to @b-butler , I'd prefer to not explain what the code does "implicitly" in order to avoid users starting to rely on that. I think it would be highly preferable for us to have the option to change the default behavior at any point in time without needing to go through a deprecation cycle.

This is similar to the debate about Python's built-in `dict` class, where the insertion order was considered implementation detail up until Python 3.7, although it has been identical to `OrderedDict` for quite a while now. See this [StackOverflow reply](https://stackoverflow.com/a/39980744) for some details. I don't think it is a good idea for us to promise any kind of order while the implementation is still highly in flux.

flow/project.py
csadorf: The reason for this is so that we can easily specify 'None' from the command line, consider the following example:
```#!bash
for order in none round-robin random; do
    ./project.py run --order=$order -o foo bar
done
```
If `none` was not a valid argument, one would need to drop the `--order` argument altogether, which means you need to either make the code more verbose:
```#!bash
for order in --order=none --order=round-robin --order=random; do
    ./project.py run $order -o foo bar
done
```
or add an awkward special case with potential code duplication:
```#!bash
./project.py run foo bar
for order in round-robin random; do
    ./project.py run --order=$order -o foo bar
done
```
I've seen this kind of pattern before and would argue that that little bit of redundancy is warranted. If it helps, we can make `none` the documented default argument and avoid the redundancy as part of the API documentation.

flow/util/misc.py
csadorf: :D I'll change that.

flow/project.py
bdice: Good reasoning, I am fine with it this way now that I understand.

flow/project.py
bdice: I guess I didn’t realize this was the same as @b-butler’s suggestion. As a new user I would have difficulty understanding the default behavior as it is. I think it would be fine to make by-job the default and even change the execution order without a full deprecation, if you view the API guarantee as the end result of the state machine instead of the exact order of execution. However, I understand your reasoning and am fine with your decisions here.

flow/project.py
csadorf: Done.

flow/project.py
csadorf: I've switched it to cyclic.

csadorf:
@bdice I think I've addressed all of your comments and suggestions.
@b-butler I've tried to clarify my motivation for the choice of default argument both in a comment to you and have also tried to clarify this as part of the API documentation.

@vyasr @b-butler I'm still open to changing the default behavior, but I would like the default argument to remain 'none'.

flow/project.py
b-butler: That makes sense. I just didn't understand the reasoning before.

flow/project.py
```diff
@@ -1490,6 +1493,26 @@ def run(self, jobs=None, names=None, pretend=False, np=None, timeout=None, num=N
             Show a progress bar during execution.
         :type progess:
             bool
+        :param order:
+            Specify the order of operations, possible values are:
+                * 'none' or None (no specific order)
+                * 'by-job' (operations are grouped by job)
+                * 'cyclic' (order operations cyclic by job)
+                * 'random' (explicitly shuffle the execution order randomly)
+                * callable (specify a key-callable used to sort operations)
```
vyasr: This description isn't too clear to me, mostly because `key-callable` isn't really a canonical term AFAIK. I would prefer to word this as something like `a callable returning a comparison key for an operation that is used to sort operations`, that matches the terminology used in the documentation of `sorted` (comparison key).

flow/project.py
```diff
@@ -1490,6 +1493,26 @@ def run(self, jobs=None, names=None, pretend=False, np=None, timeout=None, num=N
             Show a progress bar during execution.
         :type progess:
             bool
+        :param order:
+            Specify the order of operations, possible values are:
+                * 'none' or None (no specific order)
+                * 'by-job' (operations are grouped by job)
+                * 'cyclic' (order operations cyclic by job)
+                * 'random' (explicitly shuffle the execution order randomly)
```
vyasr: no need for "explicitly"

flow/project.py
```diff
@@ -1490,6 +1493,26 @@ def run(self, jobs=None, names=None, pretend=False, np=None, timeout=None, num=N
...
+                * callable (specify a key-callable used to sort operations)
+
+            The default value is `none`, which is equivalent to `by-job` in the current
+            implementation.
+
+            .. note::
+
+                Users are advised to not rely on a specific execution order, as a
+                subsitute for defining the worklow in terms of pre- and post-conditions.
```
vyasr: typos: "subsitute"->"substitute", "worklow"->"workflow"

flow/util/misc.py
```diff
@@ -199,3 +200,19 @@ def keys_used(self):
...
     """Emulate python-3.4 re.fullmatch()."""
     return re.match("(?:" + regex + r")\Z", string, flags=flags)
+
+
+def roundrobin(*iterables):
+    # From: https://docs.python.org/3/library/itertools.html#itertools-recipes
+    # roundrobin('ABC', 'D', 'EF') --> A D E B F C
+    # Recipe credited to George Sakkis
+    num_active = len(iterables)
+    nexts = cycle(iter(it).__next__ for it in iterables)
```
vyasr: Are you inclined to leave this recipe unchanged since it was copied? If so, that's fine, but if you're open to changes I would rewrite this as `cycle(iter(next(it)))` rather than calling `__next__`.

vyasr:
> @bdice I think I've addressed all of your comments and suggestions.
> @b-butler I've tried to clarify my motivation for the choice of default argument both in a comment to you and have also tried to clarify this as part of the API documentation.
> 
> @vyasr @b-butler I'm still open to changing the default behavior, but I would like the default argument to remain 'none'.

I'm fine leaving the default behavior as is for now, we can revisit later if needed.

flow/util/misc.py
csadorf: Your solution would actually be wrong (I tested it). So I'm leaving as-is.

csadorf:
@vyasr I've addressed all of your change requests.

flow/util/misc.py
vyasr: Interesting. Do you know why that is?

flow/project.py
```diff
@@ -1498,16 +1498,17 @@ def run(self, jobs=None, names=None, pretend=False, np=None, timeout=None, num=N
                 * 'none' or None (no specific order)
                 * 'by-job' (operations are grouped by job)
                 * 'cyclic' (order operations cyclic by job)
-                * 'random' (explicitly shuffle the execution order randomly)
-                * callable (specify a key-callable used to sort operations)
+                * 'random' (shuffle the execution order randomly)
+                * callable (a callable returning a a comparison key for an
```
vyasr: typo: "a a comparison key"

flow/util/misc.py
csadorf: It is because we are creating a cycle of `__next__` functions, not a cycle of *next* elements.

flow/project.py
csadorf: damn :D

csadorf:
@bdice @vyasr @b-butler Thank you for helping with this feature!!

PR #124, opened 2019-07-11
Make environments opt-out

### LINKED ISSUE #16: Use environment profiles on opt-out basis
**[Original report](https://bitbucket.org/glotzer/signac-flow/issue/16) by Carl Simon Adorf (Bitbucket: [csadorf](https://bitbucket.org/csadorf), GitHub: [csadorf](https://github.com/csadorf)).**

To use the shipped environment profiles, users are currently required to *opt-in* by adding the following line to the `project.py` file: `import flow.environments`.

We do this as a pre-cautionary measure, however I think it is fair to assume that this should be the default behavior.

I propose that we switch to an *opt-out* model with 6.0, where users can opt out to use profiles by changing the signac configuration:
```
[flow]
use_shipped_environments = False
```
or by switching a global variable: `flow.USE_SHIPPED_ENVIRONMENTS = False`.

*The exact name of the variable would be up for debate.*

vyasr:
## Description
Environments are now imported by default. This can be avoided by setting a signac configuration variable.

## Motivation and Context
In general, users expect environments to be available and are surprised when they're not.
This resolves #16. 

## Types of Changes

- [x] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/__init__.py
+++ b/flow/__init__.py
@@ -25,15 +25,9 @@ from .template import init
 from .util.misc import redirect_log
 from .operations import with_job
 
-# Enable flow environments on an opt-out basis. Users must set the relevant
-# configuration variable to avoid environment being imported. This may be done
-# on either a global or a local configuration level. The order of precedence is
-# local up to global.
+# Import packaged environments unless disabled in config:
 from .util.config import get_config_value
-
-
-auto_import_environments = get_config_value('auto_import_environments', default=True)
-if auto_import_environments:
+if get_config_value('import_packaged_environments', default=True):
     from . import environments  # noqa:F401
 
 
--- a/flow/__init__.py
+++ b/flow/__init__.py
@@ -29,7 +29,6 @@ from .operations import with_job
 # configuration variable to avoid environment being imported. This may be done
 # on either a global or a local configuration level. The order of precedence is
 # local up to global.
-from signac.common import config as _config
 from signac import get_project as _get_project
 
 # Work up in the correct order
--- a/flow/__init__.py
+++ b/flow/__init__.py
@@ -31,13 +31,9 @@ from .operations import with_job
 # local up to global.
 from .util.config import get_config_value
 
-# Work up in the correct order
-try:
-    flag = get_config_value('IMPORT_ENVIRONMENTS')
-except LookupError:
-    flag = None
 
-if flag is None or flag == "True":
+auto_import_environments = get_config_value('auto_import_environments', default=True)
+if auto_import_environments:
     from . import environments  # noqa:F401
 
 
--- a/flow/__init__.py
+++ b/flow/__init__.py
@@ -32,7 +32,10 @@ from .operations import with_job
 from signac import get_project as _get_project
 
 # Work up in the correct order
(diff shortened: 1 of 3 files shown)


vyasr:
The corresponding docs PR is glotzerlab/signac-docs#33.

I've marked this as a breaking change since it could technically cause problems if someone was depending on not importing the environments. I think we should be OK with that change, though.

I'm not sure what the behavior should be in the case when you're importing flow but not in a project, for now I just have it importing environments.

flow/__init__.py
```diff
@@ -25,6 +25,22 @@
...
 
+# Enable flow environments on an opt-out basis. Users must set the relevant
+# configuration variable to avoid environment being imported. This may be done
+# on either a global or a local configuration level. The order of precedence is
+# local up to global.
+from signac import get_project as _get_project
+
+# Work up in the correct order
+try:
+    flag = _get_project().config['flow'].get('IMPORT_ENVIRONMENTS')
```
csadorf: There is no need to deal with any kind of logic concerning the origin of the configuration value here. Just use the `get_config_value()` function from the `util.config` module:
https://github.com/glotzerlab/signac-flow/blob/master/flow/util/config.py#L44

csadorf:
> I've marked this as a breaking change since it could technically cause problems if someone was depending on not importing the environments. I think we should be OK with that change, though.

This is an explicit and intentional breaking change.

flow/__init__.py
vyasr: Got it, thanks. I'll fix.

vyasr:
Yes, the fact that the config files at different levels get merged together does make it hard to provide a nice API, but I agree that it's the expected behavior. I suppose in that sense it's sensible to have the behavior just follow the global `.signacrc` when not in a project.

I was waiting to update the Changelog until we discussed, will do so now.

flow/__init__.py
csadorf: Sorry, I did not see that earlier. So far all configuration values have been snake_case. So I'd recommend to change the config key name to *auto_import_environments*.

Please rename `flag` to `auto_import_environments` and add `True` as the default value to the `get()` call.

flow/__init__.py
```diff
@@ -25,6 +25,22 @@
...
+except LookupError:
+    flag = None
+
+if flag is None or flag == "True":
```
csadorf: This should just be `if auto_import_environments:`, because I'm pretty sure that the config framework can handle Booleans.

flow/__init__.py
```diff
@@ -25,6 +25,22 @@
 from .util.misc import redirect_log
 from .operations import with_job
 
+# Enable flow environments on an opt-out basis. Users must set the relevant
+# configuration variable to avoid environment being imported. This may be done
+# on either a global or a local configuration level. The order of precedence is
+# local up to global.
+from signac import get_project as _get_project
+
+# Work up in the correct order
```
csadorf: This comment is obsolete.

vyasr:
@csadorf I think this is ready for merge, unless @rosecers plans to review or you would like to request an extra reviewer.

PR #132, opened 2019-07-18
Implement the "fork" directive.

csadorf:
## Description

If the "fork" directive is True, operations are always executed within a
subprocess even if they could be called directly.

## Motivation and Context

This is useful whenever certain parallelization schemes must be executed
within a nondaemonic process, e.g., `multiprocessing.Pool`.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -16,7 +16,7 @@ next
 - Fix erroneous line in the torque submission template (#126).
 - Add a testing module to easily initialize a test project (#130).
 - Add feature for integrated profiling of status updates to aid with the optimization of the FlowProject imlementation (#107, #110).
-- Add 'fork' directive to enforce the execution of operations within a subprocess.
+- Add ``@fork()`` decorator to enforce the execution of operations within a subprocess.
 
 Fixes
 +++++
--- a/changelog.txt
+++ b/changelog.txt
@@ -14,6 +14,7 @@ next
 - Allow the provision of multiple operation-functions to the ``pre.after`` and ``*.copy_from`` conditions (#120).
 - Add option to specify the operation execution order (#121).
 - Fix erroneous line in the torque submission template (#126).
+- Add 'fork' directive to enforce the execution of operations within a subprocess.
 
 Fixes
 +++++
--- a/flow/__init__.py
+++ b/flow/__init__.py
@@ -19,6 +19,7 @@ from .project import label
 from .project import classlabel
 from .project import staticlabel
 from .operations import cmd
+from .operations import fork
 from .operations import directives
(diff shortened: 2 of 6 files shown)


vyasr:
I think the meaningful test would be to define a function that requires forking (e.g. one using a Pool), testing that it fails if you put it in a normal operation, and then testing that it succeeds if you put it in a forked operation.

csadorf:
I'm still not 100% convinced that introducing a new top-level decorator is the way to go here. Maybe overkill?

csadorf:
@bdice @vyasr I've implemented a unit test and also enabled the specification of a condition-function (as a function of job). That makes it easier to only fork under specific conditions.

csadorf:
There seems to be some issue import issue, which causes the unit tests to fail non-deterministically. I'm investigating.

csadorf:
@bdice @vyasr Ready for merge.

vyasr:
LGTM. Regarding whether a top-level directive is necessary, what else would you suggest? Or do you simply question whether users need this power at all? It certainly seems useful/necessary in some cases.

csadorf:
> LGTM.

Anything you would like to resolve before approving this PR?

> Regarding whether a top-level directive is necessary, what else would you suggest? Or do you simply question whether users need this power at all? It certainly seems useful/necessary in some cases.

No, I just don't like introducing top-level items into the root namespace when we are not 100% sure yet how it fits in. Even as part of this PR we switched between making it a directive and making it a top-level function.

vyasr:
I appreciate that concern; what might you suggest as an alternative in this case? Might as well discuss properly now before we merge the PR and are forced to support it for backwards compatibility.

I think this information needs to be available at the level of executing an operation, and directives are the most natural method in flow for attaching metadata to an operation. Providing a top-level function seems less natural to use, and possibly even an anti-pattern from a user perspective, no?

csadorf:
@vyasr I think you were not present for this discussion. At one of the last sprints we decided that we are going to shift the focus of execution-directives more to "submission directives". Meaning, a lose set of keyword arguments that are interpreted by the templates as a convention and are not otherwise programmatically enforced. That's why we made the `@fork` decorator a top-level function. However, the *executable* directive breaks that pattern, because it *is* utilized by both the templates and the *run* functions.

vyasr:
After a brief conversation with @csadorf, we want to hold off merging until we have another discussion on where exactly this code belongs.

csadorf:
I'm going to reopen this after we have had that discussion.

PR #135, opened 2019-08-03
Show number of eligible jobs in status view

### LINKED ISSUE #134: Summary of eligible operations
### Feature description

It would be useful to see a summary of how many jobs are eligible for each operation when you run `status`. This would help, for example, when running on a cluster with a limit of how many jobs you can keep in the queue; being able to see how many jobs are eligible for a specific operation would aid in selecting an appropriate bundle size.

### Proposed solution

For the user, this could be a simple argument passed to status (perhaps `-s` for summary?) that then adds a table at the bottom of the status view. 

### Additional context

Presumably this could be achieved with labels, which would then show up in the table that is displayed with the default `status` view. I don't like that solution though because 1) that view only shows percentages of jobs (it may be good to have that also show number of jobs) and 2) you would have to write a label function that checks eligible operations for each job.

tcmoore3:
This commit adds functionality to print the number of jobs that are
eligible for each operation during a status view. I added this to the
"overview" section of the view, since it is an overview-like thing to
print. Fixes #134.

## Types of Changes
- New feature

## Checklist:

If necessary:


tcmoore3:
What happens if someone has a project with tens, or even hundreds, of operations defined for a given project? We wouldn't want to print every line then, especially if a lot of them don't have any jobs eligible. Maybe default to sorting the operations by the number of eligible jobs, and then adding an additional option to limit the number of lines to show, like what is currently done with the `-m` option?

Also this made me realize I'm not sure what happens with this branch when you do pass `-m` to a status view. It looks like `-m` only operates on the number of labels that are shown. So maybe just adding a similar argument for number of job operations is the way to go?

flow/project.py
```diff
@@ -1267,15 +1268,16 @@ def get(k, m):
             for status in statuses.values():
                 _add_parameters(status)
 
+        column_width_operation = 5
+        for key, value in self._operations.items():
```
vyasr: Can you just iterate over `.keys()`?

flow/templates/base_status.jinja
```diff
@@ -31,6 +34,16 @@
...
 {% endfor %}
 {% endblock%}
+{% block operation_summary %}
+
+{% set field_head = field_operation_title + '%s' %}
+{{ field_head | format('operation', 'number of eligible jobs') }}
+{{ field_head | format('-' * (column_width_operation+4), '-' * column_width_num_jobs) }}
+{% for op, n_jobs in op_counter.items() %}
+{% set x = op | length %}
+{{ op }}{{ ' ' * (column_width_operation+2-x+4) }}{{ n_jobs }}
```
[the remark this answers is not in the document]
vyasr: I haven't looked at jinja in a bit, so I don't remember how exactly you format the format-string, but is what you're looking for something like `"%%-%d%%s" | format(column_width_operation) | format(op)` (I may have botched the number of percent signs somewhere in there).

flow/templates/base_status.jinja
```diff
@@ -31,6 +34,16 @@
 {{ field_progress_bar | format(label[0], label[1] | draw_progressbar(jobs | length)) }}
 {% endfor %}
 {% endblock%}
+{% block operation_summary %}
+
+{% set field_head = field_operation_title + '%s' %}
+{{ field_head | format('operation', 'number of eligible jobs') }}
+{{ field_head | format('-' * (column_width_operation+4), '-' * column_width_num_jobs) }}
+{% for op, n_jobs in op_counter.items() %}
+{% set x = op | length %}
```
vyasr: Can we name this something meaningful like `op_length`?

flow/project.py
tcmoore3: Yes

tcmoore3: Yes

flow/templates/base_status.jinja
tcmoore3: 👍

flow/templates/base_status.jinja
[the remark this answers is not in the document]
vyasr: @tcmoore3 any luck trying this?

vyasr:
I think adding an option to limit is reasonable. Alternatively, could we just make the changes in this PR optional?

flow/templates/base_status.jinja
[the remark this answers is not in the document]
tcmoore3: This works, so I'll leave it as-is.

changelog.txt
```diff
@@ -17,6 +17,8 @@ next
 - Add a testing module to easily initialize a test project (#130).
 - Add feature for integrated profiling of status updates to aid with the optimization of the FlowProject imlementation (#107, #110).
 - Print number of eligible jobs for each operation in status view (#134)
+- Add show_traceback to config options (#61, #144).
+- The support for Python version 3.4 has been dropped.
```
csadorf: This line shouldn't be here.

flow/templates/base_status.jinja
```diff
@@ -1,14 +1,16 @@
 {# The following variables are available to all scripts. #}
+{% if overview or detailed %}
+{% set column_width_operation = [column_width_operation, 5] | max %}
```
vyasr: Is this necessary? There's already a default of 5 in the Python code now.

tcmoore3: You're right. Removed.

PR #136, opened 2019-08-08
Minor updates to the development instructions.

csadorf:
1. Make conda recommended method for environment management.
2. Make commands copy & pastable by removing the prompt.
3. Enumerate instructions.


csadorf:
We've uncovered issues with using conda as the primary development environment. Closing for now.

PR #138, opened 2019-08-09
Move the pre and post classes to project specific classes owned by a FlowProject

b-butler:
## Description

Currently the `_pre` and `_post` classes are not related to a given project. When dealing
with multiple projects that work on the same functions, this can result in unexpected behavior.
All conditions applied by any project will be applied to all projects.

The change implements a class factory that provides a `_pre` and `_post` class for each
`FlowProject` derived class by registering pre and post conditions into a class level dictionary
where functions serve as keys and list of condition functions as values.

## Motivation and Context

Documentation and API structure suggest that pre and post conditions are owned by
the `FlowProject` or derived class, but this is not true currently. This solves that issue.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:


csadorf:
@b-butler I've found the bug. It was primarily related to project class inheritance.

csadorf:
We should probably add a unit test that would fail on the previously incorrect behavior.

flow/project.py
```diff
@@ -133,52 +133,64 @@ def not_(cls, condition):
...
-        def metacondition(job):
-            return all(c(job)
-                       for other_func in other_funcs
-                       for c in getattr(other_func, '_flow_post', list()))
-        return cls(metacondition)
+        def __init__(self, condition):
+            self.condition = condition
 
+        def __call__(self, func):
+            self.owner_class._OPERATION_PRE_CONDITIONS[func].insert(0, self.condition)
```
b-butler: Out of curiosity why do we use insert rather than append?

flow/project.py
```diff
@@ -457,21 +469,28 @@ def __call__(self, job=None):
...
         cls = type.__new__(metacls, name, bases, dict(namespace))
 
         # All operation functions are registered with the operation() classmethod, which is
         # intended to be used as decorator function. The _OPERATION_FUNCTIONS dict maps the
-        # the operation name to the operation function.
+        # the operation name to the operation function. In addition, pre and
+        # post conditions are registered with the class.
+
         cls._OPERATION_FUNCTIONS = list()
+        cls._OPERATION_PRE_CONDITIONS = defaultdict(list)
```
b-butler: Ah nice use of `defaultdict`, I was not familiar with that feature.

flow/project.py
```diff
@@ -2312,11 +2329,37 @@ def hello(job):
...
+    @classmethod
+    def _collect_operations(cls):
+        "Collect all operations that were add via decorator."
+        operations = []
+        for cls_ in cls.__mro__:
+            operations.extend(getattr(cls_, '_OPERATION_FUNCTIONS', []))
+        return operations
+
+    @classmethod
+    def _collect_pre_conditions(cls):
```
b-butler: Good catch in multiple inheritance.

flow/project.py
csadorf: The nature of decorators as *functions of functions* means that the ones defined on top are executed last. However, to make it easier for users, we have specified that decorators that are _defined_ earlier have higher priority. That's why we insert them at the top of the list.

csadorf: See also here: http://docs.signac.io/en/latest/flow-project.html#defining-a-workflow

tests/test_project.py
```diff
@@ -306,6 +306,45 @@ def label2(job):
...
+        class A(FlowProject):
+            pass
+
+        class B(FlowProject):
+            pass
+
+        class C(A):
+            pass
+
+        @A.pre(lambda job: True)
```
csadorf: You could use `pre.always` here.

tests/test_project.py
```diff
@@ -306,6 +306,45 @@ def label2(job):
...
+        @A.operation
+        @B.operation
+        def op1(job):
+            pass
+
+        self.assertEqual(len(A._collect_pre_conditions()[op1]), 1)
+        self.assertEqual(len(B._collect_pre_conditions()[op1]), 2)
+        self.assertEqual(len(C._collect_pre_conditions()[op1]), 3)
+
+        @A.post(lambda job: True)
```
csadorf: You could use `post.always` here.

b-butler:
@csadorf I added an unit test to test the behavior, but since it is my commit, I can no longer approve the changes.

flow/project.py
```diff
@@ -2312,11 +2329,37 @@ def hello(job):
...
+        return ret
+
+    @classmethod
+    def _collect_post_conditions(cls):
+        "Collect all post-conditions that were add via decorator."
+        ret = defaultdict(list)
+        for cls_ in cls.__mro__:
+            for func, conds in getattr(cls_, '_OPERATION_POST_CONDITIONS', dict()).items():
+                ret[func].extend(conds)
+        return ret
```
vyasr: While I don't normally advocate for deduping code when there's just two copies (our favorite rule of 3!), in this case I think it would be worthwhile to combine the two collection functions for conditions. The code is identical except for the attribute, and I think it's worth consolidating any mro logic as much as possible to make it easier for devs/contributors who are less familiar with python internals.

flow/project.py
```diff
@@ -2312,11 +2329,37 @@ def hello(job):
         cls._OPERATION_FUNCTIONS.append((name, func))
         return func
 
+    @classmethod
+    def _collect_operations(cls):
+        "Collect all operations that were add via decorator."
+        operations = []
+        for cls_ in cls.__mro__:
```
vyasr: Can we name this to something like `parent_class` rather than just `cls_`? (Same for the below functions).

flow/project.py
```diff
@@ -133,52 +133,64 @@ def not_(cls, condition):
         return cls(lambda job: not condition(job))
 
 
-class _pre(_condition):
+def setup_pre_conditions_class(cls):
```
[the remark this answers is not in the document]
csadorf: I agree, moving them into the metaclass namespace would probably make sense.

flow/project.py
csadorf: I split them, because I thought it would be good to be able to access these two different attributes separately. It comes with code duplication, but like you said, only *once*, so not that big of a problem. I would advocate against deduping, but I don't feel strongly about it.

vyasr: I agree that being able to access them separately is good. An alternative would be to define a third (helper) function that contains all of the logic and takes a `key` or `attr` argument, then having either of these functions be one-liners that call the helper function with the appropriate attribute to get. To reiterate, my primary reasoning is that centralizing mro logic makes it easier for future developers who might not know how that works as well. If you prefer not to do this, then we can just leave it as is.

csadorf:
@b-butler Do you need help with addressing the change requests?

b-butler:
@csadorf The comments have been addressed.

tommy-waltmann:
Just make Simon and Vyas satisfied with the request and I'll approve it.

csadorf:
@vyasr Your changes should be addressed.

flow/project.py
```diff
@@ -452,23 +404,89 @@ def __call__(self, job=None):
 
 
 class _FlowProjectClass(type):
-    """Metaclass for the FlowProject class."""
```
vyasr: Don't remove the docstring.

tests/test_project.py
```diff
@@ -703,6 +749,69 @@ def test_run_parallel(self):
             else:
                 self.assertFalse(job.isfile('world.txt'))
 
+    def test_run_condition_inheritance(self):
+        _project = getattr(self, 'project_class', None)
```
vyasr: I don't think this is being used.

vyasr: Never mind, this is necessary for project mocking to work below (where `self.project_class` is reassigned to). Other tests will break if we don't save this value and reassign it at the end of the function. Please add a comment to that effect.

PR #139, opened 2019-08-13
Update tested signac versions.

csadorf:
Update the signac core version we test against.

Includes the previous 4 minor releases: (0.9.5, 1.0.0, 1.1.0, 1.2.0).


csadorf:
> Looks fine. (But do we really need to test 4 versions?)

They are pretty cheap, so any reason why not?

PR #145, opened 2019-08-21
Add deprecation notices to project.py

cbkerr:
Using the deprecation module, tagged things as `@deprecated` with @vyasr's help.

## Description

Marked 5 functions as deprecated, including one that used a non-standard plural of "status".

In the future, the definition of `export_job_stati()` could become that of `export_job_statuses()`.

## Motivation and Context

Part of the move to manage deprecations, see #65.
The change to "statuses" makes the code more discoverable.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

- [ ] My name is on the [list of contributors](https://github.com/glotzerlab/signac-flow/blob/master/contributors.yaml).<sup>2</sup>

<sup>2</sup>I added my name as part of #144.

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -2119,15 +2119,15 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
     @deprecated(deprecated_in="0.8", removed_in="1.0")
     def export_job_stati(self, collection, stati):
         "Export the job stati to a database collection."
-        for status in stati:
+        self.export_job_statuses(self, collection, stati)
+
+    def export_job_statuses(self, collection, statuses):
+        "Export the job statuses to a database collection."
+        for status in statuses:
             job = self.open_job(id=status['job_id'])
             status['statepoint'] = job.statepoint()
             collection.update_one({'_id': status['job_id']},
                                   {'$set': status}, upsert=True)
-
-    def export_job_statuses(self, collection, statuses):
-        "Export the job statuses to a database collection."
-        self.export_job_stati(self, collection, statuses)
 
     @classmethod
     def _add_print_status_args(cls, parser):
--- a/flow/project.py
+++ b/flow/project.py
@@ -2133,7 +2133,7 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
 
     def export_job_statuses(self, collection, statuses):
         "Export the job statuses to a database collection."
-        export_job_stati(self, collection, statuses)
+        self.export_job_stati(self, collection, statuses)
 
     @classmethod
     def _add_print_status_args(cls, parser):
--- a/flow/project.py
+++ b/flow/project.py
@@ -29,6 +29,7 @@ import inspect
 import functools
 import contextlib
 import random
+from deprecation import deprecated
 from collections import defaultdict
 from collections import OrderedDict
 from itertools import islice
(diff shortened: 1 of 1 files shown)


flow/project.py
```diff
@@ -2114,11 +2116,13 @@ def _add_direct_cmd_arg_group(cls, parser):
             help="Manually specify all labels that are required for the direct command "
                  "to be considered eligible for execution.")
 
+    @deprecated(deprecated_in="0.6", removed_in="0.8")
```
bdice: The versions specified in this line don't match the versions listed in the RuntimeError. It looks like this function has _already_ been removed. Am I missing something?

flow/project.py
```diff
@@ -2127,6 +2131,10 @@ def export_job_stati(self, collection, stati):
             collection.update_one({'_id': status['job_id']},
                                   {'$set': status}, upsert=True)
 
+    def export_job_statuses(self, collection, statuses):
+        "Export the job statuses to a database collection."
+        self.export_job_stati(self, collection, statuses)
```
bdice: I'm not 100% sure how `@deprecated` works, but I would guess that the implementation should be swapped. That is, `export_job_stati` should call `export_job_statuses` instead of the other way. Otherwise `export_job_statuses` will call a deprecated method and show a warning, right?

cbkerr: I agree with your reasoning and will swap the flag around.

flow/project.py
cbkerr: @vyasr and I wanted to elevate the fact that this was already "removed."

I can go ahead and fully delete it in this PR?

vyasr:
At a glance, other candidates for removal are `_tr` and perhaps `eligible_for_submission`? Overall there are fewer deprecations in this module than I might have expected.

One major deprecation that we should add is the `add_operation` function. That API has been deprecated for a very long time now, and I don't want to continue supporting it in 1.0. However, since that's pretty much guaranteed to break any old user code, are you reluctant to remove it @csadorf ?

csadorf:
> At a glance, other candidates for removal are `_tr` and perhaps `eligible_for_submission`? 

Yes, those and everything related should be deprecated.

> One major deprecation that we should add is the `add_operation` function. That API has been deprecated for a very long time now, and I don't want to continue supporting it in 1.0. However, since that's pretty much guaranteed to break any old user code, are you reluctant to remove it @csadorf ?

The `add_operation()` method is certainly not deprecated and I have no intention to change that. The decorator-based-API has always been intended as an alternative to using this function directly. 

Decorators provide a very useful and concise API layer for users, but they can be awkward when used programmatically otherwise. I strongly advise against deprecating this function and I don't see any major maintenance overhead anyways since it is basically a one-liner:
https://github.com/glotzerlab/signac-flow/blob/6f4b20a0c280b5b419018b50a9e32eb0b3dce46b/flow/project.py#L2303

flow/project.py
```diff
@@ -2468,6 +2452,11 @@ def eligible_for_submission(self, job_operation):
             return False
         return True
 
+    @deprecated(deprecated_in="0.8", removed_in="1.0",
+                details="Use _eligible_for_submission instead.")
```
bdice: This might not be appropriate to recommend `_eligible_for_submission`. Users shouldn't use private methods, we're just taking this out of the public API.

PR #148, opened 2019-08-23
deprecate progressbar.py

zhou-pj:
deprecate progressbar.py

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -958,16 +958,17 @@ class FlowProject(six.with_metaclass(_FlowProjectClass,
 
                 t = time.time()
                 num_jobs = len(jobs)
-                status_interval = 0.2
                 statuses = []
                 for i, job in enumerate(jobs):
                     time.sleep(0.01)
                     statuses.append(_get_job_status(job))
-                    if time.time() - t > status_interval:
-                        print('Collecting job status info: {}/{}'.format(i+1, num_jobs), end='\r')
+                    if time.time() - t > 0.2:  # status interval
+                        print(
+                            'Collecting job status info: {}/{}'.format(i+1, num_jobs),
+                            end='\r', file=err)
                         t = time.time()
                 # Always print the completed progressbar.
-                print('Collecting job status info: {}/{}'.format(i+1, num_jobs))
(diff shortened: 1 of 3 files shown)


zhou-pj:
A side note: I also changed the pythonpath in test_project to address bug in testing.

vyasr:
@csadorf I'm able to run this script on the login nodes:

```
import tqdm

for i in tqdm.tqdm(range(int(1e7))):
    pass
```

and it renders the progressbar without error. Is that sufficient to address your concern, or is there something further that needs to be tested?

csadorf:
> @csadorf I'm able to run this script on the login nodes:
> 
> ```
> import tqdm
> 
> for i in tqdm.tqdm(range(int(1e7))):
>     pass
> ```
> 
> and it renders the progressbar without error. Is that sufficient to address your concern, or is there something further that needs to be tested?

No, because the status update is in itself parallelized. Can you run the same code within a `multiprocessing.Pool` ?

vyasr:
```
from multiprocessing import Pool
import tqdm
from pbar import with_progressbar

def testfun(i):
    for i in range(int(i*1e7)):
        pass

with Pool(2) as p:
    list(with_progressbar(
        iterable=p.map(testfun, range(8)), desc='test'))
    list(tqdm.tqdm(
        iterable=p.map(testfun, range(8)), desc='test'))
```

I copied your progressbar into `pbar.py` to test this. This script runs, but neither progressbar properly updates live; they just show up when they full iterable has been consumed. That seems like a separate problem than the deprecation this PR is intended to address, though.

For good measure, I also just checked out this branch and tested it on a live project and the status update works fine.

csadorf:
@vyasr This is the code path that prompted me to implement the fallback-progress bar. I recommend that we keep that in place, but use `tqdm` *everywhere* else.

https://github.com/glotzerlab/signac-flow/blob/e18eb6e0c97afc73a3366775b6c68f1ef72fae9c/flow/project.py#L955

Issue #86, opened 2019-03-13
Drop support for Python 3.4


csadorf:
Python 3.4 reaches EOL after March 2019.

https://www.python.org/dev/peps/pep-0429/

0.7:

- [ ] ~~Issue a `DeprecationWarning` that Python 3.4 will no longer be supported in future versions.~~
- [x] Announce deprecation in the changelog.

0.8:

- [x] Stop testing 3.4 and start testing 3.8.

csadorf:
Resolved with merge of #149 .

PR #149, opened 2019-08-30
Drop support for Python 3.4 (#86)

csadorf:
Drop support for Python version 3.4.

Resolves issue #86 .

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,4 @@
 signac>=0.9.0
-enum34 ; python_version < "3.4"
 jinja2>=2.8
 cloudpickle
 deprecation>=2
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 signac>=0.9.0
+enum34 ; python_version < "3.4"
 jinja2>=2.8
 cloudpickle
 deprecation>=2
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -1,5 +1,4 @@
 signac>=0.9.0
-enum34 ; python_version < "3.4"
```
bdice: This line is actually for Python 2.7, not 3.4 if I understand correctly. https://github.com/conda-forge/signac-flow-feedstock/blob/a99c7d3728dbae4d8534c2797d37b64fb054c509/recipe/meta.yaml#L24

setup.py
```diff
@@ -12,8 +12,6 @@
     'jinja2>=2.8',
     # To enable the parallelized execution of operations across processes.
     'cloudpickle',
-    # To define IntEnum in flow/scheduling/base.py.
-    'enum34;python_version<"3.4"',
```
bdice: Same as above, may actually be for 2.7.

bdice:
@csadorf I reviewed too fast 😆

csadorf:
> @csadorf I reviewed too fast 😆

I was too slow. ;)

PR #157, opened 2019-09-02
Drop support for Python version 2.7.

csadorf:
Remove code paths and testing related to support for Python 2.7.

PR #158, opened 2019-09-02
Drop support for Python version 2.7 - additional changes.

bdice:
Additional changes related to #157.

### DIFF
--- a/flow/__init__.py
+++ b/flow/__init__.py
@@ -26,15 +26,10 @@ from .template import init
 from .util.misc import redirect_log
 from .operations import with_job
 
-from signac.common import six
 # Import packaged environments unless disabled in config:
 from .util.config import get_config_value
 if get_config_value('import_packaged_environments', default=True):
     from . import environments  # noqa:F401
-
-
-if six.PY2:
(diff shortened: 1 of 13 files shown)


flow/__init__.py
```diff
@@ -26,17 +26,12 @@
...
 
-from signac.common import six
 # Import packaged environments unless disabled in config:
 from .util.config import get_config_value
 if get_config_value('import_packaged_environments', default=True):
     from . import environments  # noqa:F401
 
 
-if six.PY2:
-    raise RuntimeError("Python 2 is not supported by this version of signac-flow!")
```
csadorf: I know that I initially advocated against this, but the project is basically non-functional on Python 2 with all of these changes. There should at least be a very highly visible warning.

bdice: Personally I don't think this is important. You can't install the package with Python 2 anymore, so why worry about that? You can get another opinion from @vyasr or keep it if you're so inclined. It's not clear to me when (if ever) that warning could go away. We also don't support Python 3.4...

csadorf: Fair enough.

Issue #115, opened 2019-05-23
CUDA initialized before forking


berceanu:
### Description

I am trying to integrate `fbpic`, a well-known CUDA code (based on Python + Numba) for laser-plasma simulation with `signac`. The integration repo is [signac-driven-fbpic](https://github.com/berceanu/signac-driven-fbpic). 

I managed to succesfully run on a single GPU, via `python3 src/project.py run` from inside the `signac` folder, but if I add `--parallel` I get

```console
numba.cuda.cudadrv.error.CudaDriverError: CUDA initialized before forking
```

The goal is to get 8 (independent) copies of `fbpic` (with different input params) running in parallel on the 8 NVIDIA P100 GPUs that are on the same machine.

### To reproduce

Clone the `signac-driven-fbpic` repo and follow the install instructions. Then go to the `signac` subfolder, and do

```console
conda activate signac-driven-fbpic
python3 src/init.py
python3 src/project.py run --parallel
```

### Error output

```console
(signac-driven-fbpic) andrei@ServerS:~/Development/signac-driven-fbpic/signac$ python3 src/project.py run --parallel --show-traceback
Using environment configuration: UnknownEnvironment
Serialize tasks|----------------------------------------------------------------------------------Serialize tasks|#####-----------------------------------------------------------------------------Serialize tasks|##########------------------------------------------------------------------------Serialize tasks|###############-------------------------------------------------------------------Serialize tasks|####################--------------------------------------------------------------Serialize tasks|##########################--------------------------------------------------------Serialize tasks|###############################---------------------------------------------------Serialize tasks|####################################----------------------------------------------Serialize tasks|#########################################-----------------------------------------Serialize tasks|###############################################-----------------------------------Serialize tasks|####################################################------------------------------Serialize tasks|#########################################################-------------------------Serialize tasks|##############################################################--------------------Serialize tasks|###################################################################---------------Serialize tasks|#########################################################################---------Serialize tasks|##############################################################################----Serialize tasks|##################################################################################Serialize tasks|##################################################################################Serialize tasks|##############################################################################################|100%
ERROR: Encountered error during program execution: 'CUDA initialized before forking'
Execute with '--show-traceback' or '--debug' to get more information.
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/home/andrei/anaconda3/envs/signac-driven-fbpic/lib/python3.6/multiprocessing/pool.py", line 119, in worker
[...]

PR #159, opened 2019-09-03
Feature/fork directive

csadorf:
#  Description

Implement the fork-directive that allows users to explicitly trigger forking of operations.

## Motivation and Context

Some operations that are parallelized should not be run within the interpreter process and should always be run in a sub-process. There is no explicit way of enforcing forking of Python-function-based operation at the moment.

There has been some discussion in the past about whether this should be better be a separate decorator since directives have previously been used primarily in the context of templates, however I believe that there are very strong arguments for making this a directive:

 * It is extremely easy to implement this behavior as a directive. The decorator is supposed to behave like a directive and for example is automatically an optional function of job, no extra handling needed.
 * We avoid needing to carry another parameter through all the various instances of flow-operations, job-operations, etc. That is especially important for the pending groups-feature, where would need to track forking as well.
* Directives are a powerful tool to control execution and submission behavior on a per-operation basis. Let's expand  on that.
* We avoid cluttering the package root namespace. We can always make this a top-level decorator at some point if we are so inclined.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/operations.py
+++ b/flow/operations.py
@@ -106,7 +106,7 @@ class directives(object):
     such as the number of processes required for execution of parallelized operations.
 
     In addition, you can use the `@directives(fork=True)` directive to enforce that a
-    particular  operation is always executed within a subprocess and not within the
+    particular operation is always executed within a subprocess and not within the
     Python interpreter's process even if there are no other reasons that would prevent that.
     """
 
--- a/flow/operations.py
+++ b/flow/operations.py
@@ -108,6 +108,10 @@ class directives(object):
     In addition, you can use the `@directives(fork=True)` directive to enforce that a
     particular operation is always executed within a subprocess and not within the
     Python interpreter's process even if there are no other reasons that would prevent that.
++    .. note::
++
++        Setting `fork=False` will not prevent forking if there are other reasons for forking,
++        such as a timeout.
     """
 
     def __init__(self, **kwargs):
--- a/flow/operations.py
+++ b/flow/operations.py
@@ -104,6 +104,10 @@ class directives(object):
 
     Directives can for example be used to provide information about required resources
     such as the number of processes required for execution of parallelized operations.
+
+    In addition, you can use the `@directives(fork=True)` directive to enforce that a
+    particular  operation is always executed within a subprocess and not within the
+    Python interpreter's process even if there are no other reasons that would prevent that.
     """
 
     def __init__(self, **kwargs):
--- a/flow/project.py
+++ b/flow/project.py
@@ -1947,7 +1947,9 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
             # engine and warn about those that have not been.
             keys_unused = {
                 key for op in operations for key in
-                op.directives._keys_set_by_user.difference(op.directives.keys_used)}
+                op.directives._keys_set_by_user.difference(op.directives.keys_used)
+                if key not in ('fork', )  # whitelist
+            }
             if keys_unused:
                 logger.warning(
                     "Some of the keys provided as part of the directives were not used by "
--- a/flow/project.py
+++ b/flow/project.py
@@ -1523,7 +1523,7 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
                 if pretend:
                     print(operation.cmd)
                 else:
-                    self._fork(operation, timeout)
+                    self._execute_operation(operation, timeout)
         else:
             logger.debug("Parallelized execution of {} operation(s).".format(len(operations)))
             with contextlib.closing(Pool(processes=cpu_count() if np < 0 else np)) as pool:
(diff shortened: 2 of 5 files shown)


flow/operations.py
```diff
@@ -104,6 +104,10 @@ class directives(object):
 
     Directives can for example be used to provide information about required resources
     such as the number of processes required for execution of parallelized operations.
+
+    In addition, you can use the `@directives(fork=True)` directive to enforce that a
+    particular  operation is always executed within a subprocess and not within the
```
bdice:
```suggestion
    particular operation is always executed within a subprocess and not within the
```

flow/operations.py
```diff
@@ -104,6 +104,10 @@ class directives(object):
 
     Directives can for example be used to provide information about required resources
     such as the number of processes required for execution of parallelized operations.
+
+    In addition, you can use the `@directives(fork=True)` directive to enforce that a
```
bdice:
```suggestion
    In addition, you can use the `@directives(fork=True)` decorator to enforce that a
```

tests/define_test_project.py
```diff
@@ -38,10 +39,15 @@ def op1(job):
...
+def _need_to_fork(job):
+    return job.doc.get('fork')
+
+
 @TestProject.operation
+@flow.directives(fork=_need_to_fork)
 @TestProject.post.true('test')
 def op2(job):
-    job.document.test = True
+    job.document.test = os.getpid()
```
bdice: Good idea for the test. 👍

flow/operations.py
csadorf: The thing I'm referencing is a decorator. Too confusing?

flow/operations.py
```diff
@@ -104,6 +104,10 @@ class directives(object):
...
+    particular operation is always executed within a subprocess and not within the
+    Python interpreter's process even if there are no other reasons that would prevent that.
     """
```
csadorf:
```suggestion
+    .. note::
+
+        Setting `fork=False` will not prevent forking if there are other reasons for forking,
+        such as a deviating executable or a timeout argument.
    """
```

flow/operations.py
bdice: I think the word directive is directly implied by `@directives` but "decorator" might be more helpful for new users. Accept or reject the suggestion at your discretion.

flow/operations.py
bdice: "deviating executable" is not a word I recognize and couldn't find on Google. Maybe there's a better word? "An executable that creates a fork on its own"?

flow/operations.py
csadorf: I'd prefer to leave as-is.

flow/operations.py
csadorf: I've updated my comment, better?

bdice:
@csadorf Can you comment on whether this solves issue #115, and how it relates to the closed PR #132? The code looks substantially different.

csadorf:
> @csadorf Can you comment on whether this solves issue #115, and how it relates to the closed PR #132? The code looks substantially different.

I decided to create a new PR instead of re-opening #132 .

The issue #115 would be addressed with this PR, albeit it would require users to use `@directives(fork=True)` instead of `@fork`.

flow/operations.py
csadorf: fixed typo

csadorf: Using timeout as example instead of executable.

flow/project.py
```diff
@@ -1581,21 +1581,36 @@ def _run_operations_in_parallel(self, pool, pickle, operations, progress, timeou
...
-            self._operation_functions[operation.name](operation.job)
-        else:   # need to fork
+        # Check if we need to fork for operation execution...
+        if (
+            # The 'fork' directive was provided and evaluates to True:
+            operation.directives.get('fork', False)
+            # Separate process needed to cancel with timeout:
+            or timeout is not None
+            # The operation function is not registered with the class:
+            or operation.name not in self._operation_functions
```
bdice: @csadorf Can you help me understand how this condition could happen? Is this from some kind of legacy `operations.py` syntax?

csadorf: You could register an operation with `add_operation`. Operations don't *have* to be tied to functions.

flow/project.py
```diff
@@ -1581,21 +1581,36 @@ def _run_operations_in_parallel(self, pool, pickle, operations, progress, timeou
...
+            # The specified executable is not the same as the interpreter instance:
+            or operation.directives.get('executable', sys.executable) != sys.executable
+        ):
+            # ... need to fork:
+            logger.debug(
+                "Forking to execute operation '{}' with "
+                "cmd '{}'.".format(operation, operation.cmd))
```
mikemhenry: Would it be worth printing the pid here?

csadorf: The PID is not known at this point. I'm not sure how difficult it would be to determine the PID.

PR #160, opened 2019-09-06
Add project_class "wrapper" around test test_copy_conditions

b-butler:
## Description

Prevents future tests from breaking because the project's project_class attribute is changed.

## Motivation and Context

Since the `ProjectClassTest`'s `project_class` attribute is changed in the test function
`test_copy_conditions`, it needs to be reset at the end of the function to ensure that future
test beneath this one do not fail unexpectedly.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/contributors.yaml
+++ b/contributors.yaml
@@ -92,4 +92,9 @@ contributors:
     given-names: Corwin
     affiliation: "University of Michigan"
     orcid: "https://orcid.org/0000-0003-0776-2596"
+  -
+    family-names: Butler
+    given-naes: Brandon
+    affiliation: "University of Michigan"
+    orcid: "https://orcid.org/0000-0001-7739-7796"
 ...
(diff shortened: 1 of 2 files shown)


b-butler:
We could wrap all tests that change the test class to catch any error and restore the class's project class.

vyasr:
Cleanest way to do that is probably create a separate test class for tests that do this and put the reassignment into the `setUp` and `tearDown` `unittest.TestCase` methods.

csadorf:
> Cleanest way to do that is probably create a separate test class for tests that do this and put the reassignment into the `setUp` and `tearDown` `unittest.TestCase` methods.

I agree. @b-butler Could you update the PR to realize that?

b-butler:
@vyasr and @csadorf The solution is fine, but that would cause those test to not be in the correct "conceptual" test class. If you are okay with that, then I am fine implementing it.

csadorf:
I think that's fine. Alternatively, could we make the project class a function argument to `mock_project` ?

b-butler:
@csadorf I think that is the proper solution. I have added a commit that does this and changed the relevant tests.

Issue #113, opened 2019-05-15
Proposal: Render status view from Markdown document


csadorf:
## Proposal

The status view rendering is currently refactored to use jinja2 as part of PR #111 . After completion of that change, we should consider to split the generation and presentation of the status view into distinct _document generation_ and _document presentation_ steps.

## Current rendering process

The status view presentation is currently implemented as follows:

1. Fetch the status data.
2. Prepare status data for presentation.
3. Render status data into a terminal-presentable view (via jinja2 after merge of PR #111).

## Proposed rendering process

With the proposed change, step 3 would be split up, such that we have the following process:

1. Fetch the status data.
2. Prepare status data for presentation.
3. Render status data into a Markdown document via jinja2.
4. Render the Markdown document for presentation (either on screen or HTML).

## Advantages and Disadvantages

Such a distinct separation of document generation and presentation has multiple **advantages**:
1. It significantly simplifies the writing of a template, especially with respect to table generation.
2. The status view is independent of its final rendered presentation, *i.e.*, the view can either be presented within a terminal or in a browser (e.g. for dashboard) or something completely different (all we need is a Markdown renderer for the desired format).

**Disadvantages** are:

1. The status view generation involves one more step, which might be harder to understand for users and developers.
2. The template does not represent the exact output presented on screen.

I believe that both of these points could be addressed by providing an option to just output the "raw" document on screen, which for the most part will be identical to the rendered document (except tables).

## Example

Generating a Markdown table within a jinja2 template would look like this:
```markdown
job_id | operation name
------ | --------------
{%- for operation in operations %}
{{ operation.job_id }} | {{ operation.name }}
{%- endfor %}
```
*Note, that we do not have to worry about column widths and matching the horizontal lines with the header widths is actually purely cosmetic, `-|-` would suffice as well for valid Markdown.*

## Implementation Roadmap

For the implementation we would need to

1. Provide/utilize a Markdown renderer that renders a Markdown document into a terminal-friendly presentation, and
2. Refactor the status templates to generate a Markdown document instead of a rendered view.

Step 1 could be accomplished with one of the following alternatives:

- A patched version of [markdown_viewer](https://github.com/axiros/terminal_markdown_viewer).
- A renderer implemented with [mistune](https://github.com/lepture/mistune).
- A renderer implemented with [mistletoe](https://github.com/miyuchina/mistletoe).

[...]

PR #163, opened 2019-09-14
issue #113, create markdown with jinja template in status view

zhou-pj:
## Motivation and Context
see issue #113 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


zhou-pj:
NOT ready for review yet, but would like to discuss the proper markdown rendering package and how to proceed at some point. I have modified all the template to create roughly the needed markdown syntax, will need further improvements upon figuring the rendering method.

vyasr:
@zhou-pj just for future reference, Github supports Draft PRs that allow you to make a pull request and ask for feedback while indicating that it's not yet ready for a thorough review and merge. Next time when you open a PR, click the arrow next to the green button for creating the PR and you'll see that one of the options is creating a draft PR.

flow/project.py
```diff
@@ -1410,7 +1391,21 @@ def _add_dummy_operation(job):
         if n > 0:
             context['op_counter'].append(('[{} more operations omitted]'.format(n), ''))
 
-        print(template.render(**context), file=file)
+        status_output = template.render(**context)
+        print(status_output, file=file)
```
b-butler: here we have 3 consecutive prints to the specified file. Is this for comparison? There is no logic to only have one print which is the final idea correct?

flow/project.py
```diff
@@ -1410,7 +1391,21 @@ def _add_dummy_operation(job):
...
+
+        import mistune
```
b-butler: The reason for having imports near where they are used is if the import would be conditional to a factor like a function being called. Having them here doesn't sure much purpose.

flow/project.py
```diff
@@ -1410,7 +1391,21 @@ def _add_dummy_operation(job):
...
+        html_output = mistune.markdown(template.render(**context))
+        print(html_output, file=file)
+        # import html2text
+        # status_output = html2text.html2text(status_output)
+
+        import mdv
```
b-butler: See comment above.

vyasr:
@zhou-pj what was the final result of the various renderers we discussed? Were you able to get HTML to render in the terminal, or was the markdown viewer package the best option? I agree with @b-butler that we definitely need something to prettify the default markdown output for the terminal. Can you summarize the conclusions you/we came to on the various package options for this rendering (mistune, misteletoe, mdv, markdown, html2text, etc)?

zhou-pj:
Sorry for the delay... @vyasr @b-butler  To summarize, I think the first step: using jinjia template to generate markdown status output looks pretty good to me now. For second step rendering on terminal, I get [terminal_markdown_viewer](https://github.com/axiros/terminal_markdown_viewer) to work if we generate terminal view directly from markdown (but I need to avoid using any '|' in output as this is used for markdown table formatting). I tried the html2text to see if html to terminal works but it looks like this package basically convert html back to markdown, not for terminal rendering. And as for providing html version for user, I think [mistune](https://github.com/lepture/mistune/blob/master/setup.py) works well to create html. I'll push a cleaned commit later today using mistune and terminal_markdown_viewer.

flow/project.py
zhou-pj: Yes these are currently my tests on different packages and different output syntax(markdown, html or terminal rendering). I will add logic to one of them based on user input later once we decide the best way for all of these output rendering.

flow/project.py
zhou-pj: I agree, these are for my initial trial, I'll move them once I finalize which package to use

vyasr:
I'm currently a bit swamped trying to finalize the 2.0 release of freud as well as get some data collection done for a project. I'll look back into this as soon as possible, but it'll probably be another week or so. Sorry for the delay @zhou-pj, I'll get to this as soon as possible.

zhou-pj:
Since I have issue getting a clean terminal output from md or html, Vyas and I agreed on moving forward with providing md and html with new jinja templates as an addition to the previous terminal output jinja template instead of doing everything from md. The current version is a complete functioning version that supports 3 output format: terminal printing, md and html. I will see if there is cleaner way to reduce code redundancy by combining the jinjia template.

flow/project.py
```diff
@@ -1403,19 +1296,23 @@ def _add_dummy_operation(job):
         op_counter = Counter()
         for job in context['jobs']:
             for k, v in job['operations'].items():
-                if v['eligible']:
+                if k != 'None' and v['eligible']:
```
vyasr: You'll also need to change this 'None' to ''

flow/render_status.py
```diff
@@ -0,0 +1,211 @@
...
+    def generate_terminal_output(self, template, context):
+        self.terminal_output = template.render(**context)
+
+    def generate_html_output(self, template, context):
+        self.generate_markdown_output(template, context)
+        self.html_output = mistune.markdown(self.markdown_output)
+
+    def render(self, template, template_environment, context, file, detailed, expand,
+               unroll, compact, pretty, option):
+        """render the status in different format for print_status.
```
vyasr: Capitalize "render".

vyasr: Can you also make the other parameters consistently capitalized below? e.g. "user provided Jinja2 template" -> "User provided...".

flow/render_status.py
```diff
@@ -0,0 +1,211 @@
...
+
+        :param template:
+            user provided Jinja2 template file.
+        :type template:
+            str
+        :param template_environment:
+            template environment.
+        :type template_environment:
+            FlowProject._template_environment
```
vyasr: :class:`FlowProject._template_environment`

flow/render_status.py
```diff
@@ -0,0 +1,211 @@
...
+
+        def get_operation_status(operation_info, symbols):
+            """Determine the status of an operation.
+
+            :param operation_info:
+                Dicionary containing operation information
+            :type operation_info:
+                Dictionary
+            :param symbols:
+                Dicionary containing code for different job status
```
vyasr: "Dictionary"

flow/render_status.py
```diff
@@ -0,0 +1,211 @@
...
+               job_op['eligible'] or all_ops:
+                return True
+            else:
+                return False
+
+        def get_operation_status(operation_info, symbols):
+            """Determine the status of an operation.
+
+            :param operation_info:
+                Dicionary containing operation information
```
vyasr: "Dictionary"

flow/render_status.py
```diff
@@ -0,0 +1,211 @@
...
+        if option == 'terminal':
+            self.generate_terminal_output(template, context)
+            print(self.terminal_output, file=file)
+            return self.terminal_output
+        elif option == 'html':
+            self.generate_html_output(template, context)
+            return self.html_output
+        elif option == 'md' or 'markdown':
+            self.generate_markdown_output(template, context)
+            return self.markdown_output
```
vyasr: Should add a final `else` that raises a `ValueError`.

flow/render_status.py
```diff
@@ -0,0 +1,211 @@
...
+        template_environment.filters['highlight'] = highlight
+        template_environment.filters['draw_progressbar'] = draw_progressbar
+        template_environment.filters['get_operation_status'] = get_operation_status
+        template_environment.filters['job_filter'] = job_filter
+
+        template = template_environment.get_template(template)
+
+        if option == 'terminal':
+            self.generate_terminal_output(template, context)
+            print(self.terminal_output, file=file)
```
vyasr: We should make this consistent; either `render` always writes its output to a provided file, or it always returns it. Right now it's inconsistent because terminal output is both printed and returned, whereas the other outputs are only returned.

setup.py
```diff
@@ -14,6 +14,8 @@
     'cloudpickle',
     # Deprecation management
     'deprecation>=2',
+    # For status view markdown rendering
+    'mistune'
```
vyasr: Is there a minimum version we need for this to work?

Also, I would suggest making this a soft requirement by using [`extras_require`](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) instead.

csadorf:
While I'm not formally reviewing this PR and just had a brief look at it, I'm a bit concerned that the patch in its current state represents a net code size increase. One major goal of the proposed change is to reduce code complexity. Did we potentially lose sight of that?

vyasr:
@csadorf I think that the problem can be substantially ameliorated by finding a way to combine the templates further, since I think that template logic duplication is the primary culprit for the increase. However, I agree that this PR no longer reduces code complexity, it simply adds new features (the ability to render non-terminal status output).

Based on our research into alternate solutions, the only way to truly reduce code complexity is to implement terminal markdown rendering into another package, and @zhou-pj has not found any straightforward ways to accomplish this extension within existing packages. As a result, I'm now treating this PR more as a new feature than a refactoring.

csadorf:
> However, I agree that this PR no longer reduces code complexity, it simply adds new features (the ability to render non-terminal status output).

There is not a strong enough need for that to justify the significant increase in complexity IMHO and time investment.

> Based on our research into alternate solutions, the only way to truly reduce code complexity is to implement terminal markdown rendering into another package, and @zhou-pj has not found any straightforward ways to accomplish this extension within existing packages. 

What are the exact issues? I am not able to deduce that from the comment chain on this PR (maybe I missed it?).

setup.py
zhou-pj: I modified it using extras_require following what I found online, but not entirely sure, would be great if you take another look on this.

zhou-pj:
> However, I agree that this PR no longer reduces code complexity, it simply adds new features (the ability to render non-terminal status output).

> There is not a strong enough need for that to justify the significant increase in complexity IMHO and time investment.

@csadorf @vyasr I rewrite the template to avoid duplication for markdown and terminal output. I agree with Vyas that I view this as simply adding the ability to render non-terminal status output for user. The code increase is significantly reduced now.

> What are the exact issues? I am not able to deduce that from the comment chain on this PR (maybe I missed it?).

The issue is mostly getting a clean terminal output directly from md or html, Vyas and I didn't find a good package to achieve that easily, and since we already have the ability to render good terminal output from jinja, we decided to take this easy route and just use what we have. The alternative option would be contribute to a package to enable that (or find one that already does it), Vyas mentioned that you @csadorf are taking a another look on this option, is there anything you find that could work?

csadorf:
@zhou-pj I've been starting to adapt mistune, but I will need a little bit of time to figure out the proper handling of tables. It's not trivial, but I believe it's possible and would be the ideal solution to this IMO. 

For this reason I'd like to keep this PR on hold for a bit longer.

zhou-pj:
@csadorf @vyasr I have modified this branch to work with https://github.com/csadorf/mistune/tree/terminal-renderer/mistune
and so far I didn't find any other issue with the terminal render. Should be ready for review and merge once the mistune patch is finalized.

flow/project.py
```diff
@@ -944,10 +944,10 @@ def print_status(self, jobs=None, overview=True, overview_max_lines=None,
                      unroll=True, compact=False, pretty=False,
                      file=None, err=None, ignore_errors=False,
                      no_parallelize=False, template=None, profile=False,
-                     eligible_jobs_max_lines=None):
+                     eligible_jobs_max_lines=None, option='terminal'):
```
csadorf: I think the parameter name should be something like `output_format` or so, `option` is very generic.

flow/project.py
```diff
@@ -2193,6 +2064,11 @@ def _add_print_status_args(cls, parser):
             '--no-parallelize',
             action='store_true',
             help="Do not parallelize the status determination.")
+        view_group.add_argument(
+            '-op', '--option',
```
csadorf: Similar to above, I think this option should be named `--format` or similar, `--option` is too generic. Also, single-dash options with more than one symbol (`-op`) should be avoided.

requirements-test.txt
```diff
@@ -1,3 +1,4 @@
 click==7.0
 ruamel.yaml==0.15.89
 flake8==3.7.7
+mistune==0.8.4
```
csadorf: Since `mistune` version `0.8.4` does not support terminal rendering (only our fork does), we will need to either make sure that my patch is merged into a release version of mistune, or we should bundle our custom version with this package.

setup.py
```diff
@@ -67,5 +67,9 @@
 
     install_requires=requirements,
 
+    extras_require={
+        'Markdown_HTML': ['mistune>=0.8.4'],  # For status view markdown rendering
```
csadorf: Similar to my comment above, but also, status rendering is such an essential feature that any related requirement should not be specified as an `extra` IMO.

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
+# make separate python class for render_status
+import mistune
+from .scheduling.base import JobStatus
+
+
+class renderer:
```
csadorf: Could you briefly elaborate on what the exact purpose of this class is? It seems to introduce a lot of boilerplate and is probably responsible for pushing this PR into a net positive increase in code lines, whereas the main purpose of this PR is to simplify and reduce code size.

zhou-pj: I think the main idea is to separate the rendering logic into this class so the the pring_status only deal with gathering the necessary job info for renderer class to render. This way we can have a clear place for rendering code and for future extension if necessary. Also this makes it possible to provide the whole status output class object for user to use and modify without digging into the main project.py code base. @vyasr I think we also discussed this a while a ago, any thoughts?

requirements-test.txt
zhou-pj: Agreed, what do you think is the best way to include the patch version? I noticed that coincidently the mistune author is also started working on mistune very recently and is planning on a 2.0 version. Did you have any contact with the author to merge or do you prefer we bundle the fork version in signac?

setup.py
zhou-pj: Agreed, I also noted that the mistune text renderer you wrote now requires bs4 and tabulate, is it the plan to add those requirements as well?

csadorf: Yes, but I was planning to reevaluate this after we decide to move forward with this.

flow/render_status.py
csadorf: I think this is generally a good idea, however the class structure appears highly intransparent to me at the moment. Would you mind checking if this could potentially be refactored? Maybe a set of functions would be more appropriate?

I could also give it a pass, just let me know, but I'd prefer if you could have critical look at this yourself first.

requirements-test.txt
csadorf: The patch is already based on the v2 code. If we think that this is sufficient, then I can try to create a pull request for mistune. If it gets accepted then we can just add this as a dependency, otherwise I'd be ok with vendoring the fork.

flow/render_status.py
vyasr: I think the main reason for the increase is that there are a lot more comment lines now because all of the options to `print_status` are duplicated in the docstring of `renderer.render`. The total lines of code are reduced. However, I'm happy to look at this with Pengji to find possible simplifications, we could definitely benefit from simplifying this code further.

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+            """filter eligible jobs for status print.
+
+            :param job_ops:
+                Operations information for a job.
+            :type job_ops:
+                OrderedDict
+            :param scheduler_status_code:
+                Dictionary information for status code
+            :type scheduler_status_code:
+                Dictionary
```
vyasr: The type for dictionaries should just be `dict`, not `Dictionary`.

requirements-test.txt
zhou-pj: @csadorf  I think the mistune branch you made serve our need well, and I agree if you could finalize and merge it to mistune, then us adding mistune 2.0 as dependency would be the easiest solution.

csadorf: I've created a pull request: https://github.com/lepture/mistune/pull/208

csadorf:
@bdice @vyasr I am very sorry, my browser glitched and I accidentally re-requested reviews from you instead of just from me.

vyasr:
No problem. However, this change has now splintered into many related issues on a few different repositories, and I think this PR is now the last in a long chain (except perhaps for lepture/mistune#208 depending on whether we choose to vendor a fork for a while), so I think we should focus on getting the upstream changes done first, starting with csadorf/mistune#1. Let me know where my help/feedback would be most helpful.

zhou-pj:
Update on this long lived PR:
I have add our forked mistune package into signac flow within this PR as suggested and changed the status output jinja template to only generate markdown, and then provide html and terminal output with mistune. 

I tested the this PR locally and so far find one minor formatting issue that I am not sure what's the best way resolve: under --pretty option when bold setting is used, the tabulate.py will align the table inaccurately because it is counting the '\033[1m' terminal bold markup text length need for table cell size as well but it won't actually take space in terminal, so the printed table will not be correctly aligned.

Other than this issue I find, I think the bulk of the PR is ready for re-review. @vyasr @csadorf

bdice:
@zhou-pj Here's a potentially helpful method for removing the ANSI escape sequences use `re.sub`. The regex is taken from: https://stackoverflow.com/questions/14693701/how-can-i-remove-the-ansi-escape-sequences-from-a-string-in-python

```python
>>> ansi_string = '\033[1mbold\033[0m text'
>>> print(len(ansi_string))
17
>>> ansi_escape = r'\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])'
>>> plain_string = re.sub(ansi_escape, '', ansi_string)
>>> print(len(plain_string))
9
```

flow/project.py
```diff
@@ -1146,9 +1147,11 @@ def print_status(self, jobs=None, overview=True, overview_max_lines=None,
                      unroll=True, compact=False, pretty=False,
                      file=None, err=None, ignore_errors=False,
                      no_parallelize=False, template=None, profile=False,
-                     eligible_jobs_max_lines=None):
+                     eligible_jobs_max_lines=None, output_format='terminal'):
         """Print the status of the project.
 
+        .. versionchanged:: 0.9
```
bdice: We have removed `versionchanged` from the codebase, so you can cut this out.

flow/project.py
```diff
@@ -1226,6 +1225,19 @@ def print_status(self, jobs=None, overview=True, overview_max_lines=None,
             User provided Jinja2 template file.
         :type template:
             str
+        :param profile:
+            show profile result.
```
```suggestion
            Show profile result.
```

flow/project.py
```diff
@@ -1226,6 +1225,19 @@ def print_status(self, jobs=None, overview=True, overview_max_lines=None,
...
+        :type profile:
+            bool
+        :param eligible_jobs_max_lines:
+            Limit the number of operations and its eligible job count printed in the overview.
+        :type eligible_jobs_max_lines:
+            int
+        :param output_format:
+            status output format, currently supports:
```
```suggestion
            Status output format, supports:
```

flow/project.py
```diff
@@ -1226,6 +1225,19 @@ def print_status(self, jobs=None, overview=True, overview_max_lines=None,
...
+            'terminal'(default), 'markdown'/'md' and 'html'
```
bdice: Do we need to support "markdown" and "md"? I would just keep "markdown" if that's code we control (don't change this if it's in mistune).
```suggestion
            'terminal' (default), 'markdown'/'md' or 'html'.
```

flow/project.py
```diff
@@ -2426,6 +2301,11 @@ def _add_print_status_args(cls, parser):
             '--no-parallelize',
             action='store_true',
             help="Do not parallelize the status determination.")
+        view_group.add_argument(
+            '-o', '--output-format',
+            type=str,
+            default='terminal',
+            help="Define status output format, currently support termial, html and markdown/md'")
```
```suggestion
            help="Set status output format: terminal, markdown, or html.")
```

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
+# make separate python class for render_status
+from .util import mistune
+from .scheduling.base import JobStatus
+
+
+class renderer:
```
bdice: Classes should start with uppercase.
```suggestion
class Renderer:
```

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+    """A class for rendering status in different format.
+
+    This class provides method and string output for rendering status output in different format,
+    currently supports: terminal, markdown and html
+
```
bdice: Remove blank line.
```suggestion
```

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+    def render(self, template, template_environment, context, file, detailed, expand,
+               unroll, compact, pretty, option):
+        """Render the status in different format for print_status.
+
+        :param template:
+            User provided Jinja2 template file.
+        :type template:
+            str
+        :param template_environment:
+            template environment
```
bdice:
```suggestion
            Template environment.
```

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+        :type template_environment:
+            :class:'FlowProject._template_environment'
```
bdice:
```suggestion
            :class:`FlowProject._template_environment`
```

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+        :param context:
+            Context that includes all the information for rendering status output.
+        :type context:
+            dict
+        :param file:
+            Redirect all output to this file, defaults to sys.stdout.
+        :type file:
+            str
```
bdice: Is this supposed to be a `str` or something else? A file-like object? Not sure what the right word is.

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+            Print a compact version of the output.
+        :type compact:
+            bool
+        :param pretty:
+            Prettify the output.
+        :type pretty:
+            bool
+        :param option:
+            Options for rendering format, currently supports:
+            'terminal'(default), 'markdown' and 'html'.
```
bdice:
```suggestion
            'terminal' (default), 'markdown' or 'html'.
```

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+            bool
+        :param compact:
+            Print a compact version of the output.
+        :type compact:
+            bool
+        :param pretty:
+            Prettify the output.
+        :type pretty:
+            bool
+        :param option:
```
bdice: Should this be called `output_format` for consistency? I prefer that over the name `option`.

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+            Options for rendering format, currently supports:
```
bdice:
```suggestion
            Rendering output format, supports:
```

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+                template = 'status_expand.jinja'
+            elif detailed and not unroll:
+                template = 'status_stack.jinja'
+            elif detailed and compact:
+                template = 'status_compact.jinja'
+            else:
+                template = 'status.jinja'
+
+        def draw_progressbar(value, total, escape='', width=40):
+            """Visualize progess with a progress bar.
```
bdice:
```suggestion
            """Visualize progress with a progress bar.
```

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+            if detailed and expand:
+                template = 'status_expand.jinja'
+            elif detailed and not unroll:
+                template = 'status_stack.jinja'
+            elif detailed and compact:
+                template = 'status_compact.jinja'
+            else:
+                template = 'status.jinja'
+
+        def draw_progressbar(value, total, escape='', width=40):
```
bdice: Should this be using `tqdm`? We will have it as an official dependency when #247 is merged.

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+            """
+
+            assert value >= 0 and total > 0
+            ratio = ' %0.2f%%' % (100 * value / total)
+            n = int(value / total * width)
+            return escape + '|' + ''.join(['#'] * n) + ''.join(['-'] * (width - n)) \
+                          + escape + '|' + ratio
+
+        def job_filter(job_op, scheduler_status_code, all_ops):
+            """filter eligible jobs for status print.
```
bdice:
```suggestion
            """Filter eligible jobs for status print.
```

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+            :param job_ops:
+                Operations information for a job.
+            :type job_ops:
+                OrderedDict
+            :param scheduler_status_code:
+                Dictionary information for status code
+            :type scheduler_status_code:
+                dict
+            :param all_ops:
+                Boolean value indicate if all operations should be displayed
```
bdice:
```suggestion
                Boolean value indicate if all operations should be displayed.
```

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+
+        def job_filter(job_op, scheduler_status_code, all_ops):
+            """filter eligible jobs for status print.
+
+            :param job_ops:
+                Operations information for a job.
+            :type job_ops:
+                OrderedDict
+            :param scheduler_status_code:
+                Dictionary information for status code
```
bdice:
```suggestion
                Dictionary information for status code.
```

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+            :type scheduler_status_code:
+                dict
+            :param all_ops:
+                Boolean value indicate if all operations should be displayed
+            :type all_ops:
+                bool
+            """
+
+            if scheduler_status_code[job_op['scheduler_status']] != 'U' or \
```
bdice: Maybe just use `return conditions` instead of `if conditions: return True; else: return False`

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+               job_op['eligible'] or all_ops:
+                return True
+            else:
+                return False
+
+        def get_operation_status(operation_info, symbols):
+            """Determine the status of an operation.
+
+            :param operation_info:
+                Dictionary containing operation information
```
bdice:
```suggestion
                Dictionary containing operation information.
```

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+            :type operation_info:
+                dict
+            :param symbols:
+                Dictionary containing code for different job status
```
bdice:
```suggestion
                Dictionary containing code for different job status.
```

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+            elif operation_info['completed']:
+                op_status = 'completed'
+            elif operation_info['eligible']:
+                op_status = 'eligible'
+            else:
+                op_status = 'ineligible'
+
+            return symbols[op_status]
+
+        if pretty:
```
bdice: This block can be simplified. We definitely need to remove the unused args from the docstring. Furthermore, you might be able to get `pretty` from the global scope and not conditionally define the function twice:
```python
def highlight(s, eligible):
    # insert docstring here
    if eligible and pretty:
        return '**' + s + '**'
    else:
        return s
```
or even
```python
highlight = lambda s, eligible: '**' + s + '**' if pretty and eligible else s
```
but using a lambda loses the documentation so that wouldn't be my preference.

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+            self.generate_terminal_output()
+            return self.terminal_output
+        elif option == 'html':
+            self.generate_html_output()
+            return self.html_output
+        elif option == 'md' or option == 'markdown':
+            return self.markdown_output
+        else:
+            raise ValueError('Option not supported, valid option format:'
+                             'termial, html and markdown/md')
```
bdice:
```suggestion
                             'terminal, markdown, or html.')
```

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+
+        template = template_environment.get_template(template)
+        self.markdown_output = template.render(**context)
+        if option == 'terminal':
+            self.generate_terminal_output()
+            return self.terminal_output
+        elif option == 'html':
+            self.generate_html_output()
+            return self.html_output
+        elif option == 'md' or option == 'markdown':
```
bdice:
```suggestion
        elif option == 'markdown':
```

flow/render_status.py
```diff
@@ -0,0 +1,220 @@
...
+            return self.markdown_output
+        else:
+            raise ValueError('Option not supported, valid option format:'
```
bdice:
```suggestion
            raise ValueError('Output format not supported, valid options are '
```

csadorf:
There are currently 4 reviewers on this PR, that seems a bit overkill. I'm happy to look at this, but maybe we can reduce that a bit?

bdice:
> 
> 
> There are currently 4 reviewers on this PR, that seems a bit overkill. I'm happy to look at this, but maybe we can reduce that a bit?

I only read over this PR because I wanted to know how the new system worked. I am going to step back and let other reviewers help finalize this PR, I just wanted a bit of familiarity with what's being done and suggested changes while I was reading through it.

csadorf:
> > There are currently 4 reviewers on this PR, that seems a bit overkill. I'm happy to look at this, but maybe we can reduce that a bit?
> 
> I only read over this PR because I wanted to know how the new system worked. I am going to step back and let other reviewers help finalize this PR, I just wanted a bit of familiarity with what's being done and suggested changes while I was reading through it.

That's totally fine and makes a lot of sense. I just wanted to make sure that we don't "burden" too many folks with the review requests.

@b-butler I'll remove the request for you for now, but please feel free to review this PR if you feel so inclined.

zhou-pj:
> @zhou-pj Here's a potentially helpful method for removing the ANSI escape sequences use `re.sub`. The regex is taken from: https://stackoverflow.com/questions/14693701/how-can-i-remove-the-ansi-escape-sequences-from-a-string-in-python
> 
> ```python
> >>> ansi_string = '\033[1mbold\033[0m text'
> >>> print(len(ansi_string))
> 17
> >>> ansi_escape = r'\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])'
> >>> plain_string = re.sub(ansi_escape, '', ansi_string)
> >>> print(len(plain_string))
> 9
> ```

Thanks for the suggestions! It turn out the tabulate.py has this implemented and the real issue was the way I add terminal support to mistune make used tabulate.py before the bold markup(\033[1m) is applied, so it cased that problem. I modified in mistune to fix it.

flow/render_status.py
zhou-pj: the use case here is just a static progress bar visualization and I am not sure if tqdm is need as I thought it is used to generate a dynamic progress bar, might be a overkill?

bdice: If we chose to use tqdm, you can do something like the following in 2 lines:
```python
from tqdm import tqdm
def draw_progressbar(value, total, escape='', width=40):
    bar_format = escape + '|{bar}' + escape + '| {percentage:0.2f}%'
    return tqdm.format_meter(n=value, total=total, elapsed=0, ncols=width, bar_format=bar_format)
```

Example:
```python
>>> draw_progressbar(value=17, total=58)
'|█████████                      | 29.31%'
```
The advantages of this solution: slightly easier to read, ensures consistency of look and feel across signac-flow, uses Unicode characters for partially filled blocks.

bdice: Just thought it was worth suggesting -- please feel free to accept or reject this suggestion (though I think it might be nice).

bdice: Oh, one more thing -- the ratio (percentage) is included in the width here. So the characters in the percentage count towards the total width.

zhou-pj: thanks for the code example! I switched it to use tqdm.

zhou-pj: one minor issue, i try to keep consistent with the current format style and used the {bar:40} setting to make all the bar aligned regardless of the percentage value length, which turn out to be a new feature only available since this release: https://github.com/tqdm/tqdm/releases/tag/v4.35.0 and is newer then the #247 reqested ('tqdm>=4.10.0'). Would it be ok to just pump the minimum version or is there other constraints?

bdice: @zhou-pj I think that would be fine to bump the version. Another alternative might be to keep our version at 4.10.0, use `ncols` to render _only_ the bar, and keep using format strings for the percentage and just concatenate the two strings.) That would be more similar to the old code, but just replacing the bar rendering pieces with tqdm.

csadorf: @zhou-pj Please feel free to bump the version number.

flow/util/mistune/plugins/tabulate.py
```diff
@@ -0,0 +1,1792 @@
+# -*- coding: utf-8 -*-
```
csadorf: So we're vendoring *tabulate* within the vendored *mistune* package?

flow/project.py
```diff
@@ -69,6 +69,7 @@
 from .labels import classlabel
 from .labels import _is_label_func
 from .util import config as flow_config
+from . import render_status
```
```suggestion
from .render_status import Renderer as StatusRenderer
```
Since that is the only class we are using from that module.

flow/project.py
```diff
@@ -1567,32 +1442,24 @@ def get(k, m):
...
         if overview:
             context['progress_sorted'] = progress_sorted
-            context['column_width_bar'] = column_width_bar
-            context['column_width_label'] = column_width_label
-            context['column_width_operation'] = column_width_operation
         if detailed:
-            context['column_width_id'] = column_width_id
-            context['column_width_operation'] = column_width_operation
-            context['column_width_total_label'] = column_width_total_label
             context['alias_bool'] = {True: 'T', False: 'U'}
```
csadorf: I know that this was introduced earlier, but I stumbled upon this while testing and don't think that the current aliases make much sense. What is 'T' and 'U' supposed to mean? I propose we replace this with 'Y' and 'N'.

flow/project.py
```diff
@@ -1603,19 +1470,25 @@ def _add_dummy_operation(job):
...
-                if v['eligible']:
+                if k != '' and v['eligible']:
                     op_counter[k] += 1
         context['op_counter'] = op_counter.most_common(eligible_jobs_max_lines)
         n = len(op_counter) - len(context['op_counter'])
         if n > 0:
             context['op_counter'].append(('[{} more operations omitted]'.format(n), ''))
 
-        print(template.render(**context), file=file)
+        render_result = render_status.Renderer()
```
```suggestion
        status_renderer = render_status.Renderer()
```

flow/project.py
```diff
@@ -1603,19 +1470,25 @@ def _add_dummy_operation(job):
...
+        render_output = render_result.render(template, template_environment, context, detailed,
+                                             expand, unroll, compact, output_format)
+
+        print(render_output, file=file)
 
         # Show profiling results (if enabled)
         if profiling_results:
             print('\n' + '\n'.join(profiling_results), file=file)
 
+        return render_result
```
csadorf: Why do we return that?

flow/util/mistune/plugins/tabulate.py
zhou-pj: Thanks for pointing this out, I actually haven't thought about this. I think the benefit of doing this might be that if would be easier for future manage or change if all are directly from our mistune fork and easy to remove if we ever get to merge our branch to mistune. but let me know if there is a strong reason for take tabulate out as a separate verdoring package.

flow/project.py
zhou-pj: I think this was from a discussion with @vyars awhile ago. The idea was that now that we have separate the format rendering logic into the renderer class, we can provide this directly to user to allow them more flexibility to render different format without repeatedly calling print_status.

flow/project.py
zhou-pj: I agree. This was only used when the --expand option is used and it is used for eligible label of yes or no. So I think Y/N make better sense.

flow/util/mistune/plugins/tabulate.py
csadorf: No, I think it's fine, I just found it curious. Shouldn't be a licensing issue afaik.

flow/project.py
csadorf: Ok, fair. I can see some use case where one would take advantage of that in the context of dashboard for example. I'm not 100% sure that this it the right level of abstraction, but I also don't really see a strong reason against this.

csadorf:
@vyasr LGTM, please merge at your discretion.

@zhou-pj Great work!!

setup.py
```diff
@@ -15,7 +15,7 @@
     # Deprecation management
     'deprecation>=2',
     # Progress bars
-    'tqdm>=4.10.0',
+    'tqdm>=4.35.0',
```
vyasr: What about the newer version is required?

flow/project.py
```diff
@@ -1603,19 +1470,25 @@ def _add_dummy_operation(job):
         op_counter = Counter()
         for job in context['jobs']:
             for k, v in job['operations'].items():
-                if v['eligible']:
+                if k != '' and v['eligible']:
```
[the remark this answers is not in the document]
zhou-pj: yes

flow/render_status.py
```diff
@@ -0,0 +1,180 @@
...
+                int
+            :param width:
+                The character width of the drawn progress bar.
+            :type width:
+                int
+            """
+
+            assert value >= 0 and total > 0
+            bar_format = escape + '|{{bar:{}}}'.format(width) + escape + '| {percentage:<0.2f}%'
+            return tqdm.format_meter(n=value, total=total, elapsed=0, bar_format=bar_format)
```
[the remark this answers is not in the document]
zhou-pj: yes, this is following Bradley's earlier comment to unify all progress bar using tqdm, but for this case we only need a static string progress bar so format_meter is used, and for the tqdm version bump, it is related to the `bar:{width}` feature I used to keep a uniform style. This is recently introduced only in 4.35.0. I pinned you on the earlier thread regarding this.

flow/render_status.py
zhou-pj: @vyasr

flow/render_status.py
```diff
@@ -0,0 +1,180 @@
...
+
+        def job_filter(job_op, scheduler_status_code, all_ops):
+            """Filter eligible jobs for status print.
+
+            :param job_op:
+                Operation information for a job.
+            :type job_op:
+                OrderedDict
```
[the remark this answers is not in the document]
zhou-pj: I actually never question this and just copied this from project.py. But I agree i don't think OrderedDict matters. I changed it to just dict

PR #178, opened 2019-11-02
Build dag

### LINKED ISSUE #41: Export workflow graph
**[Original report](https://bitbucket.org/glotzer/signac-flow/issue/41) by Carl Simon Adorf (Bitbucket: [csadorf](https://bitbucket.org/csadorf), GitHub: [csadorf](https://github.com/csadorf)).**

As per offline discussion, we determined that it should be possible to explicitly export the directed graph defined as `FlowProject`, *.e.g.*, as NetworkX graph structure.

In loose relation to that, we should think on how to submit a complete workflow instead of individual operations to a scheduler.

vyasr:
## Description
This PR implements operation graph detection. For testing this PR, the following FlowProject and testing script may be helpful (I had to make them text files for GitHub to allow the upload):
[project.py.txt](https://github.com/glotzerlab/signac-flow/files/3800728/project.py.txt)
[test_dag.py.txt](https://github.com/glotzerlab/signac-flow/files/3800730/test_dag.py.txt)

You can create an arbitrary random project and then run this script. I will create new tests based on these ideas (and update changelog etc) once we're happy with the implementation.

@csadorf I think this change should sufficiently address your concerns about the old PR. It uses simpler logic than the original version and is much more transparent. There is a little bit of additional complexity with the use of tags, but it should be easy to follow. The tags are necessary because the `__code__.co_code` of the automatically generated lambda functions that we use internally is unchanged by the argument, i.e. `FlowProject.pre.true('a')`, and `FlowProject.pre.true('b')` will evaluate to being identical. Unfortunately this would also affect similarly automatically generated user conditions, but we can document that users can provide the tag argument to condition construction if they wish to use the graph with such conditions.

## Motivation and Context

Supersedes #71, resolves #41.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -86,16 +86,18 @@ The available template variables are:
 {template_vars}
 
 Filter functions can be used to format template variables in a specific way.
-For example: {{ project.get_id() | capitalize }}.
+For example: {{ project.get_id() | captialize }}.
 
 The available filters are:
 {filters}"""
 
 
 class _condition(object):
-    # Random tag id used to ensure that "always" and "never" conditions never
-    # match each other.
-    rand_tag_id = 0
+    # This counter should be incremented each time an "always" or "never"
+    # condition is created, and the value should be used as the tag for that
+    # condition to ensure that no pair of "always" and "never" conditions
+    # are found to be equal by the graph detection algorithm.
+    tag_id_counter = 0
 
     def __init__(self, condition, tag=None):
         # Add tag to differentiate built-in conditions during graph detection.
@@ -124,14 +126,14 @@ class _condition(object):
     @classmethod
     def always(cls, func):
         "Returns True."
-        cls.rand_tag_id += 1
-        return cls(lambda _: True, str(cls.rand_tag_id))(func)
+        cls.tag_id_counter += 1
+        return cls(lambda _: True, str(cls.tag_id_counter))(func)
 
     @classmethod
     def never(cls, func):
         "Returns False."
-        cls.rand_tag_id += 1
-        return cls(lambda _: False, str(cls.rand_tag_id))(func)
+        cls.tag_id_counter += 1
+        return cls(lambda _: False, str(cls.tag_id_counter))(func)
 
     @classmethod
     def not_(cls, condition):
--- a/flow/project.py
+++ b/flow/project.py
@@ -96,25 +96,25 @@ class _condition(object):
 
     def __init__(self, condition, tags=None):
         # Add tag to differentiate built-in conditions during graph detection.
-        condition._flow_tags = tags
+        condition._flow_tag = tags
         self.condition = condition
 
     @classmethod
     def isfile(cls, filename):
         "True if the specified file exists for this job."
-        return cls(lambda job: job.isfile(filename), (filename, ))
+        return cls(lambda job: job.isfile(filename), filename)
 
     @classmethod
     def true(cls, key):
         """True if the specified key is present in the job document and
         evaluates to True."""
-        return cls(lambda job: job.document.get(key, False), (key, ))
+        return cls(lambda job: job.document.get(key, False), key)
 
     @classmethod
     def false(cls, key):
         """True if the specified key is present in the job document and
         evaluates to False."""
-        return cls(lambda job: not job.document.get(key, False), (key, ))
+        return cls(lambda job: not job.document.get(key, False), key)
 
     @classmethod
     def always(cls, func):
@@ -758,7 +758,7 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
                     callbacks = callbacks.union(
                         unpack_conditions(cf._composed_of))
                 else:
-                    callbacks.add((cf._flow_tags, cf.__code__.co_code))
+                    callbacks.add((cf._flow_tag, cf.__code__.co_code))
 
             return callbacks
 
--- a/flow/project.py
+++ b/flow/project.py
@@ -94,31 +94,27 @@ The available filters are:
 
 class _condition(object):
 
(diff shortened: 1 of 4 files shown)


flow/project.py
```diff
@@ -86,48 +86,58 @@
 {template_vars}
 
 Filter functions can be used to format template variables in a specific way.
-For example: {{ project.get_id() | captialize }}.
+For example: {{ project.get_id() | capitalize }}.
```
csadorf: This seems unrelated?

flow/project.py
```diff
@@ -687,6 +702,89 @@ def label_func(func):
...
+
+        Given a FlowProject subclass defined in a module `project.py`, the
+        output graph could be visualized using Matplotlib and NetworkX with the
+        following code:
+
+        .. code-block:: python
+
+            import numpy as np
+            from matplotlib import pyplot as plt
+            import networkx as nx
```
csadorf: I would swap the above two lines.

flow/project.py
```diff
@@ -687,6 +702,89 @@ def label_func(func):
...
+
+            from project import Project
+
+            p = Project()
```
csadorf: I'd prefer to not abbreviate `project` to `p` if this is possible without sacrificing readability.

flow/project.py
```diff
@@ -687,6 +702,89 @@ def label_func(func):
...
+            ops = p.operations.keys()
+            adj = np.asarray(p.detect_operation_graph())
+
+            plt.figure()
+            g = nx.DiGraph(adj)
+            pos = nx.spring_layout(g)
+            nx.draw(g, pos)
+            nx.draw_networkx_labels(g, pos,
+                labels={key: name for (key, name) in zip(range(len(ops)), [o for o in ops])})
```
csadorf: Maybe we can break this line?

flow/project.py
```diff
@@ -687,6 +702,89 @@ def label_func(func):
...
+            plt.show()
+        """
+
+        def to_callbacks(conditions):
+            """Get the actual callables associated with FlowConditions."""
+            return [condition._callback for condition in conditions]
+
+        def unpack_conditions(condition_functions):
+            """Identify any metaconditions in the list and reduce them to the
+            functions they're composed of. The callbacks argument is used
```
```suggestion
            functions that they are composed of. The callbacks argument is used
```

flow/project.py
```diff
@@ -405,6 +415,18 @@ def __call__(self, job=None):
             return self._cmd.format(job=job)
 
 
+def _create_metacondition(condition_dict, *other_funcs):
```
csadorf: _I had left the most important comment as part of my review on this function, but somehow it got lost during the submission? Anyways, I'll just write it again..._

Here we make an assumption here that all meta-conditions use the `all` function to aggregate multiple condition functions. I think this should be either made explicit in the function name or the aggregation function (`all`, `any`, etc.) should be an argument.

flow/project.py
vyasr: It is, but it's such a minor typo fix that I was willing to put it into this PR.

flow/project.py
vyasr: This is a good point. For now I'm going to just make it explicit in the name because I'm not exactly certain how the `composed_of` logic for graphs would work for other aggregation functions and I don't see a purpose to generalizing until we have an actual use-case.

mikemhenry:
> I had to make them text files for GitHub to allow the upload

What year is it github?! That annoys me so much. I can take a look at this tonight or tomorrow.

vyasr:
@mikemhenry sounds good. I recall that you reviewed an earlier PR on this topic so I figured you'd be a good second reviewer (I think @bdice has had some other PR reviews requested recently so I moved it to you).

flow/project.py
csadorf: I think it's fine to move forward with this without much further consideration since this is purely internal anyways.

flow/project.py
csadorf: According to our [governance](https://signac.io/governance/) 

> Committers may commit small changes (documentation updates, code comments, CI configuration changes) without a pull request

I think this falls under this.

The reason that I am so anal about this particular rule, is because we generally squash all commits prior to merging PRs and that means that these unrelated changes are now convoluted with the actually relevant changes. This is highly confusing to anybody who is reading the git log and attempts to understand the motivation for a particular change.

mikemhenry: I'm with @csadorf on this one, I can't find the article now but if we are strict about things like this, git blame and other git features work best when we keep the git log/commit history concise and "right"

flow/project.py
```diff
@@ -86,48 +86,58 @@
...
+For example: {{ project.get_id() | capitalize }}.
 
 The available filters are:
 {filters}"""
 
 
 class _condition(object):
-
-    def __init__(self, condition):
+    # Random tag id used to ensure that "always" and "never" conditions never
```
mikemhenry: This is more of a clarifying question, but I don't see how the the tag id is 'random'. I think it either needs a better name (arbitrary?) or the comment here needs to explain.

flow/project.py
vyasr: Fair enough, I can revert this change then.

flow/project.py
vyasr: You're correct, arbitrary is a more appropriate choice of words. I'll change that name.

vyasr: For clarification, it only exists to ensure that these conditions (always and never) are not found to be equal.

vyasr:
@mikemhenry I think I've addressed your question now. Feel free to merge if you think it's ready.

PR #179, opened 2019-11-10
Add deprecation to always

### LINKED ISSUE #177: Refining conditions
There are a couple of things I find confusing about the existing conditions that should perhaps be discussed before a 1.0 release. First, should the `never` and `always` conditions be part of `pre` rather than `_condition`? Placing them as post-conditions just flips their meaning, so it's well-defined, but I'm not sure it conceptually makes sense to put them as post-conditions I think it's a worthwhile simplification. Note that the current behavior also allows nonsensical specifications such as tagging an operation with `pre.never` and `post.never`.

Second, even though neither of these take any arguments, I would still prefer to change them to require parenthesis for consistency. I never remember that the expected syntax for these is different than for other conditions. There are pros and cons to making this change, but in my opinion consistency is more important here, especially because I assume that these are among the least-used conditions so users will expect them to match the behavior of other conditions that they have already used. Is that consistent with what others have experienced?

vyasr:
## Description

Add deprecation notice to "always" condition.

## Motivation and Context

Resolves #177 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -115,6 +115,7 @@ class _condition(object):
         return cls(lambda job: not job.document.get(key, False))
 
     @classmethod
+    @deprecated(deprecated_in="0.9", removed_in="1.0")
     def always(cls, func):
         "Returns True."
         return cls(lambda _: True)(func)
--- a/flow/project.py
+++ b/flow/project.py
@@ -115,7 +115,9 @@ class _condition(object):
         return cls(lambda job: not job.document.get(key, False))
 
     @classmethod
-    @deprecated(deprecated_in="0.9", removed_in="1.0")
+    @deprecated(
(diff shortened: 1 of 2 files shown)


flow/project.py
```diff
@@ -115,6 +115,7 @@ def false(cls, key):
         return cls(lambda job: not job.document.get(key, False))
 
     @classmethod
+    @deprecated(deprecated_in="0.9", removed_in="1.0")
```
```suggestion
    @deprecated(
        deprecated_in="0.9", removed_in="1.0",
        details="This condition decorator is obsolete.")
```

vyasr:
Thanks for the fix!

PR #180, opened 2019-11-11
Add 'current_version' argument to deprecated calls.

csadorf:
To make sure that future deprecations are triggered for the right
versions and that unit tests can automatically fail when a particular
deprecated element has not actually been removed.

While reviewing #179 I noticed that we don't actually pass the `current_version` argument to the `deprecated` calls which makes it harder for the library to issue correct warnings and impossible for automatic failure of unit tests if we failed to remove deprecated elements.

### DIFF
--- a/flow/version.py
+++ b/flow/version.py
@@ -4,4 +4,4 @@
 
 __all__ = ['__version__']
 
-__version__ = '0.9.0'
+__version__ = '0.8.0'
--- a/flow/version.py
+++ b/flow/version.py
@@ -1,0 +1,7 @@
+# Copyright (c) 2019 The Regents of the University of Michigan
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+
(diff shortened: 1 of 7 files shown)


flow/version.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) 2019 The Regents of the University of Michigan
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+
+__all__ = ['__version__']
+
+__version__ = '0.9.0'
```
bdice: @csadorf The version you removed from `__init__.py` is v0.8.0. This says v0.9.0. Should this be reverted to v0.8.0?
```suggestion
__version__ = '0.8.0'
```

csadorf: I just wanted to test whether you guys pay attention..., jk, I accidentally committed that after testing, sorry about that.

Issue #174, opened 2019-10-30
Add mpi_cmd back to environments


vyasr:
### Feature description

Environments should "know" the MPI command that should be used (`mpiexec`, `ibrun`, `jsrun`, etc) and how to handle arguments.

### Proposed solution

Versions of this function exist in pre-Jinja versions of signac-flow. We would simply need to extract that code again, and potentially generalize it to respect directives. Additionally, we may want to expose these as template filters.

### Additional context

Making this change is necessary to allow `FlowProject.run` to respect execution directive in an environment-aware manner.

vyasr:
Resolved by #184

Issue #175, opened 2019-10-30
FlowProject.run should respect directives


vyasr:
### Feature description

Currently `FlowProject.run` ignores execution directives, i.e. operations marked with `@flow.directives(nranks=8)` will run serially unless they are actually submitted to a scheduler. Instead, `FlowProject.run` should properly handle these directives.

### Proposed solution

This change depends on #174, which will provide the require MPI commands (and potentially other execution directive-specific commands for e.g. OpenMP). Once that issue is resolved, we can change run to call the necessary function to modify the run command. The other important change will be that `FlowProject.run` now needs to decide whether or not to fork based on an additional set of criteria that accounts for these directives.

### Additional context

Making this change is critical to enabling groups #114.

vyasr:
There is one specific case that will require some extra work to enable, namely the offset-based bundling of multiple MPI jobs onto one node on stampede. Currently we enable this by looping and using `python project.py exec` in the template script. @b-butler and I discussed that to enable this properly we will probably need to remove all such looping from the template and instead enable it directly in run. However, this means that we may also need to generalize the way in which we generate the commands executed by `python project.py run` to naturally enable environments to perform specific modifications to the run command such as adding the offset for stampede.

vyasr:
Partially resolved by #208. The question I raised in my previous comment remains to be addressed, but that probably will be done as part of #114.

vyasr:
The part of this that remains unresolved is currently very specific to Stampede2 and is separately documented in #250, so this issue can be closed.

PR #114, opened 2019-05-18
Implementing a grouping feature to organize flow operations

b-butler:
## Description

Goal of this branch is to create a grouping feature that will allow for operations in `FlowProject`
to be organized into "metaoperations." Currently to impliment this behavior either a new
`FlowProject` has to operate on the original set of operations or user commenting has to be
done to prevent the group operations from running individually.
 
An example snippet of expected functional shows in more detail a potential API and use
of the feature,
```
a_group = Project.make_group('a')

@Pr.operation
@a_group
def foo(job):
    pass

@Pr.operation
@a_group
def bar(job):
    pass
```
Then to use the grouping a command like,
```submit --group a_group```
would output a job for the scheduler with the final command being something
akin to
```python project.py run -j abc123 -o foo bar```

## Motivation and Context

This change is a feature change that seeks to give the user more control over the workflow
of their project. The group concept will allow for operations to be tagged to be run at once
in a single job which can work even if the pre and post conditions are not met initially but will
be once other operations have run.

This abstraction of a workflow is useful for multiple scenerios such as automatically restricting
operations to particular environements, running a job repeatedly in a single submit #4 , and moving
workflow logic into the `FlowProject` and not the individual operations for idiomatic code. Another use
case would to be mark operations as submit only #33 . Another issue tangentially related would be #41 as groups could be considered to create graphs themselves or groups might make up higher
level graphs.

## Types of Changes

- [x] Documentation update
- [ ] Bug fix
- [x] New feature
- [x] Breaking change<sup>1</sup>

<sup>1</sup>The change breaks (or has the potential to break) existing functionality.

## Checklist:

- [x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac-flow/blob/master/ContributorAgreement.md).
- [x] My name is on the [list of contributors](https://github.com/glotzerlab/signac-flow/blob/master/contributors.yaml).
- [x] My code follows the [code style guideline](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md#code-style) of this project.
- [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.

If necessary:
- [ ] I have updated the API documentation as part of the package doc-strings.
[...]

Issue #38, opened 2018-10-02
Enable condition-override from the command line


csadorf:
**[Original report](https://bitbucket.org/glotzer/signac-flow/issue/38) by Jens Glaser (Bitbucket: [jens_glaser](https://bitbucket.org/jens_glaser), GitHub: [jglaser](https://github.com/jglaser)).**

----------------------------------------

Or alternatively, `--run` should allow to force operation execution even if the `post` condition is already fulfilled.

csadorf:
The `exec` command is meant to not have any logic, I therefore suggest that we enable the `run` command to override conditions. We could add three switches `--ignore-conditions`, --ignore-pre-conditions`, `--ignore-post-conditions`. In fact, these options could be universal for `script`, `run`, and `submit`.

csadorf:
@alacour Are you still working on this?

alacour:
No, I am not working on it

zhou-pj:
issue addressed in PR #209

PR #184, opened 2019-11-22
Issue#174 first step: move environment related mpi_prefix logic to environment class and make template filter

zhou-pj:
Issue #174 first step

## Description
move environment related mpi_prefix logic to environment class and make template filter

## Motivation and Context
This changes allow future change to allow run to access environment specific mpiexec

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


flow/environment.py
```diff
@@ -191,6 +191,18 @@ def get_config_value(cls, key, default=flow_config._GET_CONFIG_VALUE_NONE):
         """
         return flow_config.require_config_value(key, ns=cls.__name__, default=default)
 
+    @staticmethod
+    def generate_mpi_prefix(operation):
```
csadorf: I'd propose a slightly more concise name here: `get_mpi_prefix`.

flow/templates/base_script.sh
```diff
@@ -17,7 +17,7 @@ cd {{ project.config.project_dir }}
 {% set cmd_suffix = cmd_suffix|default('') ~ (' &' if parallel else '') %}
 {% for operation in operations %}
 {% if operation.directives.nranks and not mpi_prefix %}
-{% set mpi_prefix = "%s -n %d "|format(mpiexec|default("mpiexec"), operation.directives.nranks) %}
+{% set mpi_prefix = operation|generate_mpi_prefix %}
```
csadorf: I think we should not specify a separate variable here, but instead line 27 should read:

`{{ mpi_prefix|default(operation|generate_mpi_prefix }}`

In this way we still respect any `mpi_prefix` constants that the user provided manually, but otherwise get the prefix from the environment.

vyasr:
I agree with @csadorf's change requests. Also, we should be a little more consistent on whether and how we document these methods. Looking at `utils/template_filters.py`, this is already an issue with existing template filters since some have extensive documentation and type information while others do not. I think this should be addressed more broadly as part of a separate issue, but @csadorf would you agree that all template filters should have full function-level documentation? Since they probably won't be meaningfully used outside of a template context it's probably less important that `help` give a reasonable output, but users may want access to them for writing custom templates.

yuanzhou0827:
@csadorf We have changed the name `generate_mpi_prefix` as `get_mpi_prefix`. 

As far as the user defined `mpi_prefix`, it is included by `{% if operation.directives.nranks and not mpi_prefix %}`, we will only set up the `mpi_prefix` when it is not given. I tried `{{ mpi_prefix|default(operation|generate_mpi_prefix) }}`, but there is a problem that if mpi_prefix is not given by user and nranks=0, it will always have a default prefix like `mpirun -n 0` at the beginning. How do like if i did in the way as i did in this new commit?

In addition, Summit is a special case that should always include `mpi_prefix` with some extra_args as the summit.sh at master branch did. @bdice What do you think about this?

flow/templates/base_script.sh
```diff
@@ -16,8 +16,8 @@ cd {{ project.config.project_dir }}
 {% block body %}
 {% set cmd_suffix = cmd_suffix|default('') ~ (' &' if parallel else '') %}
 {% for operation in operations %}
-{% if operation.directives.nranks and not mpi_prefix %}
-{% set mpi_prefix = "%s -n %d "|format(mpiexec|default("mpiexec"), operation.directives.nranks) %}
+{% if not mpi_prefix %}
+{% set mpi_prefix = operation|get_mpi_prefix %}
```
csadorf: Shouldn't this use the `default` filter?

vyasr:
I'm in favor of renaming the filter to `get_prefix`. The relevant prefix can be determined from all directives, making it possible to handle all of the relevant logic in one place, and future-proofs us against other changes to the prefix we might need in the future. @zhou-pj @yuanzhou0827 could you address that request from Simon as well? Once that's done I'll give this another thorough review and we'll be close!

zhou-pj:
@vyasr @csadorf We have addressed all requests so far, ready for review.

flow/templates/base_script.sh
```diff
@@ -16,15 +16,12 @@ cd {{ project.config.project_dir }}
...
-{% if operation.directives.nranks and not mpi_prefix %}
-{% set mpi_prefix = "%s -n %d "|format(mpiexec|default("mpiexec"), operation.directives.nranks) %}
-{% endif %}
 
 # {{ "%s"|format(operation) }}
 {% if operation.directives.omp_num_threads %}
 export OMP_NUM_THREADS={{ operation.directives.omp_num_threads }}
 {% endif %}
-{{ mpi_prefix }}{{ cmd_prefix }}{{ operation.cmd }}{{ cmd_suffix }}
+{{ mpi_prefix|default(operation|get_prefix) }}{{ cmd_prefix }}{{ operation.cmd }}{{ cmd_suffix }}
```
csadorf: Since the environment provides a generic `cmd_prefix`, this should probably operate on `cmd_prefix`, correct? And `mpi_prefix` is obsolete?

csadorf:
@zhou-pj Please re-request review once this is ready, thank you.

flow/environments/incite.py
```diff
@@ -28,24 +28,27 @@ def my_operation(job):
...
                 cores_used += tasks * cpus_per_task
                 gpus_used += gpus
-                if cores_used > cores_per_node or gpus_used > gpus_per_node:
+                if (cores_used > SummitEnvironment.cores_per_node or
+                        gpus_used > SummitEnvironment.gpus_per_node):
                     nodes_used += 1
-                    cores_used = max(0, cores_used - cores_per_node)
-                    gpus_used = max(0, gpus_used - gpus_per_node)
+                    cores_used = max(0, cores_used - SummitEnvironment.cores_per_node)
+                    gpus_used = max(0, gpus_used - SummitEnvironment.gpus_per_node)
```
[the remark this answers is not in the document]
zhou-pj: Currently we are still using these method in jinja template outside the cmd_prefix section and need to keep them as static method. 
But we agree that moving forward when we move to the 'python project.py run' style, at least the get_prefix should be able to be changed to class method. But we may still need to keep calc_num_nodes and guess_resource_sets as static method for jinja template.

zhou-pj:
@csadorf @vyasr 
Ready for re-review.
We made the following changes:
1. Use lambda function to reduce `get_prefix` function duplication and unified the function API
2. Combined OMP, mpi_prefix and cmd_prefix all in the `get_prefix` function and modify jinja template to support backward compatibility
3. Improve docstring and add notes for future reference

flow/templates/base_script.sh
zhou-pj: In current version flow allows the user to provide `mpi_prefix` and `cmd_prefix` separately which allows the user to add additional `cmd_prefix` but still use flow to get proper mpi_prefix.
For now, for the purpose of backward compatibility we think it best to keep both now and the combined prefix is simply called `prefix`. But for future when breaking change is allowed we can maybe change to another consensus that `cmd_prefix` should include all if user were to provide one.

flow/environments/incite.py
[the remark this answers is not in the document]
csadorf: So filter functions must be static? That is very surprising to me. 

Maybe a work-around like `template_environment.filters['method'] = lambda cls: cls.method` could work as a stop-gap measure?

flow/project.py
```diff
@@ -601,7 +601,7 @@ def __init__(self, config=None, environment=None):
         except KeyError:
             self._use_buffered_mode = False
 
-    def _setup_template_environment(self):
+    def _setup_template_environment(self, environment=None):
```
csadorf: Could you help me understand this change?

zhou-pj: Thanks for pointing this out, I was meaning to check with you on this. I need to add this because I notice that for the unittest as well as any mock submit test where self.env and the intended submit environment is not the same, the `_setup_template_environment function` currently won't have the submit environment but it is now needed for `template_environment.filters['get_prefix'] = environment.get_prefix` to get the correct get_prefix function from the actual submit env. I think the change should not break anything but would like your thoughts on this.

flow/project.py
```diff
@@ -649,13 +649,18 @@ def _setup_template_environment(self):
             template_environment.filters['max'] = max
         if 'min' not in template_environment.filters:    # for jinja2 < 2.10
             template_environment.filters['min'] = min
+
+        template_environment.filters['get_prefix'] = \
```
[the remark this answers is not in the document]
zhou-pj: I changed the code but to answer the question, this was to provide the function with the correct environment specific mpi_cmd such as mpiexec, ibrun etc. `template_environment.filters['get_prefix'] = environment.get_prefix` will fail the test as only default 'mpiexec' is used.

I later realize jinja template can actually take in class method in this case so I changed get_prefix to classmethod and directly access cls.mpi_cmd, so this is no longer the issue.

flow/environment.py
```diff
@@ -105,6 +105,7 @@ class ComputeEnvironment(metaclass=ComputeEnvironmentType):
     hostname_pattern = None
     submit_flags = None
     template = 'base_script.sh'
+    mpi_cmd = 'mpiexec'
```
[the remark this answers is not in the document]
zhou-pj: Related to the following comment, this is to reduce duplication of the get_prefix function just to change mpi_cmd string, so they are listed here for accessing the correct one with the lambda function

zhou-pj:
Regarding making the filter function classmethod instead of static, I realize that it may not be a jinja template filter requirement but because 
```filters = {'calc_num_nodes': calc_num_nodes.__func__, 'guess_resource_sets': guess_resource_sets.__func__}```
together with
```for filter_name, filter_function in getattr(environment, 'filters', {}).items(): template_environment.filters[filter_name] = filter_function```
this way of introducing filter that was made for summit that has issue with classmethod.

The way we add template filter with `template_environment.filters['get_prefix'] = environment.get_prefix' directly actually works. So now only `calc_num_nodes` and `guess_resource_sets` remain staticmethod.

On a side note, could you also briefly explain the benefit/rationale in this case to make them classmethod? @csadorf

File: flow/project.py (hunk 6 above)
Comment (csadorf) [reply flow/project.py]: The `FlowProject` instance already has the environment as `self._environment`:
https://github.com/glotzerlab/signac-flow/blob/893017073a9ce0203e03efbdcab75f094e2d6793/flow/project.py#L588 Is this not sufficient?

File: flow/environments/incite.py (hunk 5 above)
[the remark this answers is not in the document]
Comment (csadorf) [reply flow/environments/incite.py]: Is this an option?

File: flow/environments/incite.py [hunk 9]
```diff
@@ -28,24 +28,28 @@ def my_operation(job):
...
         if cores_used > 0 or gpus_used > 0:
             nodes_used += 1
         return nodes_used
 
     @staticmethod
-    def guess_resource_sets(operation, cores_per_node, gpus_per_node):
+    def guess_resource_sets(operation):
         ntasks = max(operation.directives.get('nranks', 1), 1)
```
b-butler: Is this to prevent an error if someone give a `nranks=0`? otherwise the max is superfluous with the get.

flow/environments/xsede.py
```diff
@@ -66,6 +68,41 @@ def add_args(cls, parser):
...
+        :type prefix:
+            str
+        """
+        prefix = ''
+        if operation.directives.get('omp_num_threads'):
+            prefix += 'export OMP_NUM_THREADS={}\n'.format(operation.directives['omp_num_threads'])
+        if mpi_prefix:
+            prefix += mpi_prefix
+        elif operation.directives.get('nranks'):
+            if parallel:
```
b-butler: Is there a way that we could find the offset's ourselves by keeping track of all the requested ranks, and dividing them correctly? We could have a policy to fail if enough ranks aren't requested or that we give the extra ranks to the first operation or something like that.

flow/project.py
zhou-pj: I think the self._environment is set in the initialization of the FlowProject class but when we run submit() we currently allow user input of a different environment:
``` 
def submit(self, bundle_size=1, jobs=None, names=None, num=None, 
parallel=False, force=False, walltime=None, env=None, **kwargs):
..........
if env is None:
    env = self._environment
```
So they may not be the same and we need the environment provided in submit() to provide the correct filter function. I think this is also the way current unit test is setup so without the environment argument from submit only the base ComputeEnvironment.get_mpi will be used.

flow/environments/incite.py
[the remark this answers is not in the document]
zhou-pj: I am not sure I fully understand your suggestion, could you elaborate? On the other hand, I tried another possible solution in the last commit edfa7b754a63490f5894f0da68792c75f37a948c.
It does require the use of exec which I am not sure if it is a good idea. any thoughts?

flow/project.py
csadorf: Got it.

flow/environments/xsede.py
b-butler: My personal thought would be to raise an error, but the offset could be managed automatically regardless by keeping track of the ranks assigned and managing offset according.

csadorf: I'd prefer to generally not change the behavior as part of this patch. So we should just try to emulate the behavior of the template implementation as closely as possible (ideally exactly).

b-butler: Okay, that is fine then. I will resolve the conversation, but I think that would be useful to have in the future when dealing with submitting using run. Having the user specify offsets shouldn't be required.

flow/project.py
zhou-pj: With the PR you added this is no longer needed as template related filter is all added before this function all. Thanks for your help!

flow/environments/incite.py
```diff
@@ -69,20 +74,45 @@ def guess_resource_sets(operation, cores_per_node, gpus_per_node):
...
+        """
+        prefix = ''
+        if operation.directives.get('omp_num_threads'):
+            prefix += 'export OMP_NUM_THREADS={}\n'.format(operation.directives['omp_num_threads'])
+        if mpi_prefix:
+            prefix += mpi_prefix
+        else:
+            extra_args = str(operation.directives.get('extra_jsrun_args', ''))
+            resource_set = cls.guess_resource_sets(
+                           operation)
```
vyasr: Merge the two lines above.

flow/project.py
vyasr: I think this parameter can now be removed because the template filters are added in `_template_environment` instead.

flow/environment.py
```diff
@@ -105,6 +111,7 @@ class ComputeEnvironment(metaclass=ComputeEnvironmentType):
     hostname_pattern = None
     submit_flags = None
     template = 'base_script.sh'
+    mpi_cmd_string = 'mpiexec'
```
csadorf: Why is this not called `mpi_prefix` like everywhere else?

flow/environment.py
```diff
@@ -191,6 +198,40 @@ def get_config_value(cls, key, default=flow_config._GET_CONFIG_VALUE_NONE):
         """
         return flow_config.require_config_value(key, ns=cls.__name__, default=default)
 
+    @template_filter
+    def get_prefix(cls, operation, mpi_prefix=None, cmd_prefix=None, parallel=False):
```
csadorf: I still have trouble following why we need the `mpi_prefix` and `cmd_prefix` parameters here. Do we expect that they will go away eventually? Should they maybe be the last arguments then? Do they need deprecation warnings?

flow/environment.py
zhou-pj: `mpi_cmd_string` is meant for just the system specific mpi syntax like `mpiexec` or `ibrun` or `jsrun` etc. And `mpi_prefix` remain to mean the whole mpi prefix such as `ibrun -n -4`

flow/environment.py
zhou-pj: In order to not break anything we still need them here for cases where user provided `mpi_prefix` or `cmd_prefix`. I believe in the end we might want to move these to directives instead in a later release is we agree on this. So right now I only add a comment in docstring more as a note for future reference.

flow/environment.py
csadorf: Got it, can we rename this to `mpi_cmd` ?

flow/environment.py
csadorf: Can we at least move the `parallel` argument before those two potentially obsolete arguments then? That will make it *much* easier to drop if needed.

flow/environment.py
zhou-pj: that was what i did before but I later found out there is a function called mpi_cmd already for some legacy code in the environment class. Would that be a issue reuse the name? If you think this is ok i can change them to back mpi_cmd.

vyasr: @csadorf the `legacy_templating.py` file indicates that we intended to remove the legacy templating functionality as of version 0.8... which we are currently on. This PR was never intended to be part of the current release of flow, but rather part of the next version 0.9 which is hopefully where groups fit in as well. Should we consider removing the legacy templating prior to merging this PR?

csadorf: Yes.

flow/environment.py
```diff
@@ -191,6 +198,40 @@ def get_config_value(cls, key, default=flow_config._GET_CONFIG_VALUE_NONE):
...
+            User defined cmd_prefix string. Default is set to None.
+            This will be deprecated and removed in the future.
+        :return prefix:
+            The prefix should be added for the operation.
+        :type prefix:
+            str
+        """
+        prefix = ''
+        if operation.directives.get('omp_num_threads'):
+            prefix += 'export OMP_NUM_THREADS={}\n'.format(operation.directives['omp_num_threads'])
```
b-butler: The OpenMP part is the same for all the environments right? I think that having staticmethod helper functions in the base class would allow for `prefix += _open_mp_prefix(operation)` which is shorter and gives a modular foundation for other classes to add special logic to specific parts without duplicating code.

flow/environment.py
```diff
@@ -191,6 +198,40 @@ def get_config_value(cls, key, default=flow_config._GET_CONFIG_VALUE_NONE):
...
+        if mpi_prefix:
+            prefix += mpi_prefix
+        elif operation.directives.get('nranks'):
+            prefix += '{} -n {} '.format(cls.mpi_cmd, operation.directives['nranks'])
```
b-butler: Likewise, the MPI part could be put into a classmethod. These changes would allow for environments with special logic to just implement the MPI part.

zhou-pj: I made the changes you suggested. It actually increased the line of code a little bit because most of the line are docstring boilerplates, but I agree this could make it more modular for future code change for different environment.

zhou-pj:
@csadorf I have addressed all requested changes so far, would you like to give another go before approval?

vyasr:
@csadorf when you have another look, FYI @b-butler and I approved and merged the changes from #208 into this PR. That PR handles #175, so merging this PR will resolve both #174 and #175, leaving just #38 (which is being worked on in #209) as a final prerequisite before revisiting groups (#114).

flow/environment.py
```diff
@@ -191,6 +198,73 @@ def get_config_value(cls, key, default=flow_config._GET_CONFIG_VALUE_NONE):
         """
         return flow_config.require_config_value(key, ns=cls.__name__, default=default)
 
+    @classmethod
+    def _get_omp_prefix(cls, operation):
+        """function for getting the omp prefix based on proper directives.
```
csadorf:
```suggestion
        """Get the OpenMP prefix based on the `omp_num_threads` directive.
```
While *omp* is a legitimate abbreviation for OpenMP, I think we should be a bit more precise on the documentation.

Furthermore, the first line of the doc-string should be a proper sentence and therefore also be capitalized.

flow/environment.py
```diff
@@ -191,6 +198,73 @@ def get_config_value(cls, key, default=flow_config._GET_CONFIG_VALUE_NONE):
...
+        :return omp_prefix:
+            The prefix should be added for the operation.
+        :type omp_prefix:
+            str
+        """
+        return 'export OMP_NUM_THREADS={}\n'.format(operation.directives['omp_num_threads'])
+
+    @classmethod
+    def _get_mpi_prefix(cls, operation, parallel):
+        """function for getting the mpi prefix based on proper directives.
```
csadorf: Please fix doc-string like above.

flow/environments/incite.py
```diff
@@ -69,20 +74,32 @@ def guess_resource_sets(operation, cores_per_node, gpus_per_node):
...
-    def jsrun_extra_args(operation):
-        return str(operation.directives.get('extra_jsrun_args', ''))
-
-    filters = {'calc_num_nodes': calc_num_nodes.__func__,
-               'guess_resource_sets': guess_resource_sets.__func__,
-               'jsrun_options': jsrun_options.__func__,
-               'jsrun_extra_args': jsrun_extra_args.__func__}
+    @classmethod
+    def _get_mpi_prefix(cls, operation, parallel):
+        """function for getting the mpi prefix based on proper directives.
```
csadorf: Please fix doc-string like above.

flow/environments/xsede.py
```diff
@@ -66,6 +68,31 @@ def add_args(cls, parser):
                   'If omitted, uses the system default '
                   '(slurm default is "slurm-%%j.out").'))
 
+    @classmethod
+    def _get_mpi_prefix(cls, operation, parallel):
+        """function for getting the mpi prefix based on proper directives.
```
csadorf: Please fix doc-string like above.

flow/project.py
```diff
@@ -1656,12 +1660,16 @@ def _execute_operation(self, operation, timeout=None):
             or operation.name not in self._operation_functions
             # The specified executable is not the same as the interpreter instance:
             or operation.directives.get('executable', sys.executable) != sys.executable
+            # the operation requires MPI or OpenMP
```
```suggestion
            # The operation requires MPI and/or OpenMP parallelization:
```

PR #188, opened 2019-12-05
Use mistune for terminal output.

csadorf:
Minimal example on how to use mistune for the status terminal output.

This branch requires the following mistune code version: https://github.com/csadorf/mistune/tree/terminal-renderer/mistune

### DIFF
--- a/flow/render_status.py
+++ b/flow/render_status.py
@@ -20,7 +20,7 @@ class renderer:
         self.markdown_output = template.render(**context)
 
     def generate_terminal_output(self, template, context):
-        self.terminal_output = template.render(**context)
+        self.terminal_output = mistune.text(template.render(**context))
 
     def generate_html_output(self, template, context):
         self.generate_markdown_output(template, context)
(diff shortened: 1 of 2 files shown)


zhou-pj:
@csadorf The output looks good to me in the minimal example. I'm going to put it to test on the other output styles and see if any issue comes up.

vyasr:
@csadorf @zhou-pj what's the status of this PR? I already see `mistune.text` used in #163, but the template is slightly different. Were these (or equivalent) changes incorporated separately to that PR using https://github.com/csadorf/mistune? If not, should we go ahead and merge this PR? While we're still figuring out how to get our text renderer merged into mistune, at this point I think we are mostly committed to using mistune in general for Markdown rendering.

zhou-pj:
> @csadorf @zhou-pj what's the status of this PR? I already see `mistune.text` used in #163, but the template is slightly different. Were these (or equivalent) changes incorporated separately to that PR using https://github.com/csadorf/mistune? If not, should we go ahead and merge this PR? While we're still figuring out how to get our text renderer merged into mistune, at this point I think we are mostly committed to using mistune in general for Markdown rendering.

I have incorporated all suggestions in this PR to the #163, and has been editing directly in #163 since. Sorry I should have communicated this on here before.

vyasr:
No problem, thanks for the update. I'll close this PR then.

PR #189, opened 2019-12-06
Implement execution hooks framework

### LINKED ISSUE #14: Proposal for the improved management of provenance and reproducibility
**[Original report](https://bitbucket.org/glotzer/signac-flow/issue/14) by Carl Simon Adorf (Bitbucket: [csadorf](https://bitbucket.org/csadorf), GitHub: [csadorf](https://github.com/csadorf)).**

# About

This is a proposal for the integration of features into signac-flow that would simplify provenance management and increase the chance for reproducibility.
Such features would enable users to keep track of the exact *path* that data took from its source to the current / latest stage.

*Note: This feature depends on and is developed in tandem with the snapshot feature [(signac/issue #69)](https://bitbucket.org/glotzer/signac/issues/69/) for the signac core application.*

# Provenance vs. Reproducibility

First,  we need to clearly define and distinguish what we mean by *data provenance* and *reproducibility*.
The first refers to the ability to know exactly where data came from and which operations were applied to to them, while *reproducibility* means that we can also execute these operations again.

To preserve the **provenance** of the data set, we need to track

1. The original state of the (input) data,
2. the order of operations that were applied to the data set,
3. any parameters that were provided as arguments,
4. the state of the data set before application,
5. the software and hardware environment that was used for execution [1].

To be able to **immediately reproduce** the workflow we also need to

1. Have access to all scripts used for execution,
2. access to the original or an equivalent execution environment (hardware and software) or at least the ability to create such an equivalent environment.

*[1]* The approach for tracking the environment is described in the next section. 

# Approach

The general idea is to create a snapshot before the execution of a data space operation and to store operation related metadata to a log file or a log document entry.
This may be done manually, but could also happen automatically, for example when operations are executed through the `flow.run()` or the `Project.run()` interface.

Such a log entry would have the following fields:
```python
{
  timestamp: float,  # The UTC time in seconds since the UNIX epoch when the operation was triggered.
  snapshot_id: int, # The id of the snapshot taken immediately before execution.
  operation: str,  # The name of the operation.
  cmd: str, # The command used for execution of this operation.
  canonical: bool, # Indicates whether the command is notated in canonical form [1].
  environment: dict,  # Information about the execution environment [2].
  num_ranks: int, # The number of MPI ranks used for execution.
  note: str,  # Optional user-defined remarks or comments.
}
```
A log entry is either appended to a field called `log` as part of the *job document* and/ or written to a collection-based log file.

*[1]* The recorded command is assumed to be executed within the project root directory and should ideally be in *canonical notation*.
A canonical command is a command that is a function of job and is therefore easily *transferable* between jobs.
For example, the user may have executed operation `my_op` in parallel using the `flow.run()` interface with the command `python operations.py my_op`, however, the canonical command would be `python operations.py my_op {job._id}`.
All commands executed through the `Project.run()` and the `flow.run()` interface would automatically be canonical, other commands might require some extra treatment.

**CAVEAT**: The actual command used for execution can not always be accurately preserved.
For example, it is possible to record the number of MPI ranks used for execution, but it is not possible to automatically record which `mpirun` wrapper variant was actually used for execution and with which parallelization layout.
We are assuming that this is desired behavior and that the operations logic is not a function of the parallelization layout.

*[2]* Capturing the hardware and software (library) environment accurately and comprehensively is only possible under specific circumstances, *e.g.*, when using a containerized environment. Existing tools such as `pip freeze` are capable to capture most software versions within the current python environment, however local extensions to the python path or other locally installed packages may not be captured.
Non-python tools are even harder to automatically detect.
We therefore propose a *good enough approach*, where it is mostly user responsibility to ensure that all *essential* environment parameters are accurately tracked, but the framework assists with that as much as reasonably possible.
See below for the proposed API on how to track the environment.

# Proposed API

## Keeping an operations log

Logging of operations would be enabled in the following ways:

 1. Manually via the *internal* `flow._log_operation()` function,
 2. Manually via the *low-level* `flow.log_operation()` functor,
 3. through arguments to `flow.run()`,
 4. through arguments to `Project.run()`.

The API would also be exposed on the command line and the configuration, but for the sake of brevity, we are going to focus on the python API here.

### Manual logging

This is how we would create a manual log entry:
```python
flow.log_operation(
    # An instance of JobOperation
    operation,
    # Indicate whether the command is in canonical form (see above) (default: False).
    canonical=False,
    # Generate snapshots before execution if True (default: True).
    snapshot=True,
    # The field within the job document that contains the operations log.
    # None means 'do not store' (default: 'log').
    field='log',
    # The filename of a log file relative to the job's workspace path.
    file='operations.log',
    # An additional user-defined comment (default: None).
    note="Executed through aprun.",
    # Restore the job if the execution of the operation fails (default: True) [1].
    restore_on_fail=True,
): {<log entry>} (dict)
```
 * The `file` argument can alternatively be a *file-like object*.
 * The `snapshot` argument may optionally be a mapping, to be able to forward arguments to the snapshot function, *e.g.*, `{deep': True}`.

Manual tracking is useful where operations are not executed through the `flow` interface. It is up to the user to ensure that the entry is correct and that the operation is actually executed after creating the log entry.

*[1]* The `log_operation()` functor may be used as *context manager*, where the log entry is only created if the execution of the operation actually succeeded, otherwise the snapshot is restored:
```python
op = JobOperation('myop', job, './src/my_op {}'.format(job))
with log_operation(op):
    subprocess.run(op.cmd)
```
This would be roughly equivalent to:
```python
op = JobOperation('myop', job, './src/my_op {}'.format(job))
log_entry = log_operation(op, field=None)
try:
    subprocess.run(op.cmd)
except:
    project.restore(_id=log_entry['snapshot_id'])
    raise
else:
    flow._log_operation(log_entry)
```

### Semi-automatic logging through the *flow* interface

Alternatively, we can use the extended `flow.run()` interface to keep track:
```python
flow.run(
    # An optional sequence of job-operations (default: None).
    operations=None,
    # The field within the job document that contains the operations log.
    # None means 'do not store' (default: 'log').
    log='log',
    # Specify a log file name relative to the job's workspace (default: None).
    logfile='operations.log',
    # Generate snapshots before execution if True (default: True).
    snapshot=True,
    # An optional parser argument to add the run CLI arguments to (default: None).
    parser=None,
    # Additional keyword arguments to be forwarded to the operations function.
    **kwargs
    )
```

 * When the first argument is `None`, the command line interface is started (replicating previous behavior).
 * The `logfile` argument can alternatively be a *file-like object*.
 * The `snapshot` argument may optionally be a mapping, to be able to forward arguments to the snapshot function, *e.g.*, `{deep': True}`.

Using the `flow.run()` interface, the logged command would be logged in canonical form for the execution of a single operation for a single job, that means even if the operation was actually executed for all jobs like this:
`python operations.py my_op`, it would be logged as `python operations.py my_op {job._id}`.

## Keeping track of the hardware and software (library) environment

The capturing of environment parameters would be embedded into the existing `flow.environment` structure.
The `DefaultEnvironment` would be extended to capture a few parameters that are always accessible, such as the python version environment or information provided by the [platform module](https://docs.python.org/3/library/platform.html).
For everything else, user input is needed.
The current definition of the `Environment` class could be extended in this form:
```python
class MyEnvironment(DefaultEnvironment):

    MyEnvironment.track_version('gcc', cmd='gcc --version')
```
Some tracking may also be enabled through the configuration:
```
[flow]
[[environment]]
[[[track]]]
[[[[gcc]]]]
cmd="gcc --version"
```

# Example Usage

Listing all operations that have been applied to a specific job from the beginning to now, would be trivial:
```python
pprint(job.document['log'])
```

Assuming we wanted to replicate the workflow applied to `a_job` to `another_job`, and that commands have been recorded *canonically*, this is a possible approach:
```python
log = a_job.document['log']
for ts in sorted(log):
    subprocess.run(log[ts]['cmd'].format(job=another_job))
```
A *canonical* command is a function of the job's state point and document, *e.g.*: `./src/operations.py my_op {job._id}`

The logic implemented above would also be implemented as a *high-level* function, for instance to **replay** the exact same sequence of operations:
```python
# Execute with subprocess.run():
for job_op in flow.replay(a_job.document['log'], job=another_job):
    subprocess.run(job_op.cmd)

# Execute with flow.run():
flow.run(flow.replay(job.document['log']), job=another_job)
```

### LINKED ISSUE #28: Proposal for operation hooks
**[Original report](https://bitbucket.org/glotzer/signac-flow/issue/28) by Carl Simon Adorf (Bitbucket: [csadorf](https://bitbucket.org/csadorf), GitHub: [csadorf](https://github.com/csadorf)).**

# About

This is a proposal for a feature that would allow users to specify hook functions that are executed either before or after the execution of an operation.
The purpose is to simplify the execution of specific functions before and after the execution of a job-operation, such as logging or data upload functions.

A hook function takes an instance of `JobOperation` as first argument which allows to access the name of the operation, the job it operates on, the command, and optional execution directives.
For example:
```
#!python
def log_operation(op):
    log("Operation {} started.".format(op))
```

This log function could then be specified to be executed on operation start:
```
#!python

@Project.operation
@Project.hook.on_start(log_operation)
def hello(job):
    print('hello', job)
```

# API

Hooks are in general defined as lists of functions, as part of the `flow.hooks.Hooks` class.

- `on_start`: Executed *before* a operation is executed.
- `on_finish`: Executed *after* a operation has completed (with or without error).
- `on_success`: Executed *after* a operation has completed without error.
- `on_fail`: Executed *after* a operation has completed with error.

## Low-level API

The low-level API allows the installation of hooks on the project instance level *via* the manipulation of the `hooks` instance variable:
```
#!python
project = Project()
project.hooks.on_start.append(my_log_function)
```
And during the adding of operations, again, via a dictionary:
```
#!python
project.add_operation(
    name='hello',
    cmd='echo hello {job._id}',
    hooks={'on_start': [my_log_function]})
```

The hooks are executed in order of definition, where project-wide hooks are executed *prior* to operation-wide hooks.

## Decorator API

Alternatively, hooks can be installed *via* the decorator API:

```
#!python
from flow import hook

@Project.hook.on_start(log_operation)
def my_op(job):
    pass
```
Note, that the hooks are only executed if the operation is executed through the `FlowProject` interface.

## Configurational API

Finally, hooks can be installed directly in the signac configuration by specifying a function that installs the hooks during project initialization.

```
#!bash
# signac.rc

[flow]
hooks = my_hooks.install_hooks
```
The `hooks` config key is interpreted as list, multiple modules would be provided as a comma-separated list.
The `my_hooks` module must be directly importable and should have a `install_hooks()` function in the global namespace similar to this:
```
#!python
# my_hooks.py

def log_operation(op):
    log("Operation {} started.".format(op))

def install_hooks(project):
    project.hooks.on_start.append(log_operation)
    return project
```

The `install_hooks()` function should return the project argument so that we can use the following idiom: `install_hooks(Project()).main()`.

The default hook installation function is called *install_hooks* so in this case the following configuration would be equivalent:
```
#!bash
[flow]
hooks = my_hooks
```
As a final example, assuming that the `flow` package had a module that would install hooks for the automatic generation of *snapshots* after execution, it would likely be installed like this:
```
#!bash
[flow]
hooks = flow.snapshots
```

csadorf:
## Description

This pull request implements an execution hooks framework that enables users to execute certain functions *on start*, *on finish*, *on success*, and *on fail* of the execution of an operation.

Resolves #28, resolves #14.

## Motivation and Context

The motivation for this framework is to enable users to automatically execute certain functions with operations for example for logging and tracking purposes. For example, to log when a specific operation is executed, a user can provide the following function:

```python
@Project.operation
@Project.hooks.on_start(lambda op: logger.info(f"Executing {op}..."))
def foo(job):
    pass
```

The framework comes with a selection of pre-implemented hook systems to cover all the currently expected use cases, roughly in order of expensiveness:

1. `LogOperation` - Log basic information about the execution of operations to a log file.
2. `TrackOperations` - Keep detailed metadata records about the state of the project root directory and the operation, including directives, to a log file, optionally in conjunction with git.
3. `SnapshotProject` - Create a snapshot of the project root directory to keep track of the code used for the execution of an operation.
4. `TrackWorkspaceWithGit` - The workspace is treated as a git repository and automatically committed to before and after the execution of an operation. The can be done either on a per-job basis or workspace global.

A *hook system* is meant to describe a collection of hook functions that together achieve a specific purpose. The shipped hook systems are implemented as classes that can be installed project-wide via the respective `install_hooks` function.

## High-level API

Hooks can be installed either on a per-operation or a per-project basis. In this way users have the option to execute hook functions either with specific operations or with all operations of a project.

Furthermore, there are two ways that hooks can be installed. One way is directly in Python, for example within the `__main__` clause of the `project.py` file:
```python
# ...

if __name__ == '__main__':
    from flow.hooks import LogOperations
    LogOperations().install_hooks(Project()).main()
```
Alternatively, hooks can also be installed through the (project) configuration file:
```
project = my_project
[flow]
hooks=flow.hooks.LogOperations,flow.hooks.SnapshotProject(compress=True)
```
It is assumed that the entities provided here, are callable, i.e., are either a function or a functor. The first argument must be the instance of `FlowProject`.

## Low-level API

The `FlowProject` class has a `hooks` attribute with four lists: `on_start`, `on_finish`, `on_success`, and `on_fail`, which can be appended to. Hook functions installed in this way are executed for *all* operations.

Furthermore, hook functions can be installed for individual operations either with a decorator:
```python
@Project.operation
@Project.hook.on_start(my_hook_function)
def op(job):
    pass
```
or by passing a dictionary to the `add_operation()` function: `project.add_operation(..., hooks=dict(on_start=my_hook_function))`.

## Custom hook systems

Users can very easily implement their own hook systems and install them in similar manner.
For example:
```python
# myhooks.py
def log_op(operation):
    with open(operation.job.fn('operations.log'), 'a') as logfile:
        logfile.write(f"Executed operation '{operation}'.")

def install(project):
    project.hooks.on_success.append(log_op)
    return project
```

This could then be installed either in the project module:
```python
# ...

if __name__ == '__main__':
    import myhooks
    myhooks.install(Project()).main()
```
or via the configuration file:
```
project = myproject
[flow]
hooks=myhooks.install
```

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


csadorf:
@bdice You were not assigned as reviewer, but might still be interested in this feature.

csadorf:
Note: This feature won't work with "exec operations", but since these are going away anyways, I think it's safe to introduce this feature regardless.

csadorf:
Resolves issue #28 .

flow/hooks/git_workspace_tracking.py
```diff
@@ -0,0 +1,115 @@
...
+
+_WARNINGS = defaultdict(set)
+
+
+def _ignore_hidden_files(root):
+    "Do not commit hidden files to the git repository."
+    with open(os.path.join(root, '.gitignore'), mode='a+') as file:
+        file.seek(0)
+        if '.*\n' not in file.readlines():
+            file.write('.*\n')
```
vyasr: I would add a logger entry or something here (or maybe in the `_get_or_init_git_repo` function`, just update the existing log entry to say that the gitignore file is being initialized to ignore hidden files) to indicate to users that a repo didn't exist, so you're making one and automatically ignoring hidden files.

vyasr: It may also make sense for this to be an option (which is what we discussed the last time around with this PR). However, since it's relatively deep in the call stack and hidden from the user, just notifying the user loudly is probably good enough (and they can change it if they want), rather than trying to provide an API.

flow/hooks/git_workspace_tracking.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def _commit(repo, title):
+    try:
+        repo.git.commit('-m', GIT_COMMIT_MSG.format(title=title))
+    except git.exc.GitCommandError as error:
+        if "nothing to commit, working tree clean" in str(error):
+            pass
+        else:
+            raise
```
vyasr: This will also raise an error if there are *untracked* files in the repository (see [here](https://bitbucket.org/glotzer/signac-flow/pull-requests/35/execution-hooks/diff#comment-72666460))

flow/hooks/git_workspace_tracking.py
```diff
@@ -0,0 +1,115 @@
...
+            raise RuntimeError(
+                "Unable to get project source git commit id, repository is dirty.")
+        else:
+            if project.root_directory() not in _WARNINGS['git_dirty']:
+                logger.warn("Unable to get project source git commit id, repository is dirty.")
+                _WARNINGS['git_dirty'].add(project.root_directory())
+    return {
+        'commit_id': str(repo.commit()),
+        'dirty': repo.is_dirty(),
+    }
```
vyasr: Is this function meant to be a distinct hook for users? It's not used by the `TrackWorkspaceWithGit` class so I'm not quite sure what its purpose is.

flow/hooks/git_workspace_tracking.py
```diff
@@ -0,0 +1,115 @@
...
+        if self._per_job:
+            return _get_or_init_git_repo(root=operation.job.workspace())
+        else:
+            return _get_or_init_git_repo(root=operation.job._project.workspace())
+
+    def commit_before(self, operation):
+        metadata = collect_metadata(operation)
+        metadata['stage'] = 'prior'
+        repo = self._get_repo(operation)
+        repo.git.add(A=True)
```
vyasr: I think `u=True` is probably more appropriate, right? Users could put temporary testing scripts, or worse, temporary data files, into the root of the repository. I would prefer automatic adding to be opt-in, such that `commit_before` returns a callable and has an argument that controls whether the returned callable does a `git add -u` or a `git add -A`. I'm happy to brainstorm the best API to support such parametrizable hooks, I assume it would just be through a constructor argument of the class similar to the snapshot tracking hook class.

vyasr: If we do stick with this behavior, then my concern with respect to untracked files in `_commit` is moot.

flow/hooks/util.py
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) 2018 The Regents of the University of Michigan
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+import time
+
+
+def collect_metadata(operation):
+    return {
+        '_schema_version': 1,
```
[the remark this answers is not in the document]
csadorf: I've updated the schema to include both and added comments for clarification.

flow/hooks/util.py
```diff
@@ -0,0 +1,21 @@
...
+        'time': time.time(),
+        'project': {
+            'path': operation.job._project.root_directory(),
+            'git': None,
```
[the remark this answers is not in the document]
csadorf: Agreed, removed.

flow/hooks/git_workspace_tracking.py
```diff
@@ -0,0 +1,115 @@
+# Copyright (c) 2018 The Regents of the University of Michigan
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+from __future__ import print_function
```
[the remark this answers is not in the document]
csadorf: Tx.

flow/hooks/log_operations.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def log_operation(self, operation, error=None):
+        if str(operation.job) in self._loggers:
+            logger = self._loggers[str(operation.job)]
+        else:
+            logger = self._loggers[str(operation.job)] = logging.getLogger(str(operation.job))
+            logger.setLevel(logging.DEBUG)
+
+            fh = logging.FileHandler(operation.job.fn(self._fn_logfile))
+            fh.setLevel(logging.DEBUG)
```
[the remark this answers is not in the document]
csadorf: The handler handles *all* calls to the logging framework and will log *all* messages, including those on the *DEBUG* level.

See here: https://docs.python.org/3/howto/logging.html#handlers

csadorf:
@vyasr Thank you very much for your comprehensive feedback. I'm working on addressing the issues. I was wondering whether it might make sense to split this up into introducing the framework and the individual "bundled" hook systems separately? In this way we don't stall the whole PR just because there some issues with the individual systems.

vyasr:
I like the idea of splitting the two PRs. However, I suspect that developing the framework is much easier if you have these example hooks to test with, so I'll leave it up to you to decide whether you want to do that split now or at the end. Since we squash PR commits, I would be fine with working on everything here until we're happy with the framework, then moving the bundled hooks to a different branch and deleting them from this one. That will lead to some scattering of comments relating to the example hooks, but I think it's a reasonable tradeoff.

flow/project.py
```diff
@@ -686,6 +720,33 @@ def _show_template_help_and_exit(self, template_environment, context):
...
+        return self._hooks
+
+    def _install_config_hooks(self):
+        try:
+            hooks_config = self._config['flow'].as_list('hooks')
+        except KeyError:
+            return  # no hooks configured
+        for entry in hooks_config:
+            try:
+                m = re.match(r'^(?P<path>[\w.]+?)(\((?P<constructor>.*)\))?$', entry)
```
mikemhenry: would it make sense to compile this regex (outside of the for loop) then use it for searches in the loop?

csadorf: No, the `re` package has been optimized to automatically cache those and this loop would only be executed a few times, not hundreds or thousands of times anyways.

flow/hooks/git_workspace_tracking.py
csadorf: I don't think that makes sense, because then the user would need to manually switch into all workspace directories and add files.

flow/hooks/git_workspace_tracking.py
csadorf: Shouldn't be an issue, because we should automatically add all files (see response below).

csadorf:
I merged master and I think I addressed all comments, however, this is still missing documentation. Let me know if we can move forward with this, and then I'll prepare that.

flow/hooks/git_workspace_tracking.py
vyasr: I'm fine with tracking the workspace automatically. I'm less comfortable with tracking the project root automatically since I think that's much more susceptible to user clutter, but in the interests of reproducibility I guess it's fine to force the user to think about this and do their own cleanup rather than worrying about it ourselves.

flow/hooks/git_workspace_tracking.py
```diff
@@ -0,0 +1,101 @@
...
+            _commit(repo, "Executed operation {}.\n\nThe execution failed "
+                          "with error '{}'.".format(operation, error))
+        else:
+            _commit(repo, "Executed operation {}.".format(operation))
+        repo.git.notes('append', repo.commit(), '-m', 'signac:' + json.dumps(metadata))
+
+    def install_hooks(self, project):
+        project.hooks.on_start.append(self.commit_before)
+        project.hooks.on_success.append(self.commit_after)
+        project.hooks.on_fail.append(self.commit_after)
```
vyasr: Why not just do `on_finish` rather than do both of these?

flow/hooks/log_operations.py
```diff
@@ -0,0 +1,38 @@
...
+class LogOperations(object):
+
+    def __init__(self, fn_logfile='.operations.log'):
+        self._fn_logfile = fn_logfile
+        self._loggers = dict()
+
+    def log_operation(self, operation, error=None):
+        if str(operation.job) in self._loggers:
+            logger = self._loggers[str(operation.job)]
+        else:
```
vyasr: Why not use a try-except here?

flow/hooks/log_operations.py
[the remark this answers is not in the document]
vyasr: But it only affects loggers that you have explicitly added the logger to via `logger.addHandler`, which in this case is `self._loggers[str(operation.job)]`, right? It won't affect other loggers, and these loggers are only used internally by this class when `log_operation` is called.

csadorf:
@vyasr Could you resolve conversations that you consider resolved?

vyasr:
Done! Sorry, I meant to do that during the review but forgot by the end of it :)

csadorf:
> Done! Sorry, I meant to do that during the review but forgot by the end of it :)

No problem, thx!

vyasr:
@csadorf is planning to try to wrap this up by May 8th. If he cannot, at that point we've agreed that someone else should take over this project.

mikemhenry:
I feel like there is a lot of utility with this feature. I'm not able to take ownership and finish the effort, but I'm willing to provide review or test things out if it is at that stage.

vyasr:
Simon and I would like to pair program and finish this, but my data meeting (pre-defense meeting) is scheduled for August 18th so I won't be available at least until then.

vyasr:
@cbkerr @klywang thanks for handling the merge! Here are the next steps:
1. Install pre-commit hooks and run them locally to fix style. Then tests should start passing.
2. Go through all my outstanding inline comments. A number of them have yet to be addressed. At this stage I would only address the simpler ones (reformatting, renaming, etc).
3. Write up documentation describing the hooks (we can keep working on what you two started earlier). While doing so, try and address conceptually some of the bigger questions I asked Simon during my review. In particular, I think it will be very important to have clear distinctions between "hooks" and "hook systems", where the latter is effectively a collection of hooks along with a means for installing them. This also needs clearer delineation of operation- vs project-specific hooks.
4. Rewrite the code as needed.

flow/project.py
```diff
@@ -3161,13 +3244,19 @@ def _register_operations(self):
...
             # Construct FlowOperation:
             if getattr(func, '_flow_cmd', False):
                 self._operations[name] = FlowCmdOperation(cmd=func, **params)
             else:
                 self._operations[name] = FlowOperation(name=name, **params)
                 self._operation_functions[name] = func
 
+            # Update operation hooks
+            self._operation_hooks[name].update(Hooks.from_dict(self.hook[func]))
+
```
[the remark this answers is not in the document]
cbkerr: Fixed in our merge

flow/hooks/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+            super(Hooks, self).__setattr__(name, _HooksList(value))
+        else:
+            super(Hooks, self).__setattr__(name, value)
+
+    @classmethod
+    def from_dict(cls, mapping):
+        "Instantiate the hook class from a dictionary."
+        hook = cls()
+        for key in mapping:
+            getattr(hook, key)[:] = list(mapping[key])
```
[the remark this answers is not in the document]
cbkerr: Improved naming in commit dc3bcdb888c14fb25730d72b0e87ae758b02fee4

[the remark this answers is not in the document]
klywang: When we try to add an operation using the following code:
```python
from flow import FlowProject

def my_hook_func(operation):
    print(f"Start {operation.name}")

p = FlowProject()
p.add_operation('hello', cmd='echo $PATH', hook=dict(on_start=my_hook_func))
p.main()
```

and we get the following error:
```cmd
  File "/home/kelwang/Software/signac-flow/scratch.py", line 21, in <module>
    p.add_operation('hello', cmd='echo $PATH', hooks=dict(on_start=my_hook_func))
  File "/home/kelwang/Software/signac-flow/flow/project.py", line 4054, in add_operation
    self._operation_hooks[name].update(Hooks.from_dict(hooks))
  File "/home/kelwang/Software/signac-flow/flow/hooks/__init__.py", line 70, in from_dict
    getattr(hook, trigger_type)[:] = list(mapping[trigger_type])
TypeError: 'function' object is not iterable
```

When you suggested your change, did you mean that we should change `list(mapping[trigger_type])` to something else? 

Note: This works fine if we set up hooks using decorators and it worked without error.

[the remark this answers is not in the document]
vyasr: @klywang this isn't your fault, it looks to be a bug. I've traced the problem and will try to propose a fix, but no need for you to worry about this bit for now.

csadorf:
@cbkerr Is this ready for another review?

flow/hooks/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+    "Define execution hooks that are executed on certain conditions."
+
+    _hook_triggers = [
+        "on_start",
+        "on_finish",
+        "on_success",
+        "on_fail",
+    ]
+
+    def __init__(self, **kwargs):
```
bdice, line 55: I think we could make this more readable by avoiding `**kwargs` and defining all the hooks explicitly.

flow/hooks/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+        else:
+            super().__setattr__(name, value)
+
+    @classmethod
+    def from_dict(cls, mapping):
+        "Instantiate the hook class from a dictionary."
+        hook = cls()
+        for trigger_type in mapping:
+            getattr(hook, trigger_type)[:] = list(mapping[trigger_type])
+        return hook
```
bdice, line 71: Could this alternate constructor be replaced by an unpacking with the normal constructor?
```
Hooks.from_dict(d)
Hooks(**d)
```

flow/project.py
```diff
@@ -305,6 +308,25 @@ def __iter__(self):
...
 
+class _HooksDecorator:
+    def __init__(self):
+        self._operation_hooks = defaultdict(lambda: defaultdict(list))
+
+    def __getitem__(self, func):
+        return self._operation_hooks[func]
+
+    def __getattr__(self, name):
+        class install_hook:
```
bdice, line 319: This should be 
```suggestion
        class _InstallHook:
```

bdice:
I read through some of this PR with @klywang. We're scheduling a time to dig deeper on this and consider the architecture of the code. I have two broad considerations:

- It seems like this PR's current design has prioritized _generality_ at the expense of readability and document-ability. I think there are some clear tradeoffs where I would prefer a less-general and more-specific way. Two examples:
  1. The supported hooks are given as a list `["on_start", ..., "on_error"]` hook attributes are assigned dynamically with `setattr`. I think it would be more readable to avoid dynamic attributes and just use `self.on_start = ...`. The supported hook types seem more like `pre` and `post` (a short list of known elements) rather than an arbitrary list that would require dynamic assignments.
  2. Some classes use `**kwargs` and define alternate constructors, rather than having a single path of entry with specifically-named parameters (and calling code that uses those names). This has a slight cost in generality but offers much-more readable code in my opinion.

- I think we may be able to simplify the "registration" side of this PR after we work through the follow-up implications of #487 and #382 and begin to reconcile class-level and instance-level data members and try to simplify FlowProject instantiation. I would like to think through those problems first, before diving into this PR further, because I think the simplifications will pay off.

flow/hooks/git_workspace_tracking.py
csadorf: I think you're right, that should be equivalent.

flow/project.py
```diff
@@ -3161,13 +3244,19 @@ def _register_operations(self):
                 'pre': pre_conditions.get(func, None),
                 'post': post_conditions.get(func, None)}
 
+            # Update operation hooks
+            self._operation_hooks[name].update(Hooks.from_dict(self.hook[func]))
```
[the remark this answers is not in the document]
csadorf: If I remember correctly, that's by design. I always considered hooks to be registered on a *per-instance* level, and not to be inherited. That's because hooks are something that executed for a particular runtime and unlike the DAG, not something that is *inherent* to a project's workflow.

vyasr:
@bdice @cbkerr @klywang most of my comments are still relevant after #482 and addressing them should make the merge easier later. I'd prioritize docs/addressing open discussions over further refactoring work on this branch while other work happens on the main branch. I definitely agree that we should aim to simplify before finalizing this feature, but I think writing docs should come first so that @cbkerr and @klywang can clearly articulate the goals of this PR first since understanding how it should work is critical to writing the best version of the code.

setup.py
```diff
@@ -57,6 +57,7 @@
     entry_points={
         "console_scripts": [
             "flow = flow.__main__:main",
+            "flow-track-operations = flow.hooks.track_operations.__main__:main",
```
bdice, line 60: This main function doesn't exist.
```suggestion
```

flow/hooks/__init__.py
bdice: Resolved in #508.

flow/project.py
bdice: Resolved in #508.

bdice:
Now that #508 has been merged, we need to decide whether to implement the suggested hooks from this PR. I think the best way to do that would be to close this PR and open a new issue listing the proposed hooks and linking to this PR for reference implementations that could be adapted in separate PRs.

>The framework comes with a selection of pre-implemented hook systems to cover all the currently expected use cases, roughly in order of expensiveness:
> 1. `LogOperation` - Log basic information about the execution of operations to a log file.
> 2. `TrackOperations` - Keep detailed metadata records about the state of the project root directory and the operation, including directives, to a log file, optionally in conjunction with git.
> 3. `SnapshotProject` - Create a snapshot of the project root directory to keep track of the code used for the execution of an operation.
> 4. `TrackWorkspaceWithGit` - The workspace is treated as a git repository and automatically committed to before and after the execution of an operation. The can be done either on a per-job basis or workspace global.

bdice:
> @bdice This can be closed, right?

The only reason we left it open was because this PR also implements the specific hook classes that were not included in the general framework that was merged.

b-butler:
Closed see #637 for further discussion.

PR #191, opened 2019-12-12
Define filters by reference.

csadorf:
And thus avoid the need to use 'exec'.


csadorf:
This doesn't actually work.

PR #197, opened 2019-12-13
Prevent failure on operation creation when inspection fails

b-butler:
## Description

In `guess_cmd`, for certain `functool` objects, `inspect.getsourcefile` can fail which prevents the creation of the operation. A fix for this is to allow  setting `_flow_path`  to avoid calling `inspect.getsourcefile`. This just changes the logic's flow to allow for `_flow_path` to prevent `inspect.getsourcefile` be called.

## Motivation and Context

See above

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -2585,7 +2585,7 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
                     path = inspect.getsourcefile(inspect.getmodule(func))
                 except TypeError:
                     raise RuntimeError("Cannot find file associated with "
-                                       "function {}. Specify `_func_path` "
+                                       "function {}. Specify `_flow_path` "
                                        "to use.".format(name))
             cmd_str = "{} {} exec {} {{job._id}}"
 
--- a/flow/project.py
+++ b/flow/project.py
@@ -2578,7 +2578,15 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
             except (KeyError, TypeError):
                 executable = sys.executable
 
-            path = getattr(func, '_flow_path', inspect.getsourcefile(inspect.getmodule(func)))
+            if hasattr(func, '_flow_path'):
+                path = func._flow_path
+            else:
+                try:
+                    path = inspect.getsourcefile(inspect.getmodule(func))
+                except TypeError:
+                    raise RuntimeError("Cannot find file associated with "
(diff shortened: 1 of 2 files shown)


flow/project.py
```diff
@@ -2578,7 +2578,15 @@ def _guess_cmd(func, name, **kwargs):
...
 
-            path = getattr(func, '_flow_path', inspect.getsourcefile(inspect.getmodule(func)))
+            if hasattr(func, '_flow_path'):
+                path = func._flow_path
+            else:
+                try:
+                    path = inspect.getsourcefile(inspect.getmodule(func))
+                except TypeError:
+                    raise RuntimeError("Cannot find file associated with "
+                                       "function {}. Specify `_func_path` "
```
bdice: Typo?
```suggestion
                                       "function {}. Specify `_flow_path` "
```

b-butler:
@bdice I will give a more complete answer later, but in part the problem is that currently `flow` assumes that `Operations` are defined in the file that `Project.main()` is called from (which can/has caused errors). This is fixed in the `groups` branch with its new command generation model. A temporary solution could be had, but it may need to be removed when `FlowGroup` enters the scene. Currently, to generate a valid command, the `FlowProject` attempts to use the object's source file given by `inspect`. The behavior can be overwritten by setting `_flow_path`. However,  on `master` this would still cause error for objects that raise an error with `inspect.getsourcefile` since `getattr` with a default is used.

vyasr:
@b-butler how did you trigger this error? I agree that this problem is fixed on the groups branch, and I don't think any temporary solution is required between now and when groups are merged unless you're encountering something significant. Are you specifically running into this because you're importing operations from one file into another?

@bdice the `_flow_path` pretty much exists for development, debugging, and testing. It already existed when I wrote the template tests, and I took advantage of it in that case for the mocked project creation. It should continue to exist, but not as part of any public API.

b-butler:
@vyasr I was trying to use `functools.partial`. Importing wouldn't work either if executable was set. Of course, if you don't fork then it all works regardless. Also, I think with the idea of an `entrypoint` `_flow_path` may be able to be removed.

b-butler:
Since this bug doesn't seem to effect many users (I know of one other case than mine where this didn't work), I am closing this PR. The bug is fixed on the `groups` branch in #114

PR #203, opened 2019-12-23
Improve clarity of docstrings.

bdice:
## Description
I wanted to better familiarize myself with the code base of signac-flow, so I read through most of the files. While doing so, I found several opportunities to improve the clarity of docstrings and fix typos. I also made some minor changes to the code itself.

## Motivation and Context
General improvements to docs.

## Types of Changes

- [x] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:
- [ ] I have updated the [changelog](https://github.com/glotzerlab/signac-flow/blob/master/changelog.txt). _(Not needed for these kind of small changes.)_

### DIFF
--- a/flow/environment.py
+++ b/flow/environment.py
@@ -40,9 +40,9 @@ def setup(py_modules, **attrs):
     """Setup function for environment modules.
 
     Use this function in place of setuptools.setup to not only install
-    a environments module, but also register it with the global signac
-    configuration. Once registered, is automatically imported when the
-    get_environment() function is called.
+    a environment's module, but also register it with the global signac
+    configuration. Once registered, the environment is automatically
+    imported when the get_environment() function is called.
     """
     import setuptools
     from setuptools.command.install import install
@@ -73,9 +73,9 @@ def setup(py_modules, **attrs):
 
 
 class ComputeEnvironmentType(type):
-    """Meta class for the definition of ComputeEnvironments.
+    """Metaclass for the definition of ComputeEnvironments.
 
-    This meta class automatically registers ComputeEnvironment definitions,
+    This metaclass automatically registers ComputeEnvironment definitions,
     which enables the automatic determination of the present environment.
     """
 
@@ -91,10 +91,10 @@ class ComputeEnvironment(metaclass=ComputeEnvironmentType):
     """Define computational environments.
 
     The ComputeEnvironment class allows us to automatically determine
-    specific environments in order to programatically adjust workflows
+    specific environments in order to programmatically adjust workflows
     in different environments.
 
-    The default method for the detection of a specific environemnt is to
+    The default method for the detection of a specific environment is to
     provide a regular expression matching the environment's hostname.
     For example, if the hostname is my-server.com, one could identify the
     environment by setting the hostname_pattern to 'my-server'.
@@ -123,7 +123,7 @@ class ComputeEnvironment(metaclass=ComputeEnvironmentType):
 
     @classmethod
     def get_scheduler(cls):
-        """Return a environment specific scheduler driver.
+        """Return an environment-specific scheduler driver.
 
         The returned scheduler class provides a standardized interface to
         different scheduler implementations.
@@ -167,8 +167,7 @@ class ComputeEnvironment(metaclass=ComputeEnvironmentType):
         """Request a value from the user's configuration.
 
         This method should be used whenever values need to be provided
-        that are specific to a user's environment. A good example are
-        account names.
+        that are specific to a user's environment, e.g. account names.
 
         When a key is not configured and no default value is provided,
         a :py:class:`~.errors.SubmitError` will be raised and the user will
--- a/flow/environment.py
+++ b/flow/environment.py
@@ -40,9 +40,9 @@ def setup(py_modules, **attrs):
     """Setup function for environment modules.
 
     Use this function in place of setuptools.setup to not only install
-    a environment's module, but also register it with the global signac
+    an environment's module, but also register it with the global signac
     configuration. Once registered, the environment is automatically
-    imported when the get_environment() function is called.
+    imported when the :py:meth:`~.get_environment` function is called.
     """
     import setuptools
     from setuptools.command.install import install
@@ -362,10 +362,10 @@ def registered_environments(import_configured=True):
 def get_environment(test=False, import_configured=True):
     """Attempt to detect the present environment.
 
-    This function iterates through all defined ComputeEnvironment
-    classes in reversed order of definition and and returns the
-    first EnvironmentClass where the is_present() method returns
-    True.
+    This function iterates through all defined :py:class:`~.ComputeEnvironment`
+    classes in reversed order of definition and and returns the first
+    environment where the :py:meth:`~.ComputeEnvironment.is_present` method
+    returns True.
 
     :param test:
         Whether to return the TestEnvironment.
--- a/flow/environment.py
+++ b/flow/environment.py
@@ -363,7 +363,7 @@ def get_environment(test=False, import_configured=True):
     """Attempt to detect the present environment.
 
     This function iterates through all defined :py:class:`~.ComputeEnvironment`
-    classes in reversed order of definition and and returns the first
+    classes in reversed order of definition and returns the first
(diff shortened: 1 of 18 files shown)


flow/project.py
```diff
@@ -84,12 +84,12 @@
 {template_dir}
 
 All template variables can be placed within a template using the standard jinja2
-syntax, e.g., the project root directory can be written like this: {{ project._rd }}.
+syntax, e.g., the project root directory can be written like this: {{{{ project._rd }}}}.
```
bdice: This may appear to be subtle -- we want to render two literal braces, like `{{ project._rd }}`. However, in strings designed for use with `str.format()`, it is necessary to include two literal braces for one brace to appear in the result.

Example:
```python
>>> "{{ project._rd }}, stuff, hello {world}".format(world='earth')
'{ project._rd }, stuff, hello earth'
>>> "{{{{ project._rd }}}}, stuff, hello {world}".format(world='earth')
'{{ project._rd }}, stuff, hello earth'
```

flow/project.py
```diff
@@ -1975,10 +1972,10 @@ def submit_operations(self, operations, _id=None, env=None, parallel=False, flag
             Show information about available template variables and filters and exit.
         :type show_template_help:
             bool
-        :param kwargs:
+        :param \*\*kwargs:
```
bdice: The current formatting implies that we have a keyword argument literally named `kwargs`. However, it is difficult to document `*args` and `**kwargs` so that they appear correctly in the rendered documentation. We must use a raw string (`r"""` above) and escape the asterisks so that it will render as `**kwargs` in the docs as desired.

flow/project.py
```diff
@@ -314,7 +314,7 @@ class FlowCondition(object):
     :param callback:
         A function with one positional argument (the job)
     :type callback:
-        :py:class:`~signac.contrib.job.Job`
+        function(:py:class:`~signac.contrib.job.Job`)
```
vyasr: Our standard is to refer to these as callables (to allow for e.g. functors as well).

flow/project.py
```diff
@@ -342,20 +342,20 @@ class FlowOperation(object):
...
+    which expects a job id as its first argument, we could construct the following
     two equivalent operations:
 
     .. code-block:: python
 
         op = FlowOperation('hello', cmd='hello {job._id}')
         op = FlowOperation('hello', cmd=lambda 'hello {}'.format(job._id))
 
-    Here is another example for possible str-substitutions:
+    Here is another example of possible string substitutions:
```
vyasr: "...of *a* possible..."

flow/project.py
```diff
@@ -701,9 +701,9 @@ def foo_label(job):
         The ``foo-label-text`` label will now show up in the status view for each job,
         where the ``foo`` key evaluates true.
 
-        If instead of a ``str``, the label functions returns any other type, the label
-        name will be the name of the function if and only if the return value evaluates
-        to ``True``, for example:
+        If the label functions returns any type other than ``str``, the label
+        name will be the name of the function if and only if the return value
+        evaluates to ``True``, for example:
```
vyasr: While we're here, could you also add a proper parameter docstring?

flow/environment.py
```diff
@@ -40,9 +40,9 @@ def setup(py_modules, **attrs):
     """Setup function for environment modules.
 
     Use this function in place of setuptools.setup to not only install
-    a environments module, but also register it with the global signac
-    configuration. Once registered, is automatically imported when the
-    get_environment() function is called.
+    a environment's module, but also register it with the global signac
+    configuration. Once registered, the environment is automatically
+    imported when the get_environment() function is called.
```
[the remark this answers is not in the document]
bdice: I added crossref links, but I can't test this properly since neither `setup` nor `ComputeEnvironment` are documented.

flow/project.py
bdice: I added a parameter docstring but I'm not sure if it's accurate. I don't fully understand _how_ this decorator resolves between name/function, but that's how I documented it.

flow/project.py
```diff
@@ -1687,7 +1684,7 @@ def run(self, jobs=None, names=None, pretend=False, np=None, timeout=None, num=N
 
         This function will run in an infinite loop until all pending operations
         have been executed or the total number of passes per operation or the total
-        number of exeutions have been reached.
+        number of executions have been reached.
```
cbkerr: I have trouble parsing this parallel structure.
I propose:
This function will run in an infinite loop until executing all
pending operations, unless it reaches the maximum number of
passes per operation or the maximum number of executions.

bdice: Good rewrite. I committed that.

flow/util/template_filters.py
```diff
@@ -11,12 +11,12 @@
 
 
 def identical(iterable):
-    """Check that all elements of an iterator are identical"""
+    """Check that all elements of an iterator are identical."""
```
vyasr: Technically this should be an iterable.
```suggestion
    """Check that all elements of an iterable are identical."""
```

flow/project.py
```diff
@@ -342,20 +342,20 @@ class FlowOperation(object):
...
 
     For example, if we wanted to define a command for a program called 'hello',
-    which expects a job id as its first argument, we could contruct the following
+    which expects a job id as its first argument, we could construct the following
     two equivalent operations:
 
     .. code-block:: python
 
         op = FlowOperation('hello', cmd='hello {job._id}')
         op = FlowOperation('hello', cmd=lambda 'hello {}'.format(job._id))
```
vyasr: This lambda needs an argument.
```suggestion
         op = FlowOperation('hello', cmd=lambda job: 'hello {}'.format(job._id))
```

flow/project.py
```diff
@@ -1685,9 +1686,9 @@ def run(self, jobs=None, names=None, pretend=False, np=None, timeout=None, num=N
             num_passes=1, progress=False, order=None):
         """Execute all pending operations for the given selection.
 
-        This function will run in an infinite loop until all pending operations
-        have been executed or the total number of passes per operation or the total
-        number of exeutions have been reached.
+        This function will run in an infinite loop until executing all pending
+        operations, unless it reaches the maximum number of passes per
+        operation or the maximum number of executions.
```
```suggestion
        This function will run in an infinite loop until all pending operations
        are executed, unless it reaches the maximum number of passes per
        operation or the maximum number of executions.
```

flow/project.py
```diff
@@ -342,20 +342,20 @@ class FlowOperation(object):
 
     Any FlowOperation is associated with a specific command, which should be
     a function of :py:class:`~signac.contrib.job.Job`. The command (cmd) can
-    be stated as function, either by using str-substitution based on a job's
+    be stated as function, either by using string substitution based on a job's
     attributes, or by providing a unary callable, which expects an instance
     of job as its first and only positional argument.
```
vyasr: I just realized how confusing this wording is with regard to the command string option. I'd suggest
```suggestion
    Any FlowOperation is associated with a specific command which should be
    a function of :py:class:`~signac.contrib.job.Job`. The command (cmd) may
    either be a unary callable that expects an instance of
    :class:`~signac.contrib.job.Job` as its only positional argument and returns
    a string containing valid shell commands, or the string of commands itself.
    In either case, the resulting string may contain any attributes of the job placed
    in curly braces, which will then be substituted by Python string formatting.
```

vyasr: I don't think this is actually recognizing the commit as changing multiple lines, unfortunately. It's highlighting all the relevant lines, but I'm not sure it will actually overwrite anything other than the last one.

flow/project.py
vyasr: Same concern as previously with respect to multiple line suggestion.

flow/project.py
bdice: You're right -- the suggestions feature isn't compatible with multi-line edits, in my understanding. I'll apply this manually.

bdice:
@cbkerr approved this PR verbally. (@cbkerr For next time, there's a review form and a button to approve -- let me know if you want me to show you, or check out [this page in the GitHub docs](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-request-reviews#about-pull-request-reviews)!)

PR #207, opened 2019-12-23
Revise deprecations

bdice:
## Description
Revises deprecations discussed with @vyasr via Slack. We want to move deprecations targeting 1.0 to future 0.x versions, because our roadmap to 1.0 is not yet clear. Of course, if 1.0 comes early, we're still technically correct to say that it will be removed in a 0.x version (because 1.0 > 0.x).

I also deprecated the function `make_bundles` and replaced it with a private method.

## Motivation and Context
Clarifying software roadmap.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature? Not sure.

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -148,10 +148,10 @@ class _condition(object):
                    'not_'.encode() + condition.__code__.co_code)
 
 
-def make_bundles(operations, size=None):
+def _make_bundles(operations, size=None):
     """Utility function for the generation of bundles.
 
-    This function splits an iterable of operations into  equally
+    This function splits an iterable of operations into equally
     sized bundles and a possibly smaller final bundle.
     """
     n = None if size == 0 else size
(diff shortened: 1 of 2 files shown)


flow/project.py
```diff
@@ -128,7 +128,7 @@ def false(cls, key):
 
     @classmethod
     @deprecated(
-        deprecated_in="0.9", removed_in="1.0",
+        deprecated_in="0.9", removed_in="0.11",
```
zhou-pj: Should we add `, current_version=__version__` for consistency?

zhou-pj: Just curious, is the reason for removing in 0.11 here vs 0.10 as most others being that we usually make deprecated and removed version to be 2 versions apart?

bdice: Good find! I hadn't noticed that `current_version` was missing. I fixed it.

bdice: Yes, that is why. We have a deprecation cycle of 2 minor versions: https://docs.signac.io/projects/flow/en/latest/support.html#deprecation-policy

PR #209, opened 2019-12-23
Issue#38 add ignore_condition for submit and run

zhou-pj:
## Description
This PR address issue #38 to allow condition overwrite feature and is another pre-requirement for issue #114 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


zhou-pj:
a prototype for adding ignore_condition for submit and run for preliminary review.

Currently I only added it for submit() as a example for review. My understanding is that this PR should allow the following 3 user case that I can think of now:
1. ignore_pre_condition=True for submit with group to allow downstring operation to be submitted
2. ignore_post_condition=True to address the original issue #38 to allow force rerun, etc
3. ignore_condition=True within run in compute node for cases where some condition check is not feasible

Would like some feedback on whether the changes can cover all needs for the group PR.

I currently add`ignore_condition` ,`ignore_pre_conditions`, `ignore_post_conditions` as argument for submit but keep the `only_eligible` argument for downstream functions to minimize changes in current function arguments.

b-butler:
I am not sure if we need to have the level of granularity provided. I think ignoring all conditions is all that is needed even though it is in #38. However, if we do provide those, I suggest that we make `ignore_conditions` a container of options like `'pre'`, `'post'`, `'all'` and that any conditions kinds in `ignore_conditions` are ignored. Having three arguments seems a bit verbose for the use case.

csadorf:
> I am not sure if we need to have the level of granularity provided. I think ignoring all conditions is all that is needed even though it is in #38. However, if we do provide those, I suggest that we make `ignore_conditions` a container of options like `'pre'`, `'post'`, `'all'` (or just the string) and that any conditions kinds in `ignore_conditions` are ignored. Having three arguments seems a bit verbose for the use case.

One solution to this could be a bit mask, that could be implemented manually or via a [`enum.Flag`](https://docs.python.org/3/library/enum.html#enum.Flag). Alternatively, one can also just use a tuple of strings, or space-concatenated strings.

zhou-pj:
@csadorf I tried to simply the input with enum.Flag but I am not sure If what i did is in line what you are suggesting. Would you mind take a look and see it that make sense to you. Also I only notice afterwards that enum.flag is new in python3.6 so 3.5 test failed. Would that mean we can't use it, or if we have any near plan to drop 3.5?

bdice:
@zhou-pj Python 3.5 end of life is 2020-09-13. I don't think we'll drop support for Python 3.5 before that date at the earliest.

zhou-pj:
> @zhou-pj Python 3.5 end of life is 2020-09-13. I don't think we'll drop support for Python 3.5 before that date at the earliest.

Thanks Bradley, I changed the Enum to support python 3.5.

flow/project.py
```diff
@@ -405,11 +414,13 @@ def __init__(self, cmd, pre=None, post=None, directives=None):
     def __str__(self):
         return "{type}(cmd='{cmd}')".format(type=type(self).__name__, cmd=self._cmd)
 
-    def eligible(self, job):
+    def eligible(self, job, ignore_condition=Ignore_condition.NONE):
```
```suggestion
    def eligible(self, job, ignore_conditions=Ignore_condition.NONE):
```
Since this is applied to *all* conditions, right?

flow/project.py
```diff
@@ -95,6 +97,13 @@
 {filters}"""
 
 
+class Ignore_condition(IntEnum):
```
```suggestion
class IgnoreConditions(IntEnum):
```
Classes should **always** be camel case.

flow/project.py
zhou-pj: I agree, thanks for pointing this out.

flow/project.py
```diff
@@ -405,11 +414,13 @@ def __init__(self, cmd, pre=None, post=None, directives=None):
     def __str__(self):
         return "{type}(cmd='{cmd}')".format(type=type(self).__name__, cmd=self._cmd)
 
-    def eligible(self, job):
+    def eligible(self, job, ignore_conditions=IgnoreConditions.NONE):
         "Eligible, when all pre-conditions are true and at least one post-condition is false."
-        pre = all(cond(job) for cond in self._prereqs)
+        pre = (ignore_conditions & IgnoreConditions.PRE) \
```
b-butler: For `pre` we don't do the length check like we do with post. I think consistency would be nice. In addition, since we are dealing with booleans, we can just use multiple conditions like `bool = cond1 or cond2 or cond3` since `or` short circuits.

zhou-pj:
Completed for full review. It should fully address #38 for command line condition ignore as well as the `FlowProject.run needs an extra option to tell it not to check conditions, but to just run everything it was provided. This feature will enable the primary use-case of "exec mode".` comment in #114

flow/project.py
```diff
@@ -418,11 +435,20 @@ def __init__(self, cmd, pre=None, post=None, directives=None):
     def __str__(self):
         return "{type}(cmd='{cmd}')".format(type=type(self).__name__, cmd=self._cmd)
 
-    def eligible(self, job):
-        "Eligible, when all pre-conditions are true and at least one post-condition is false."
-        pre = all(cond(job) for cond in self._prereqs)
+    def eligible(self, job, ignore_conditions='none'):
+        """Eligible, when all pre-conditions are true and at least one post-condition is false,
+        or corresponding conditions are ignored """
```
csadorf: Please follow conventions about doc-string formatting outlined in the active [PEP 257](https://www.python.org/dev/peps/pep-0257).

@mikemhenry I think it will be crucial streamline our development workflow to automatically lint for these formatting issues as soon as possible.

flow/project.py
```diff
@@ -1772,6 +1798,12 @@ def run(self, jobs=None, names=None, pretend=False, np=None, timeout=None, num=N
 
             The default value is `none`, which is equivalent to `by-job` in the current
             implementation.
+        :param ignore_conditions:
+            Specify if pre or post conditions check is to be ignored for eligibility check
+                * 'none' (default: all conditions are checked)
+                * 'pre' (ignore pre condition check)
+                * 'post' (ignore post condition check)
+                * 'all' (ignore all condition check)
```
csadorf: I'm not sure that a dual approach like this makes sense. I understand that we need to accept strings on the command line, but on the Python code level we could actually use the enum, right?

zhou-pj: I had the way you said implemented in a earlier commit where string is only used for terminal and Enum class is required for python `run` and `submit`. My main thoughts for moving to all string for user facing argument is this looks simpler and more unified for `run` and `submit` across using in python or terminal, and I thought it might be a bit too much requirement for user to be aware and use enum. But I want to get more thoughts on this and I am happy to change to whichever people prefers. @vyasr @b-butler

vyasr: I agree with Simon, everything in Python should use the enum. As long as the function docstring indicates that the enum should be used (with the appropriate Sphinx link to `~.IgnoreConditions`) it's fine to enforce that requirement.  Setting up an argparse Action like I showed you for the CLI is fine, but everything passed around in the code should be the enum object.

flow/project.py
```diff
@@ -94,6 +96,21 @@
...
+    ALL = PRE | POST
+    NONE = ~ ALL
+
+
+_IGNORE_CONDITIONS = {
+    'pre': IgnoreConditions.PRE,
+    'post': IgnoreConditions.POST,
+    'all': IgnoreConditions.ALL,
+    'none': IgnoreConditions.NONE
+}
```
vyasr: Following Simon's outstanding comment, we should be able to get rid of this entirely.

flow/project.py
```diff
@@ -418,11 +435,21 @@ def __init__(self, cmd, pre=None, post=None, directives=None):
...
         return "{type}(cmd='{cmd}')".format(type=type(self).__name__, cmd=self._cmd)
 
-    def eligible(self, job):
-        "Eligible, when all pre-conditions are true and at least one post-condition is false."
-        pre = all(cond(job) for cond in self._prereqs)
+    def eligible(self, job, ignore_conditions='none'):
+        """Eligible, when all pre-conditions are true and at least one post-condition is false,
+        or corresponding conditions are ignored
+        """
+        if ignore_conditions in _IGNORE_CONDITIONS.keys():
```
vyasr: Once you change the API, the correct test would be `if type(ignore_conditions) == IgnoreConditions`. @b-butler do you think this test is necessary, or would you prefer to allow users to pass in integers directly and ducktype them? My initial thought was that ducktyping is fine, but in Python maybe it's asking too much of users to not misuse enums, which are far less common in Python than in lower-level languages.

flow/project.py
```diff
@@ -418,11 +435,21 @@ def __init__(self, cmd, pre=None, post=None, directives=None):
...
+            ignore_conditions = _IGNORE_CONDITIONS[ignore_conditions]
+        else:
+            raise ValueError(
+                "The ignore_conditions argument of FlowProject.run() must be a string in "
+                "'none', 'pre', 'post' or 'all'")
+        pre = (not len(self._prereqs)) or (ignore_conditions & IgnoreConditions.PRE) \
```
vyasr: Is the `len` check here as an optimization? If so, I'd include a comment to that effect since you can remove that check and the conditions are still equivalent (because `all([])` evaluates to `True`).

flow/project.py
```diff
@@ -2932,12 +2996,24 @@ class MyProject(FlowProject):
...
         parser_script = subparsers.add_parser(
             'script',
             parents=[base_parser],
         )
+        parser_script.add_argument(
+            '--ignore-conditions',
+            type=str,
+            choices=['none', 'pre', 'post', 'all'],
+            default='none',
+            help="Specify conditions to ignore for eligibility check.")
```
vyasr: Since the Python code is being changed to use the enum instead of strings, you'll need to switch these parsers to translate the strings to the enums (ideally using a custom [`argparse.Action`](https://docs.python.org/3/library/argparse.html#argparse.Action)).

flow/project.py
```diff
@@ -94,6 +96,29 @@
 {filters}"""
 
 
+class IgnoreConditions(IntEnum):
+    """IntEnum class for IgnoreConditions settings when checking job eligibility, setting includes:
+    * IgnonreConditions.PRE: ignore pre conditions
+    * IgnonreConditions.POST: ignore post conditions
+    * IgnonreConditions.ALL: ignore all conditions
+    * IgnonreConditions.NONE: check all conditions
```
vyasr: Typos on all of these.

flow/project.py
```diff
@@ -1772,6 +1815,11 @@ def run(self, jobs=None, names=None, pretend=False, np=None, timeout=None, num=N
 
             The default value is `none`, which is equivalent to `by-job` in the current
             implementation.
+        :param ignore_conditions:
+            Specify if pre and/or post conditions check is to be ignored for eligibility check.
+            The default is `IgnoreConditions.NONE`.
+        :type ignore_conditions:
+            py:class:`~.IgnoreConditions`
```
```suggestion
            :py:class:`~.IgnoreConditions`
```

PR #222, opened 2020-01-22
Add Bridges AI Template

bdice:
## Description
I am running some jobs on the Bridges GPU-AI partition, which doesn't have a signac-flow template. I updated the `bridges.sh` template and the XSEDE environments to support this part of the cluster.

Documentation about the nodes: https://www.psc.edu/bridges/user-guide/running-jobs#ai

There are 40 CPU cores and 8 V100 GPUs in each node. I am ignoring the DGX-2 since there is only one of them (there are several V100 nodes) and the DGX-2 is probably only needed for jobs that require large GPU memory (32 GB instead of 16 GB).

## Motivation and Context
Need to run deep learning jobs on Bridges-AI.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


bdice:
I changed the reviewers since @vyasr has some experience with this cluster and @tommy-waltmann will be using this feature with me.

flow/templates/bridges.sh
```diff
@@ -29,14 +37,18 @@
 #SBATCH -N {{ nn|default(1, true)|check_utilization(gpu_tasks, 1, threshold, 'GPU') }}
 #SBATCH --ntasks-per-node=16
 #SBATCH --gres=gpu:p100:{{ 2 if gpu_tasks > 1 else 1 }}
+{% elif partition == 'GPU-AI' %}
+#SBATCH -N {{ nn|default(1, true)|check_utilization(gpu_tasks, 8, threshold, 'GPU') }}
+#SBATCH --ntasks-per-node=40
```
bdice: I don't know if this is the right choice. This requests all the CPU cores on a node, which matches the behavior on line 31 (for the GPU and GPU-small partitions). However, these nodes have 8 GPUs and it seems wasteful to request the whole node's CPU if the job can't use all 8 GPUs as well. I suppose this is what the "check utilization" is for, but I can imagine wanting to use these nodes as if they're part of a shared queue (which is allowed). Should we use the number of `{{ cpu_tasks }}` instead?

vyasr: @joaander I'd appreciate some feedback on this. I would definitely want to use the GPU-AI partition for shared jobs, but I'm not sure what the right choice is for how many CPUs to request based on the number of GPUs, or if there's even any good general way to make this decision implicitly.

bdice: The number of CPU cores (40) does evenly divide by the number of GPUS (8).
Some of our options:
- `{{ gpu_tasks }}` (one CPU core per GPU)
- `{{ gpu_tasks * 5 }}` (ensure the CPU cores are proportional to the number of GPUs)
- `{{ cpu_tasks }}` (leave it to the user to request the right amount of CPUs, which I think defaults to one core per task?)

joaander: The user must be able to choose the number of CPU cores requested. HOOMD multi-GPU jobs may only need 1 CPU but several GPUs while traditional MPI jobs need 1 CPU per GPU. AI training jobs may need several GPUs and several CPU cores per GPU to feed them with data. PSC does not explicitly document this for GPU-AI, but the GPU-shared partition charges SUs based on the number of CPU cores requested. It should be an error to request more than 5 CPUs per GPU.

bdice: Thanks @joaander! I'll change to `{{ cpu_tasks }}`, and error if it's more than 5 CPUs/GPU.

PR #233, opened 2020-02-01
Replace mode argument with partition

cbkerr:
## Description

Allows the user to specify the partition with the --partition flag like for other Slurm clusters.
## Motivation and Context

There was no way to submit to a GPU cluster on Great Lakes before this.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/environments/umich.py
+++ b/flow/environments/umich.py
@@ -19,7 +19,7 @@ class GreatLakesEnvironment(DefaultSlurmEnvironment):
         super(GreatLakesEnvironment, cls).add_args(parser)
         parser.add_argument(
             '--partition',
-            choices=('standard', 'gpu'),
+            choices=('standard', 'gpu', 'largemem'),
             default='standard',
             help="Specify whether to submit to the CPU or the GPU queue. "
                  "(default=standard)")
--- a/flow/environments/umich.py
+++ b/flow/environments/umich.py
@@ -21,7 +21,7 @@ class GreatLakesEnvironment(DefaultSlurmEnvironment):
             '--partition',
             choices=('standard', 'gpu', 'largemem'),
             default='standard',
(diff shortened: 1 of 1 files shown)


cbkerr:
Should we also allow the options for other Great Lakes partitions listed here?
https://arc-ts.umich.edu/greatlakes/rates/

vyasr:
I would say that we don't need to support viz or debug, both of which are primarily for interactive purposes. Adding largemem is probably worthwhile.

bdice:
What should happen if we request more GPUs than are available in one node? e.g. I would expect `nranks=24`, `ngpu=24` to give 12 nodes with 2 tasks per node (each node has 2 GPUs). Instead, it gives 24 tasks per node. This is important for MPI + multi-GPU communication.

```bash
#!/bin/bash
#SBATCH --job-name="aa16a1af-b00/7f3e901b/op_0/0000/8c81aafce7d9ae767f4f20809e2fdaf1"
#SBATCH --partition=gpu
#SBATCH --nodes=12
#SBATCH --ntasks-per-node=24
#SBATCH --gres=gpu:24

set -e
set -u

cd /tmp/tmpui7veles

# op_0(7f3e901b4266f28348b38721c099d612)
mpiexec -n 24 /home/bdice/miniconda3/envs/dice/bin/python ./interactive_template_test.py exec op_0 7f3e901b4266f28348b38721c099d612
```

I believe the Bridges-AI template may be a good example of how to handle this. When requesting more GPUs than are available on a GPU node, I think we want to split the CPU tasks across nodes correspondingly.

vyasr:
I agree with @bdice that this is an issue that needs to be addressed in some way (perhaps in the proposed manner), but I think that is a separate issue and should be addressed independently. I'll open an issue for that, but this pull request can remain as is.

PR #243, opened 2020-02-06
Deprecation Warning for --cmd option in script

### LINKED ISSUE #218: Inconsistent eligibility check for _main_script
### Description

As pointed out by @vyasr, there is inconsistent eligibility check in def _main_script() depending on whether cmd option is provided or not. Is there specific reason for this or should we make them consistent?

vishav1771:
## Description

Added eligibility check for operations generated when `cmd` is provided in `main_script`

## Motivation and Context

Fixes #218 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:


csadorf:
@vishav1771 For bug-fixes it is usually a good idea to implement the unit test that catches the bug *first*!

vishav1771:
@csadorf I tried that but don't know how to make an operation (generated by cmd ineligible). Can you please help me with that.

csadorf:
> @csadorf I tried that but don't know how to make an operation (generated by cmd) ineligible. Can you please help me with that.

You can provide a list of labels that must be present for the cmd-operation to be eligible. So should be easy to just provide a label that is not present. Have a look at the command-line help.

vyasr:
I think perhaps this is a nonissue cause by some misunderstanding from @zhou-pj and me. It looks like `_generate_operations` does check the `requires` argument, it just doesn't use any of the `eligible` or `eligible_for_submission` tests that everything else does. However, given that such commands are provided directly from the command line, the only valid conditions are the ones that are provided from the `--requires` command line argument, right? So perhaps there's nothing to change. @csadorf @zhou-pj would you agree, or am I missing something.

csadorf:
Yeah, it seems that this might be invalid. I thought #218 was saying that the script generation command does not properly check eligibility as opposed to `submit --pretend`.

zhou-pj:
I agree we overlooked the `requires` argument for `cmd` and it might not be a bug but a prior design decision. I have two questions/thoughts though: 1. it seems the `requires` option was designed to be only used if it is a `cmd`, but I am not sure why this is the case? 2. If I understand correctly the `requires` logic gives a different level of flexibility by checking job eligibility on label level (but without the (pre is true and post is false -> eligible) logic and only exist for `script`, this essentially means `script` now have two set of ways for user to define eligible check(by `requires` and by the `ignore_conditions` recently added), so it might be good to revisit this and agree on a unify way to avoid confusion. And since `requires` option is unique only for `script` here, I think it would be nice to have the default be the same as `submit --pretend` when `requires` is not provided, and `requires` can have higher priority since it have more flexibility.

csadorf:
@zhou-pj I'm open to unifying the interface, but I'm not sure that we should extend the use of `--requires` in any sense, because this kind of usage is inherently less reproducible and  should therefore be discouraged.

If you would like to create a proposal as how to unify the interface, maybe you can make a separate issue for that? Otherwise, I'd vote to close this issue for now.

vyasr:
What is the rationale for even supporting this use-case (specifying an _arbitrary_ cmd on the command line)? I would prefer to deprecate it entirely and require users to create a `flow.cmd` operation in order to allow this kind of behavior. This usage feels contrary to the design of flow. The use of the `requires` argument is especially nonintuitive because in general there is no requirement that conditions be label functions. That makes it confusing that the requirements here must be label functions.

csadorf:
> What is the rationale for even supporting this use-case (specifying an _arbitrary_ cmd on the command line)?

This feature was originally introduced by @PaulDodd , I personally have never used it. I can see some use case where users want to quickly apply some command to a specific set of jobs. Something like this:

```bash
python project.py run -f a 0 --cmd "rm data.json" 
```
> I would prefer to deprecate it entirely and require users to create a flow.cmd operation in order to allow this kind of behavior.

This feature does not appear to be inherently broken. I would request that maintainers show a bit more restrain when it comes to requesting the deprecation of long-supported features.

bdice:
I understand there is disagreement here among maintainers, and respect your opinion @csadorf. However, I agree with @vyasr that this should be deprecated. A reduction of scope is worth considering here, especially because its interface is dissimilar from other API endpoints in flow (e.g. its use of `requires`). This behavior is (to the best of my knowledge) not documented, not tested, and not utilized by our end-users. I don't think that _being non-broken_ is a good enough reason to support `run --cmd` in a stable release of signac-flow, especially when alternative options like `signac shell` and `signac find | xargs` exist.

vyasr:
While I agree in general that we need to be careful about deprecating features, I do not think that it is beneficial to the long-term health of the project to simply retain _all_ features until they break or introduce a maintenance burden. Other considerations, such as whether a feature still falls within the scope of the project and whether its existence would cause confusion for newer users, are also important to take into account. Despite having never reached a 1.0 release, I recognize that signac-flow has been stable enough for a long time that users have some expectation of stability. However, the project has evolved substantially from its original iterations, and as we've honed in more closely on our core concepts and scope I think it's important to reconsider scope. 

I also think it is also realistic to acknowledge that its user base for the first few years of life was very limited, so features that were introduced at that point and have since been essentially forgotten by the current user base. Otherwise we run the risk of supporting highly dissimilar APIs for backwards compatibility with a potentially non-existent set of end-users. I'm also not exactly sure where you feel comfortable drawing this line; for instance, what differentiates this situation from the recent deprecations in the core package? Many of those were also functions that are not broken, but we deprecated them because there were better options and we gave plenty of lead time to users to make the necessary switches.

Also agree with Bradley's comments, his came up while I was still typing :)

csadorf:
I really stirred up a controversy here. :laughing: 

I can see the motivation for deprecation and won't stand in its way. My main argument against it was that it does not appear inherently broken or problematic to me. As I said, I don't think I have ever used this feature before, so I don't think I have a subjective bias here.

vyasr:
Landmark decisions in the signac framework 😂 

If everyone is OK with that, I think that should be the new focus of this PR. Let's add deprecation warnings to the `--cmd` and `--requires` arguments (the direct cmd group) of `_main_script` and plan to remove them in a couple of releases.

@zhou-pj I just read fully through your previous post. In case the issue is not clear, the `--requires` is not related to operations decorated with `@flow.cmd` in Python. It is specifically to allow users to specify arbitrary commands on the command line (as shown in Simon's example). My contention, which I think Bradley and Simon are in agreement with, is that this behavior is out of scope for flow and not something we need to support.

zhou-pj:
@vyasr thanks for your clarification! And I agree with deprecate --requires now that it is all clear to me. My previous concern was mostly that the coexistence of  --requires and eligible check with ignore options may be confusing to user, and deprecating --requires would also clear this concern.

vyasr:
@vishav in that case, it looks like your task on this PR has changed a bit. Sorry for the runaround. Can you deprecate the functionality we discussed? Normally in flow we use the `deprecation` package to deprecate functions, but since in this case it's a command line option you'll just need to raise DeprecationWarnings yourself. The place to put this would be in `_main_script` if `--cmd` or `--requires` are specified.

csadorf:
> Normally in flow we use the `deprecation` package to deprecate functions, but since in this case it's a command line option you'll just need to raise DeprecationWarnings yourself. The place to put this would be in `_main_script` if `--cmd` or `--requires` are specified.

Raising a `DeprecationWarning` is not the right approach either in this case, because we don't want the function to fail. Instead you want to update the command's description to explicitly label it as deprecated and issue a warning to STDERR whenever it is used. So completely "manual".

vishav1771:
Sorry for the delay. I have done the changes. I have found a minor error while using `--require`. I have done that change also. Should I add these changes to `changelog.py`?

vishav1771:
@vyasr Sorry you tagged someone else :)

vyasr:
Whoops sorry I trusted Github's autocomplete a bit too much on this perhaps :)

flow/project.py
```diff
@@ -2816,6 +2817,8 @@ def _main_script(self, args):
         # Gather all pending operations or generate them based on a direct command...
         with self._potentially_buffered():
             if args.cmd:
+                print("DeprecatedWarning: --cmd option for script is deprecated as of 0.9 "
```
```suggestion
                print("DeprecationWarning: --cmd option for script is deprecated as of 0.9 "
```

vyasr:
@zhou-pj once you get a chance to review feel free to merge.

vishav:
You have added the wrong person. My GitHub alias is vishav. Please remove
me from your list.

zhou-pj:
looks good to me pending Vyas's comments. And regarding @vishav1771 's question, is it the general rule that we put the deprecation info in changelog when they are actually removed, so not in this PR but a later v0.11 clean up PR? @vyasr

vyasr:
We should update the changelog now, good catch. @vishav1771 can you update the changelog? Feel free to copy one of the previous entries, and make sure to include the relevant PR/issue numbers (#243 and #218).

changelog.txt
```diff
@@ -19,6 +19,11 @@ Added
 - Add IgnoreConditions option for submit(), run() and script() (#38).
 - Add pytest support for Testing Framework (#227, #232).
 
+Changed
++++++++
+
+- The commad line option ``--cmd`` for ``script`` has been deprecated and will trigger a ``DeprecationWarning`` upon use until removed (#243, #218).
```
vyasr:
```suggestion
- The command line option ``--cmd`` for ``script`` has been deprecated and will trigger a ``DeprecationWarning`` upon use until removed (#243, #218).
```

Issue #244, opened 2020-02-07
Bloated testing output due to deprecated warnings


b-butler:
### Description
Reliance on deprecated functions, methods, etc. needs to be removed from out testing. 

### To reproduce

Just run the flow testing suite.

### Fix

Go through the testing output and find all references to deprecated functions and remove them.

csadorf:
Thank you for creating this issue. I specifically asked to split fixing this from the test refactoring PR, because I would like that change to be 100% refactoring, no other changes.

I think we should absolutely refactor the package code to stop relying on deprecated functions internally, which should already reduce the noise, however I'd like to be a bit careful with updating all the tests, and rather mark them as explicitly testing deprecated API and suppress warnings instead of changing them all. Unit tests that test deprecated API and thus trigger deprecation warnings are vital to actually ensure that deprecated behavior is still supported and thus actually maintaining backwards compatibility.

Deprecated API that should be removed should of course lead to failed tests.

b-butler:
Many tests that aren't designed explicitly to test deprecated API are uses deprecated functions. I do see your point with maintaining backwards compatible with deprecated code until the correct version. I do think however that a few contained test explicitly for deprecated code with the main code base relying on non-deprecated code would be preferable. I do believe for the most part the package does not call deprecated functions.

kidrahahjo:
Can I work on removing the deprecated warnings? While working for pytest implementation, this issue occurred for me too, and it became frustrating scrolling to the top to see what errors I'm getting.

csadorf:
@kidrahahjo Yes, of course! Let's just make sure that #232 is merged first.

bdice:
@kidrahahjo This should be unblocked. Thanks for offering to help, and let us know if you have questions! You'll want to start from the current `glotzerlab:master` branch (not the branch you created for the previous PR).

kidrahahjo:
> @kidrahahjo This should be unblocked. Thanks for offering to help, and let us know if you have questions! You'll want to start from the current `glotzerlab:master` branch (not the branch you created for the previous PR).

Okay, thanks. 
I have started working on this. Will raise a PR soon.

kidrahahjo:
@bdice @csadorf 
To fix the deprecation warning of get_statepoint, I'd need to commit to the signac repository too. Should I do that?

csadorf:
@kidrahahjo Yes, please create a separate PR for that.

bdice:
[...]

PR #247, opened 2020-02-12
Use tqdm as a requirement, remove vendored tqdm module.

bdice:
## Description
This PR replaces the vendored tqdm module with a tqdm requirement.

## Motivation and Context
The vendored tqdm module was outdated and it is easier to maintain a requirement. The vendored module was causing DeprecationWarnings: https://github.com/glotzerlab/signac-flow/issues/244#issuecomment-585280247

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


requirements.txt
```diff
@@ -2,3 +2,4 @@ signac>=1.0.0
 jinja2>=2.8
 cloudpickle
 deprecation>=2
+tqdm>=4.10.0
```
csadorf:
```suggestion
tqdm~=4.10.0
```
[The tqdm package follows semantic versioning](https://tqdm.github.io/contributing/#semantic-versioning). Should we therefore maybe use the [compatible release](https://www.python.org/dev/peps/pep-0440/#compatible-release) specifier instead?

setup.py
```diff
@@ -14,6 +14,8 @@
     'cloudpickle',
     # Deprecation management
     'deprecation>=2',
+    # Progress bars
+    'tqdm>=4.10.0',
```
csadorf:
```suggestion
    'tqdm~=4.10.0',
```
See comment for `requirements.txt` above.

requirements.txt
bdice: @csadorf Our use of tqdm is essentially the simplest use case. It looks like v5 is in progress and I don't think they're going to break `from tqdm import tqdm` according to the list of breaking changes. https://github.com/tqdm/tqdm/milestone/8

Usually I would say compatible versions are preferable, but I don't think we need to worry about that here. If you feel strongly about that, feel free to make the change here and in the signac repo (that PR was already merged).

csadorf: Fine, no need for changes.

csadorf:
@yuanzhou0827 I hope you don't mind, but I figured this PR was so trivial that I merged it without your review.

yuanzhou0827:
@csadorf Never mind! I jam sorry... I just saw the review request. Thank you!

PR #249, opened 2020-02-15
Add error when operation function passed to FlowProject.pre() and non-operation function passed to FlowProject.pre.after()

### LINKED ISSUE #248: Fail gracefully when operations are used as preconditions
### Feature description

Currently both using operations as the argument to `FlowProject.pre` and using non-operations as arguments to `FlowProject.pre.after` throw no errors, but simply silently function as no-ops. This behavior can confuse users who don't know why the conditions apparently are not obeyed. We should provide useful errors to users who try to do this.

### Proposed solution

Since operations are stored in the `FlowProject` in a list, we could check this list when preconditions are created (it may have to be done when they're collected) to ensure that the functions provided to the different types of decorators are appropriate.

ac-optimus:
Resolves #248 
## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


flow/project.py
```diff
@@ -557,6 +557,9 @@ def __init__(self, condition, tag=None):
                 super(pre, self).__init__(condition, tag)
 
             def __call__(self, func):
+                for operation_i in self._parent_class._OPERATION_FUNCTIONS:
```
vyasr: Since `_OPERATION_FUNCTIONS` variable is currently a list of tuples of the form (`name`, `func`), it's cleaner to loop over the tuples directly and check equality with the second component:

```suggestion
                for (name, func) in self._parent_class._OPERATION_FUNCTIONS:
```

and then check if self.condition is equal to func.

flow/project.py
```diff
@@ -557,6 +557,9 @@ def __init__(self, condition, tag=None):
...
+                    if self.condition in operation_i:
+                        raise ValueError("Passing operation function is not allowed")
```
```suggestion
                        raise ValueError("Operation functions cannot be used as preconditions.")
```

flow/project.py
```diff
@@ -569,6 +572,13 @@ def copy_from(cls, *other_funcs):
             @classmethod
             def after(cls, *other_funcs):
                 "True if and only if all post conditions of other operation-function(s) are met."
+                for condition_i in other_funcs:
+                    check_flag = 0
+                    for operation_i in cls._parent_class._OPERATION_FUNCTIONS:
```
vyasr: Same as above, I would unpack the tuple and then check equality directly.

flow/project.py
```diff
@@ -569,6 +572,13 @@ def copy_from(cls, *other_funcs):
...
+                        if condition_i in operation_i:
+                            check_flag = 1
+                    if check_flag != 1:
+                        raise ValueError("Operation function not passed.")
```
```suggestion
                        raise ValueError("The argument to pre.after must be an operation function.")
```

ac-optimus:
Thanks for the review @vyasr. I will look into the details the next time when I sit!

ac-optimus:
> Thanks for putting together this PR! Your tests are currently failing for a somewhat subtle reason. As you've seen, operations are registered to a given subclass of FlowProject, not to an instance. This feature allows workflows to inherit from one another. In order to enable this kind of inheritance, whenever we're looking at all the operations for a given class, we need to walk the entire inheritance hierarchy and collect all operations defined by the parents of the current class.
> 
> If you look at `FlowProject.__init__`, one of the last functions called is `self._register_operations`, which in turn calls `_collect_operations`. This method (which is a classmethod) is where we currently check all parents to find all valid operations. You will need to do the same thing in your checks to get the expected behavior.
> 
> Practically, what this means is that instead of looping over `self._parent_class._OPERATION_FUNCTIONS`, you will want to loop over `self._parent_class._collect_operations()`.
> 
> I hope that's clear! Feel free to ask if you have any questions about that. This is a pretty complicated setup.

hi, modified according to the suggested changes. I have a small doubt -- If I see `self._parent_class._collect_operations()` and `self._parent_class._OPERATION_FUNCTIONS` have same elements stored in it. Then what is the point of having them differently? And let say if I do not replace `self._parent_class._OPERATION_FUNCTIONS` with `self._parent_class._collect_operations` then what will be the difference? The only difference I can see is that the tests are now proceeding as expected, but I don't know why?

vyasr:
The two are only equivalent if your Project class doesn't inherit from anything. Try running this script, hopefully it illustrates the difference more clearly. 

```python
from flow import FlowProject

class ParentProject(FlowProject):
    pass

class ChildProject(ParentProject):
    pass

@ParentProject.operation
def op1(job):
    pass

@ChildProject.operation
def op2(job):
    pass

print(ParentProject._collect_operations())
print(ChildProject._collect_operations())
print(ParentProject._OPERATION_FUNCTIONS)
print(ChildProject._OPERATION_FUNCTIONS)
```

The underlying reason for the difference is that subclasses of `FlowProject` only store operations that were added to them using the `@operation` decorator, but we also need them to include operations added to their parents. The `_collect_operations` method finds all of those methods as well.

vyasr:
The two are only equivalent if your Project class doesn't inherit from anything. Try running this script, hopefully it illustrates the difference more clearly. 

```python
from flow import FlowProject

class ParentProject(FlowProject):
    pass

class ChildProject(ParentProject):
    pass

@ParentProject.operation
def op1(job):
    pass

@ChildProject.operation
def op2(job):
    pass

print(ParentProject._collect_operations())
print(ChildProject._collect_operations())
print(ParentProject._OPERATION_FUNCTIONS)
print(ChildProject._OPERATION_FUNCTIONS)
```

The underlying reason for the difference is that subclasses of `FlowProject` only store operations that were added to them using the `@operation` decorator, but we also need them to include operations added to their parents. The `_collect_operations` method finds all of those methods as well.

ac-optimus:
> The two are only equivalent if your Project class doesn't inherit from anything. Try running this script, hopefully it illustrates the difference more clearly.
> 
> ```python
> from flow import FlowProject
> 
> class ParentProject(FlowProject):
>     pass
> 
> class ChildProject(ParentProject):
>     pass
> 
> @ParentProject.operation
> def op1(job):
>     pass
> 
> @ChildProject.operation
> def op2(job):
>     pass
> 
> print(ParentProject._collect_operations())
> print(ChildProject._collect_operations())
> print(ParentProject._OPERATION_FUNCTIONS)
> print(ChildProject._OPERATION_FUNCTIONS)
> ```
> 
> The underlying reason for the difference is that subclasses of `FlowProject` only store operations that were added to them using the `@operation` decorator, but we also need them to include operations added to their parents. The `_collect_operations` method finds all of those methods as well.

Just ran the script. Now I get the difference, thanks for clearing the doubt.

flow/project.py
```diff
@@ -569,6 +572,13 @@ def copy_from(cls, *other_funcs):
             @classmethod
             def after(cls, *other_funcs):
                 "True if and only if all post conditions of other operation-function(s) are met."
+                for condition_i in other_funcs:
+                    check_flag = 0
+                    for (name, fun) in cls._parent_class._collect_operations():
```
mikemhenry: Same thing here as above, not sure about `fun` but a good argument could change my mind.

ac-optimus: @mikemhenry changed `fun` and `_func`!😅

flow/project.py
```diff
@@ -557,6 +557,9 @@ def __init__(self, condition, tag=None):
                 super(pre, self).__init__(condition, tag)
 
             def __call__(self, func):
+                for (name, fun) in self._parent_class._collect_operations():
```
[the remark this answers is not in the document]
bdice: @mikemhenry @ac-optimus The underscore at the beginning indicates a private/internal symbol. To re-use a symbol name, you want `func_` instead of `_func`. See PEP 8: https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles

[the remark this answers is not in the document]
ac-optimus: ya you are correct, I have changed it to func_ @bdice @mikemhenry!

flow/project.py
```diff
@@ -569,6 +572,13 @@ def copy_from(cls, *other_funcs):
             @classmethod
             def after(cls, *other_funcs):
                 "True if and only if all post conditions of other operation-function(s) are met."
+                for condition_i in other_funcs:
```
bdice: This set of nested for loops could be written in a much shorter form, and we should only need to call `_collect_operations` once at the beginning. Also the error message should be plural since `*other_funcs` can include multiple functions. Try this code (it may need slight modification).
```python
operation_functions = [operation[1] for operation in cls._parent_class._collect_operations()]
if not all(condition in operation_functions for condition in other_funcs):
    raise ValueError("The arguments to pre.after must be operation functions.")
```

flow/project.py
```diff
@@ -557,6 +557,9 @@ def __init__(self, condition, tag=None):
                 super(pre, self).__init__(condition, tag)
 
             def __call__(self, func):
+                for (name, func_) in self._parent_class._collect_operations():
```
bdice: Try this instead. It's slightly easier to understand, in my opinion.
```python
operation_functions = [operation[1] for operation in cls._parent_class._collect_operations()]
if self.condition in operation_functions:
    raise ValueError("Operation functions cannot be used as preconditions.")
```

vyasr:
@ac-optimus for future PRs, when you request review please just re-request the reviews of currently assigned reviewers rather than adding new ones. We will add additional reviewers if we think they are needed, but in general we try to load balance the reviewing process, so it's best if you don't ask for extra reviewers to look at things. Feel free to ping assigned reviewers if they're unresponsive, though.

ac-optimus:
> I suggested a couple changes to make the code easier to read. Please also implement a similar check in `__call__` for `class post`, to raise `"Operation functions cannot be used as postconditions."`

done that!

ac-optimus:
> @ac-optimus for future PRs, when you request review please just re-request the reviews of currently assigned reviewers rather than adding new ones. We will add additional reviewers if we think they are needed, but in general we try to load balance the reviewing process, so it's best if you don't ask for extra reviewers to look at things. Feel free to ping assigned reviewers if they're unresponsive, though.

@vyasr I am extremely sorry for disturbing the process. Will surely take care of this from the next time.

mikemhenry:
No worries @ac-optimus! I did a quick glance through our documentation and we don't explicitly mention how reviews should be re-requested and not to manually add new reviewers. I've created an issue in the docs repo as a reminder to ourselves to improve our documentation.

vyasr:
@ac-optimus definitely not a big deal! I just wanted to let you know now so that you know for future PRs. @mikemhenry I think your requests have been addressed, could you re-review when ready?

bdice:
@ac-optimus I see I was re-requested for review. I have some previous comments about replacing `for` loops with list comprehensions. I would like those to be addressed before I review again.

ac-optimus:
> @ac-optimus definitely not a big deal! I just wanted to let you know now so that you know for future PRs. @mikemhenry I think your requests have been addressed, could you re-review when ready?

sure @mikemhenry @vyasr !

ac-optimus:
> @ac-optimus I see I was re-requested for review 3 days ago. I have some previous comments about replacing `for` loops with list comprehensions. I would like those to be addressed before I review again.

@bdice I requested to review after doing the suggested changes. I did not choose to use list comprehension as the line was getting too long and flake8 indicated an error. I still do not understand why is it important to use list comprehension. Is there any benefit in using that as in either cases the space required and run time are the same? The current implementation (`for`) just adds two extra lines and makes the code more readable to my understanding, please correct if I am wrong.

bdice:
@ac-optimus List comprehensions are always faster and are considered to be a better style (more "Pythonic") in most cases. See below for some references. While my suggestions did not include line breaks for flake8, that's just because I was making suggestions via GitHub and not in a real code editor. I still support making those changes.

https://stackoverflow.com/questions/30245397/why-is-a-list-comprehension-so-much-faster-than-appending-to-a-list
https://realpython.com/list-comprehension-python/

vyasr:
Also @ac-optimus when you make the other changes, can you also update the Changelog?

ac-optimus:
> Also @ac-optimus when you make the other changes, can you also update the Changelog?

@bdice switched to list comprehensions. Thank a lot for sharing the links, they were really helpful. @vyasr updated the `changelog`, will take care of this from the next time.

changelog.txt
```diff
@@ -24,6 +24,11 @@ Added
 Changed
 +++++++
 
+- Raises ValueError when operation funtion passed to FlowProject.per() and FlowProject.post() and non-operation function passed to FlowPorject.pre.after() as an argument (#248, #249).
```
bdice:
```suggestion
- Raises ``ValueError`` when an operation function is passed to ``FlowProject.pre()`` and ``FlowProject.post()``, or a non-operation function passed to ``FlowProject.pre.after()`` (#248, #249).
```

bdice:
Sorry, I approved too early. I realized this PR is missing tests. @ac-optimus can you please:
- [ ] Add a test for each error
  - [ ] Provide operation to pre
  - [ ] Provide operation to post
  - [ ] Provide non-operation to pre.after
- [ ] Add yourself to the contributors list.

ac-optimus:
> Be sure to add yourself to the contributors list :)

@mikemhenry added!
@bdice I have added tests for the same, I can not come up with a good name then `test_pre_post_condition`😅 can you suggest some other, if this is not good. And can you please review the test method as well?

changelog.txt
```diff
@@ -24,6 +24,12 @@ Added
 Changed
 +++++++
 
+- Raises ``ValueError`` when an operation function is passed to ``FlowProject.pre()`` and ``FlowProject.post()``, or a non-operation function passed to ``FlowProject.pre.after()`` (#248, #249).
+- Adding test for the same in tests/test_project.py.
+
+Changed
++++++++
+
```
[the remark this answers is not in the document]
ac-optimus: @vyasr ya that was the case, I have merged with master and updated accordingly.

tests/test_project.py
```diff
@@ -519,6 +519,34 @@ def op4(job):
...
+            @A.operation
+            @A.pre.after(condition_fun)
+            def op2(job):
+                pass
+
+        with pytest.raises(ValueError):
+            @A.operation
+            @A.pre(op1)
+            def op3(job):
+                pass
```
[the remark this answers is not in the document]
vyasr: @ac-optimus looks like this is the last change request from me, then I can approve.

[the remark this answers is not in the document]
ac-optimus: @vyasr already done!

PR #253, opened 2020-02-23
Added changes to resolve #245 : Simplify Environment overrides 

### LINKED ISSUE #245: Simplify environment overrides
Currently users can override the detected environment in multiple places. In the context of a `FlowProject`, here are the possible ways that the environment to be used is determined in descending order of precedence:

1. Providing the `env` argument to `submit` and `submit_operations`.
2. Setting the environment on construction of a `FlowProject`.
3. Setting the `ENVIRONMENT` global variable in the `flow.environment` module.
4. Using the `SIGNAC_FLOW_ENVIRONMENT` environment variable.
5. Using hostname-based detection of the environment.

In the vast majority of cases, users simply fall through to (5). The ability to override environments is useful, and it is necessary for testing, so we should not remove the feature entirely. However, the presence of this many entry points is confusing, and it complicates the implementation of other features (see #224). We should reduce the number of ways that the default environment detection can be overridden.

In my opinion, option (1) should definitely be deprecated, because there is nothing that can be accomplished with that that cannot be accomplished using (2). For this reason, (2) should be maintained. (3) and (4) are somewhat redundant, but I can see reasons for potentially keeping both. We should come to a consensus on how to handle these, and then implement the resulting deprecations.

CryoZEUS:
## Description

Currently users can override the detected environment in multiple places. In the context of a FlowProject, here are the possible ways that the environment to be used is determined in descending order of precedence:

  1.  Providing the env argument to submit and submit_operations.
  2. Setting the environment on construction of a FlowProject.
  3. Setting the ENVIRONMENT global variable in the flow.environment module.
  4. Using the SIGNAC_FLOW_ENVIRONMENT environment variable.
  5. Using hostname-based detection of the environment.

Via this pull request,
1.  **option (1) has been deprecated** with an adequate DeprecationWarning.
2. **option (3) has been removed** (without Deprecation Period).
3. Otherwise, current order of precedence has been retained.

Alignment changes have been implemented as per PEP8 style guidelines.
## Motivation and Context

Issues fixed : https://github.com/glotzerlab/signac-flow/issues/245
## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

## Checklist:

If necessary:

### DIFF
--- a/contributors.yaml
+++ b/contributors.yaml
@@ -111,4 +111,8 @@ contributors:
     family-names: Ojha
     given-names: Hardik
     affiliation: "Indian Institute of Technology Roorkee"
+  -
+    family-names: Bhardwaj
+    given-names: Yash
+    affiliation: "BITS Pilani Goa Campus"
 ...
--- a/contributors.yaml
+++ b/contributors.yaml
@@ -114,5 +114,5 @@ contributors:
   -
     family-names: Bhardwaj
     given-names: Yash
-    affiliation: "BITS Pilani Goa Campus"
+    affiliation: "Birla Institute of Technology and Science, Pilani - Goa Campus"
 ...
--- a/flow/environment.py
+++ b/flow/environment.py
@@ -28,9 +28,7 @@ from .scheduling.fakescheduler import FakeScheduler
 from .util import config as flow_config
 from .errors import NoSchedulerError
 
-
 logger = logging.getLogger(__name__)
-
 
 def setup(py_modules, **attrs):
     """Setup function for environment modules.
--- a/flow/environment.py
+++ b/flow/environment.py
@@ -41,7 +41,6 @@ def setup(py_modules, **attrs):
     import setuptools
     from setuptools.command.install import install
 
-
     class InstallAndConfig(install):
 
         def run(self):
--- a/flow/environment.py
+++ b/flow/environment.py
@@ -30,6 +30,7 @@ from .errors import NoSchedulerError
 
 logger = logging.getLogger(__name__)
 
+
 def setup(py_modules, **attrs):
     """Setup function for environment modules.
 
--- a/flow/environment.py
+++ b/flow/environment.py
@@ -32,10 +32,6 @@ from .errors import NoSchedulerError
 logger = logging.getLogger(__name__)
 
 
-# Global variable can be used to override detected environment
-ENVIRONMENT = None
-
-
 def setup(py_modules, **attrs):
     """Setup function for environment modules.
 
@@ -441,9 +437,6 @@ def get_environment(test=False, import_configured=True):
     if test:
         return TestEnvironment
     else:
-        # Return a globally specified environment
-        if ENVIRONMENT is not None:
-            return ENVIRONMENT
 
         # Obtain a list of all registered environments
         env_types = registered_environments(import_configured=import_configured)
--- a/flow/environment.py
+++ b/flow/environment.py
@@ -41,6 +41,7 @@ def setup(py_modules, **attrs):
     import setuptools
     from setuptools.command.install import install
 
+
     class InstallAndConfig(install):
 
         def run(self):
--- a/flow/environment.py
+++ b/flow/environment.py
@@ -436,7 +436,6 @@ def get_environment(test=False, import_configured=True):
     if test:
         return TestEnvironment
     else:
-
         # Obtain a list of all registered environments
         env_types = registered_environments(import_configured=import_configured)
         logger.debug(
--- a/flow/project.py
+++ b/flow/project.py
@@ -320,11 +320,11 @@ class JobOperation(object):
 
     def __repr__(self):
         return "{type}(name='{name}', job='{job}', cmd={cmd}, directives={directives})".format(
-            type=type(self).__name__,
-            name=self.name,
-            job=str(self.job),
-            cmd=repr(self.cmd),
-            directives=self.directives)
+                type=type(self).__name__,
+                name=self.name,
+                job=str(self.job),
+                cmd=repr(self.cmd),
+                directives=self.directives)
 
     def __hash__(self):
         return int(sha1(self.id.encode('utf-8')).hexdigest(), 16)
@@ -1521,7 +1521,7 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
                             'scheduler_status': scheduler_status,
                             'eligible': eligible,
                             'completed': completed
-                    }
+                            }
 
         for key in sorted(status_dict):
             yield key, status_dict[key]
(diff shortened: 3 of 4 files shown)


contributors.yaml
```diff
@@ -111,4 +111,8 @@ contributors:
     family-names: Ojha
     given-names: Hardik
     affiliation: "Indian Institute of Technology Roorkee"
+  -
+    family-names: Bhardwaj
+    given-names: Yash
+    affiliation: "BITS Pilani Goa Campus"
```
bdice: If it's okay with you, I would suggest spelling out the abbrevation. It looks like [Wikipedia uses this for the official name](https://en.wikipedia.org/wiki/Birla_Institute_of_Technology_and_Science,_Pilani_%E2%80%93_Goa_Campus).
```suggestion
    affiliation: "Birla Institute of Technology and Science, Pilani - Goa Campus"
```

flow/environment.py
```diff
@@ -441,9 +437,6 @@ def get_environment(test=False, import_configured=True):
     if test:
         return TestEnvironment
     else:
-        # Return a globally specified environment
-        if ENVIRONMENT is not None:
-            return ENVIRONMENT
```
bdice: Also remove this blank line.

flow/project.py
```diff
@@ -2498,6 +2499,8 @@ def submit_operations(self, operations, _id=None, env=None, parallel=False, flag
             _id = self._store_bundled(operations)
         if env is None:
             env = self._environment
+        else:
+            print("DeprecationWarning: Providing the env argument to submit and submit_operations is deprecated. Set the environment while constructing a FlowProject", file=sys.stderr)
```
bdice: Please use a `DeprecationWarning` class instead of just a message that says "DeprecationWarning." For an example, see below. I based this message text on the format used by the `deprecation` package. We use that package for deprecating methods, but it is not possible to use that package for deprecating a single argument.
```suggestion
            warnings.warn("The env argument is deprecated as of 0.10 and will be removed in 0.12. "
                          "Instead, set the environment when constructing a FlowProject.",
                          warnings.DeprecationWarning)
```

flow/project.py
```diff
@@ -2596,6 +2599,8 @@ def submit(self, bundle_size=1, jobs=None, names=None, num=None, parallel=False,
                 "provided a single string: {}.".format(names))
         if env is None:
             env = self._environment
+        else:
+            print("DeprecationWarning: Providing the env argument to submit and submit_operations is deprecated. Set the environment while constructing a FlowProject", file=sys.stderr)
```
```suggestion
            warnings.warn("The env argument is deprecated as of 0.10 and will be removed in 0.12. "
                          "Instead, set the environment when constructing a FlowProject.",
                          warnings.DeprecationWarning)
```

contributors.yaml
CryoZEUS: Sure. I'll make the necessary changes

CryoZEUS:
@bdice Should I fix all of these issues (especially the unrelated style changes) and start a new PR?

CryoZEUS:
@bdice I am sorry about the style changes that came up because I ran autopep8. Although I fixed that part, I believe the only thing that is causing errors is the DeprecationWarning issue. I'm working on the same for now.

bdice:
@CryoZEUS You can probably fix the test errors by changing `warnings.DeprecationWarning` to `DeprecationWarning`. I copy-pasted that from someplace but it had the class name wrong. It's a built-in class (not in a module). https://docs.python.org/3/library/exceptions.html#DeprecationWarning

CryoZEUS:
@bdice Yeah, that is why added an import of `warning` at the top of `project.py` Do you think I should revert to simply `DeprecationWarning` and remove import for `warning` module as well ?

flow/project.py
```diff
@@ -2596,6 +2601,10 @@ def submit(self, bundle_size=1, jobs=None, names=None, num=None, parallel=False,
                 "provided a single string: {}.".format(names))
         if env is None:
             env = self._environment
+        else:
+            warnings.warn("The env argument is deprecated as of 0.10 and will be removed in 0.12. "
+                          "Instead, set the environment when constructing a FlowProject.",
+                          warnings.DeprecationWarning)
```
```suggestion
                          DeprecationWarning)
```

flow/project.py
```diff
@@ -2498,6 +2499,10 @@ def submit_operations(self, operations, _id=None, env=None, parallel=False, flag
             _id = self._store_bundled(operations)
         if env is None:
             env = self._environment
+        else:
+            warnings.warn("The env argument is deprecated as of 0.10 and will be removed in 0.12. "
+                          "Instead, set the environment when constructing a FlowProject.",
+                          warnings.DeprecationWarning)
```
```suggestion
                          DeprecationWarning)
```

CryoZEUS: Oh okay. I understand that now

CryoZEUS:
@bdice Is there any other change you would like to request?

flow/project.py
```diff
@@ -319,11 +320,11 @@ def __str__(self):
 
     def __repr__(self):
         return "{type}(name='{name}', job='{job}', cmd={cmd}, directives={directives})".format(
-                   type=type(self).__name__,
-                   name=self.name,
-                   job=str(self.job),
-                   cmd=repr(self.cmd),
-                   directives=self.directives)
+            type=type(self).__name__,
```
bdice: Revert whitespace changes.

flow/project.py
```diff
@@ -1520,7 +1521,7 @@ def _get_operations_status(self, job, cached_status):
                             'scheduler_status': scheduler_status,
                             'eligible': eligible,
                             'completed': completed
-                            }
+                    }
```
bdice: Revert whitespace changes.

flow/project.py
```diff
@@ -1797,7 +1798,7 @@ def print_status(self, jobs=None, overview=True, overview_max_lines=None,
                 profiling_results.append(
                     "{rank:>4} {impact:>6.0%} {code.co_filename}:"
                     "{code.co_firstlineno}:{code.co_name}".format(
-                        rank=i+1, impact=impact, code=code))
+                        rank=i + 1, impact=impact, code=code))
```
bdice: Revert whitespace changes.

CryoZEUS: Done

flow/project.py
CryoZEUS: Made the requested corrections

CryoZEUS:
@bdice Do the changes look good enough for a merge?

CryoZEUS:
@bdice Thank you! I wanted to mention that earlier, the `DeprecationWarning` that you changed was the reason I thought I'll stick to using the print statement myself. I am glad that you reviewed and corrected me about the same! Thanks again.

bdice:
@CryoZEUS We almost always have two reviewers for pull requests. I saw you removed @tcmoore3's review request but I will add him back.

CryoZEUS:
@bdice Oh, I'm so sorry, I was looking at the reviewers tab and must have removed him by mistake. I really apologize and thanks for letting me know about the same.

PR #256, opened 2020-02-25
Add --memory flag for default Slurm scheduler

jdice:
## Description
- Modified default Slurm scheduler to allow memory flag
- Removed memory flag from Xsede
- Edited argparse to allow redefinition of memory default for Great Lakes

## Motivation and Context
Setting memory requirements for job submission was not possible with default Slurm scheduler. 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change<sup>1</sup>

<sup>1</sup>The change breaks (or has the potential to break) existing functionality.

## Checklist:

- [x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac-flow/blob/master/ContributorAgreement.md).
- [x] My name is on the [list of contributors](https://github.com/glotzerlab/signac-flow/blob/master/contributors.yaml).
- [x] My code follows the [code style guideline](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md#code-style) of this project.
- [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.

If necessary:
- [ ] I have updated the API documentation as part of the package doc-strings.
- [ ] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
- [x] I have updated the [changelog](https://github.com/glotzerlab/signac-flow/blob/master/changelog.txt).

PR #257, opened 2020-02-25
Remove inconsistent assumptions about memory units being gigabytes

bdice:
## Description
Follow-up from #256 (depends on that PR). Unifies memory units by requiring the user to provide a suffix and suggesting an example value of `4g` in the help message to make that clear. **This changes the behavior on Bridges and Comet clusters, which previously implied `G`.**

## Motivation and Context
Currently the meaning of the `--memory` flag is inconsistent across environments. Some clusters (Great Lakes) require that the value includes a suffix like `g` to indicate that it's in gigabytes, while other clusters (Bridges, Comet) assume that the input is a number and append `G` to it in the template (the suffix seems to be case-insensitive).

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


flow/environment.py
```diff
@@ -344,7 +344,7 @@ def add_args(cls, parser):
         super(DefaultSlurmEnvironment, cls).add_args(parser)
         parser.add_argument(
             '--memory',
-            help=("Specify how much memory to reserve per node in GB. "
+            help=("Specify how much memory to reserve per node, e.g. \"4g\" for 4 gigabytes. "
```
tommy-waltmann: How does a user, for example, request 256MB of memory per node now?

flow/templates/bridges.sh
```diff
@@ -60,7 +60,4 @@
 {% if account %}
 #SBATCH -A {{ account }}
 {% endif %}
-{% if memory %}
```
tommy-waltmann: Why have we removed this part of the jobscript without replacing it with something else? We still need to be able to request certain amounts of memory right?

flow/environment.py
bdice: With `--memory=256m`.

flow/templates/bridges.sh
bdice: It's handled by the memory snippet in the parent template instead of by each cluster: https://github.com/glotzerlab/signac-flow/blob/1b2d44ea088e812f96da6ccb3521b70a04ecc009/flow/templates/slurm.sh#L8-L10

flow/environment.py
tommy-waltmann: Maybe it would be good to include something about MB in the help message.

bdice: I have updated the memory message.

bdice:
@vyasr @tommy-waltmann This is ready for further review.

flow/templates/bridges.sh
vyasr: @tommy-waltmann just in case you were curious, that line was added to slurm.sh by #256, which was a recent PR that this one depends on. Since the changes got split into two PRs, it's not obvious from this PR that we're not just removing the feature.

PR #260, opened 2020-02-27
remove directives default logic in class joboperation

### LINKED ISSUE #259: Remove Duplicate Directives Logic in FlowGroup and JobOperation
### Description
The logic for aggregating directives in `FlowGroup`s has superseded the logic in `JobOperation.__init__`; however, that logic still exists. We should remove this and ensure that `FlowGroup`'s methods for creating `JobOperation`s deals with this correctly.

### Proposed solution
Looking at the logic in `JobOperation.__init__` and ensuring that all the functionality exists in `FlowGroup._create_run_job_operations` and `FlowGroup._create_submit_job_operations`, before deleting the directives logic in `JobOperation.__init__`. The one exception to this is that the logic for converting directives into a `TrackGetItemDict` should remain in `JobOperation`.

michiboo:
Remove Duplicate Directives Logic in FlowGroup and JobOperation 
## Description
Remove the default directive logic in JobOperation's __init__ function

## Motivation and Context
Resolves #259 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature
- [x] Clean Code 

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -943,21 +943,19 @@ class FlowGroup(object):
             # get directives for operation
             op_dir = self._resolve_directives(name, default_directives, job)
             # Find the correct number of processors for operation
-            np = op_dir.get('np')
-
             if parallel:
                 # In general parallel means add resources
                 directives['ngpu'] += op_dir['ngpu']
                 directives['nranks'] += op_dir['nranks']
                 directives['omp_num_threads'] += op_dir['omp_num_threads']
-                directives['np'] += np
+                directives['np'] += op_dir['np']
             else:
                 # In serial we take the max
                 directives['ngpu'] = max(directives['ngpu'], op_dir['ngpu'])
                 directives['nranks'] = max(directives['nranks'], op_dir['nranks'])
                 directives['omp_num_threads'] = max(directives['omp_num_threads'],
                                                     op_dir['omp_num_threads'])
-                directives['np'] = max(directives['np'], np)
+                directives['np'] = max(directives['np'], op_dir['np'])
         return directives
 
 
--- a/flow/project.py
+++ b/flow/project.py
@@ -281,38 +281,11 @@ class JobOperation(object):
         # script engine later.
         keys_set_by_user = set(directives)
 
-        nranks = directives.get('nranks', 1)
-        nthreads = directives.get('omp_num_threads', 1)
-
-        if callable(nranks) or callable(nthreads):
-            def np_callable(job):
-                nr = nranks(job) if callable(nranks) else nranks
-                nt = nthreads(job) if callable(nthreads) else nthreads
-                return nr*nt
-
-            directives.setdefault('np', np_callable)
-        else:
-            directives.setdefault('np', nranks*nthreads)
-
-        directives.setdefault('ngpu', 0)
-        directives.setdefault('nranks', 0)
-        directives.setdefault('omp_num_threads', 0)
-        directives.setdefault('processor_fraction', 1)
-
-        # Evaluate strings and callables for job:
-        def evaluate(value):
-            if value and callable(value):
-                return value(job)
-            elif isinstance(value, str):
-                return value.format(job=job)
-            else:
-                return value
-
         # We use a special dictionary that allows us to track all keys that have been
         # evaluated by the template engine and compare them to those explicitly set
         # by the user. See also comment above.
         self.directives = TrackGetItemDict(
-            {key: evaluate(value) for key, value in directives.items()})
+            {key: value for key, value in directives.items()})
         self.directives._keys_set_by_user = keys_set_by_user
 
     def __str__(self):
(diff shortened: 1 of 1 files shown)


flow/project.py
```diff
@@ -3359,14 +3337,14 @@ def _main_script(self, args):
 
         # Gather all pending operations or generate them based on a direct command...
         with self._potentially_buffered():
+            default_directives = self._get_default_directives()
             if args.cmd:
                 warnings.warn("The --cmd option for script is deprecated as of "
                               "0.9 and will be removed in 0.11.",
                               DeprecationWarning)
-                operations = self._generate_operations(args.cmd, jobs, args.requires)
+                operations = self._generate_operations(args.cmd, jobs, default_directives, args.requires)
```
b-butler: Here the functionality is to run a command provided on the command line. As such the default directives would have no directives for it. Also the default directives here is a `dict` of `dict`s (i.e a dictionary with operation name keys and operation directives, dictionaries, as values. Having this become the directives for a `JobOperation` does not make much sense. An empty `dict` would make more sense.

flow/project.py
```diff
@@ -261,18 +261,14 @@ class JobOperation(object):
         :class:`dict`
     """
 
-    def __init__(self, id, name, job, cmd, directives=None):
+    def __init__(self, id, name, job, cmd, directives):
```
b-butler: Due to backwards compatibility issues, we need to allow for directives to be an optional argument still.

flow/project.py
```diff
@@ -261,18 +261,14 @@ class JobOperation(object):
...
         self._id = id
         self.name = name
         self.job = job
         if not (callable(cmd) or isinstance(cmd, str)):
             raise ValueError("JobOperation cmd must be a callable or string.")
         self._cmd = cmd
-
-        if directives is None:
```
b-butler: From previous comment, this logic for checking if directives have been set is still needed.

flow/project.py
```diff
@@ -2328,13 +2306,13 @@ def select(operation):
             self.run_operations(operations, pretend=pretend,
                                 np=np, timeout=timeout, progress=progress)
 
-    def _generate_operations(self, cmd, jobs, requires=None):
+    def _generate_operations(self, cmd, jobs, directives, requires=None):
```
b-butler: With previous comments this extra parameter is unnecessary.

b-butler:
Hey, @michiboo it seems that some of the logic still remains in `JobOperation` we just need to set directives to an empty dictionary if not passed, and the track item logic. The rest can go.

michiboo:
@b-butler  Hi apparently there are 3 test which failed when np value is not set in the init function of JobOperations, should I just set it to default value of 1 if directives is none? Also the test_callable_directives require the directive login in the init function of Joboperation, should I remove this test?

b-butler:
The `test_callable_directives` function fails because the directives logic is not being used in `FlowGroups._create_run_job_operation`. Once you add it there the tests should pass.

flow/project.py
```diff
@@ -280,24 +280,8 @@ def __init__(self, id, name, job, cmd, directives=None):
...
-
-            directives.setdefault('np', np_callable)
-        else:
-            directives.setdefault('np', nranks*nthreads)
-
-        directives.setdefault('ngpu', 0)
-        directives.setdefault('nranks', 0)
-        directives.setdefault('omp_num_threads', 0)
-        directives.setdefault('processor_fraction', 1)
+        if not directives.get('np'):
```
b-butler: This won't be necessary once `_create_run_job_operations` uses `_resolve_directives`.

flow/project.py
```diff
@@ -281,24 +281,6 @@ def __init__(self, id, name, job, cmd, directives=None):
...
-        else:
-            directives.setdefault('np', nranks*nthreads)
-
-        directives.setdefault('ngpu', 0)
-        directives.setdefault('nranks', 0)
-        directives.setdefault('omp_num_threads', 0)
-        directives.setdefault('processor_fraction', 1)
-
         # Evaluate strings and callables for job:
         def evaluate(value):
```
b-butler: Resolve directives should also evaluate each value in the dictionary if the value is callable.

flow/project.py
```diff
@@ -689,9 +671,21 @@ def _determine_entrypoint(self, entrypoint, directives, job):
 
     def _resolve_directives(self, name, defaults):
```
b-butler: When this also evaluates the values, we should also change the `get_directives` function in `_get_submission_directives` to not check for callable since it will have already been evaluated.

flow/project.py
```diff
@@ -687,11 +660,37 @@ def _determine_entrypoint(self, entrypoint, directives, job):
...
             return "{} {}".format(entrypoint['executable'], entrypoint['path']).lstrip()
 
-    def _resolve_directives(self, name, defaults):
+    def _resolve_directives(self, name, defaults, job):
         if name in self.operation_directives:
-            return deepcopy(self.operation_directives[name])
+            directives = deepcopy(self.operation_directives[name])
         else:
-            return deepcopy(defaults.get(name, dict()))
+            directives = defaults.get(name, dict())
```
b-butler: We want a deepcopy here to ensure that we don't write over the default dictionary provided.

flow/project.py
```diff
@@ -940,19 +939,9 @@ def _get_submission_directives(self, default_directives, job, parallel=False):
...
-            op_dir = self._resolve_directives(name, default_directives)
-            # if operations are callable handle appropriately with job
-            op_dir.update(dict(
-                ngpu=get_directive(op_dir, 'ngpu', 0, job),
-                nranks=get_directive(op_dir, 'nranks', 0, job),
-                omp_num_threads=get_directive(op_dir, 'omp_num_threads', 0, job)))
-
+            op_dir = self._resolve_directives(name, default_directives, job)
             # Find the correct number of processors for operation
             np = op_dir.get('np', max(op_dir['nranks'], 1) * max(op_dir['omp_num_threads'], 1))
```
b-butler: This logic for `np` is already in `_resolve_directives`, so we can safely remove it.

flow/project.py
```diff
@@ -940,21 +939,11 @@ def _get_submission_directives(self, default_directives, job, parallel=False):
...
-            # if operations are callable handle appropriately with job
-            op_dir.update(dict(
-                ngpu=get_directive(op_dir, 'ngpu', 0, job),
-                nranks=get_directive(op_dir, 'nranks', 0, job),
-                omp_num_threads=get_directive(op_dir, 'omp_num_threads', 0, job)))
-
+            op_dir = self._resolve_directives(name, default_directives, job)
             # Find the correct number of processors for operation
-            np = op_dir.get('np', max(op_dir['nranks'], 1) * max(op_dir['omp_num_threads'], 1))
+            np = op_dir.get('np')
```
b-butler: We can just directly use `op_dir['np']` since we are guarenteed to have a value thanks to `self._resolve_directives`.

flow/project.py
```diff
@@ -943,7 +943,7 @@ def _get_submission_directives(self, default_directives, job, parallel=False):
             # get directives for operation
             op_dir = self._resolve_directives(name, default_directives, job)
             # Find the correct number of processors for operation
-            np = op_dir.get('np')
+            np = op_dir['np']
```
b-butler: What I meant is that we didn't need to store `np` in a temporary variable.

yuanzhou0827:
LGTM! Thank you, @michiboo @b-butler

PR #269, opened 2020-03-10
Conditional use of Pool and ThreadPool

### LINKED ISSUE #264: Potential conditional use of Pool v. ThreadPool in status Parallization
### Feature description
We currently use `ThreadPool` when querying statuses i.e. `FlowProject._fetch_status`. Conditionality, allowing for process pooling instead if a label(s) is more computationally heavy than IO heavy would likely improve performance.

### Proposed solution
Currently the code that uses the `ThreadPool` could be essentially reused to work with `Pool` or `ThreadPool` given a function argument or command-line flag.

### Additional context

This idea came from a conversation I had with @justinGilmer about slow status output.

Edit: Add the function where threading currently occurs.

kidrahahjo:
Using Pool or ThreadPool conditionally

## Description
We can now use the command line to state whether we want to use I/O bound or CPU bound parallelization

## Motivation and Context

Fixes #264 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,7 +37,7 @@ jobs:
       - run:
           name: check-code-style
           command: |
-            python -m flake8 --show-source flow/
+            python -m flake8 --show-source .
 
   test-3.8: &test-template
     environment:


csadorf:
@kidrahahjo It looks like the PR assigner has failed to identify reviewers for you. We'll look into fixing this.

kidrahahjo:
> @kidrahahjo It looks like the PR assigner has failed to identify reviewers for you. We'll look into fixing this.

Thanks @csadorf 
Also, if there's something I can do to fix this bug, please do let me know.

flow/project.py
```diff
@@ -1619,13 +1619,22 @@ def _print_progress(x):
...
+                            iterable=_map(_get_job_status, jobs),
+                            desc="Collecting job status info", total=len(jobs), file=err))
+                else:
+                    with contextlib.closing(Pool()) as pool:
+                        _map = map if no_parallelize else pool.imap
+                        # First attempt at parallelized status determination.
+                        # This may fail on systems that don't allow threads.
+                        return list(tqdm(
+                            iterable=_map(_get_job_status, jobs),
+                            desc="Collecting job status info", total=len(jobs), file=err))
```
tcmoore3: The code in the if and else blocks are the same (right?). Can you refactor to get rid of that duplication. Something like

```python
if execute_ThreadPool:
    pool_ = ThreadPool()
else:
    pool_ = Pool()
with contextlib.closing(pool_) as pool:
    # do stuff
````

kidrahahjo: Will do and push the changes

kidrahahjo:
> How do you select which to use from the command line? Do we need to add an argument to `parser_status` to select which behavior to use?

@tcmoore3  I was not sure how to deal with parser_status so didn't touch that part but can surely implement that if you want.

kidrahahjo:
> The status parser is an `argparse` subparser found in `_add_print_status_args`. We would need an option to specify the parallel nature of status command: thread, process, or none. Default should be thread. You can do this looking at something similar the `IgnoreConditions` class and command line options. You would then need to change the function signature to one use one parallelization argument and the corresponding conditional statement.
> 
> Also having a test that all three modes work would be nice

As suggested by @vyasr 
I'm currently writing code for `Pool` to work because we'll be facing the error mentioned below if we don't Pickle the `_get_job_status` method manually then we'll always get this kind of error. 
> AttributeError: Can't pickle local object '_condition.isfile.<locals>.<lambda>'

I'll then proceed with what you suggested and eventually complete this pull request by writing some tests.

csadorf:
We need to make sure to have condition functions in our tests that cannot be pickled. So that this behavior is 100% tested.

kidrahahjo:
I'll make sure about that while writing the tests

b-butler:
Given the conversation on slack by @bdice and @vyasr, the choice of parallelization will be in the config variable instead of a command line option.

kidrahahjo:
@csadorf @b-butler @bdice 
I accidentally messed up with the code.
I'll create a new pull request in order to clear up this mess.
Sorry for the inconvenience

csadorf:
> @csadorf @b-butler @bdice @tcmoore3
> I accidentally messed up with the code.
> I'll create a new pull request in order to clear up this mess.
> Sorry for the inconvenience

@kidrahahjo Next time you feel like you "messed up the code", please contact me on slack or so and we can try to work this out. There are only very rare circumstances at which it is not possible to recover a situation like this using git if you know what you are doing. That is far preferable over creating a new branch and a new PR.

kidrahahjo:
@csadorf  I'll take care of it next time, sorry for the inconvenience caused.

PR #271, opened 2020-03-12
Conditional use of Pool and ThreadPool

### LINKED ISSUE #264: Potential conditional use of Pool v. ThreadPool in status Parallization

kidrahahjo:
We can now use Pool or ThreadPool conditionally.

## Description
This pull request is a consequence of #269, which was accidentally closed by me. 

## Motivation and Context

Fixes #264 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


kidrahahjo:
@csadorf @b-butler @bdice 
I've now created a new pull request associated with #264

kidrahahjo:
@vyasr 
I've made some changes as you suggested, still some of them remain hence resulting in failure of tests. Can you please have a look and let me know where was I lagging?

kidrahahjo:
@b-butler 
Can you please help me understand that what do you mean by the statement, _"choice of parallelization will be in the config variable instead of a command line option."_, [mentioned here](https://github.com/glotzerlab/signac-flow/pull/269#issuecomment-597836696)

bdice:
@kidrahahjo We also have the `signac.rc` file that provides some configuration information. A user's preference for `Pool` or `ThreadPool` isn't likely to change often, so we can set this as a configuration parameter (more permanent than a flag). You can see how that works with flow here: https://github.com/glotzerlab/signac-flow/blob/5438b047f1481292d0869a6b8889364f83dad43b/flow/util/config.py#L12-L18
You can search the codebase for some of those keys to see how they're used in practice.

kidrahahjo:
> @kidrahahjo We also have the `signac.rc` file that provides some configuration information. A user's preference for `Pool` or `ThreadPool` isn't likely to change often, so we can set this as a configuration parameter (more permanent than a flag). You can see how that works with flow here:
> 
> https://github.com/glotzerlab/signac-flow/blob/5438b047f1481292d0869a6b8889364f83dad43b/flow/util/config.py#L12-L18
> 
> 
> You can search the codebase for some of those keys to see how they're used in practice.

@bdice Okay so that means there's no point in using argparse due to not much user interaction with the variable. I'll remove the lines I edited and will get back to you once I placed my changes in the config variable. Thanks

flow/util/config.py
```diff
@@ -15,6 +15,7 @@
 status_performance_warn_threshold = float(default=0.2)
 show_traceback = boolean()
 eligible_jobs_max_lines = int(default=10)
+parallel_nature = str(default='thread')
```
csadorf: I do not think that this is a good config variable name, because it is too general and the wording is somewhat informal. I propose: *status_parallelization*.

kidrahahjo: Will change the name

flow/project.py
```diff
@@ -1613,19 +1613,58 @@ def _print_progress(x):
...
+                                except self._PickleError as error:
+                                    raise RuntimeError(
+                                        "Unable to parallelize execution due to a pickling "
+                                        "error: {}.".format(error))
+                        return list(tqdm(
+                            iterable=results,
+                            desc="Collecting job status info", total=len(jobs), file=err))
+                elif status_parallelization is None:
+                    raise RuntimeError(
+                        "Unable to parallize execution due to specified parallel_nature. "
```
b-butler: This is still the old name.

kidrahahjo:
> Since we now are offering options, with regards to the parallel nature of status updates; we should have a valid configuration option to say don't parallelize instead of an additional function parameter.

@b-butler I'll then add a configuration option with the name `no_parallelize` and remove the function parameter.

b-butler:
@kidrahahjo I don't think we need another configuration option. We could just use the same one we are using with three options `none`, `threads`, `processes`.

flow/project.py
kidrahahjo: Thank you for reminding.

vyasr:
This pull request is currently waiting on me to try and resolve some issues Hardik has had. I'll ping on this thread again once I've had a chance to resolve them.

vyasr:
I've resolved the problems with serialization. There were two separate issues. One was relatively straightforward, the `Job` instances weren't being pickled correctly (presumably because they contain a reference to the project in `_project`) so I changed it to just passing the job id through and reopen the job in the inner function.

The second issue was much harder to diagnose. As best as I can tell, the usage of multiprocessing pools to parallelize the status fetching process triggered some very subtle bug in Python's memory management whereby the template environments (which are stored in a dictionary keyed on different execution environments and reused whenever requested for the same execution environment) were decaying to weak references on the master process. It's unclear if the underlying objects were actually having their reference counts erroneously decremented (in which case the objects may or may not have been garbage collected). In any case, the template environments attached to the project instance were therefore unpickleable, so subsequent calls to `print_status` on the same project instance would fail.

@kidrahahjo you'll need to do some cleanup so that the status check mode is no longer hardcoded. You will also need to make some decision about the most sensible default choice for parallelization is, @csadorf and @b-butler can help you decide that since they're reviewing. You will also want to test the different types of parallelization (including no parallelization) to make sure they all work as expected.

kidrahahjo:
@csadorf @b-butler 
The tests that I've added are able to set the config variables as I want. (I checked this by printing out `project.config`) but when `project.print_status` is called then I have verified that `flow_config.get_config_value('status_parallelization')` always gives out the default value, i.e. thread.

tests/test_project.py
```diff
@@ -703,6 +703,26 @@ def op2(job):
         self.mock_project(project_class=B).detect_operation_graph()
 
 
+class TestProcessProject(TestProject):
```
b-butler, line 706: These classes seem a bit excessive. Could we not just have test functions in `TestProject` that just used the `config_overrides` in the `mock_project` function?

flow/project.py
```diff
@@ -1594,18 +1594,56 @@ def _print_progress(x):
...
+                                    results = self._fetch_status_in_parallel(
+                                        pool, cloudpickle, jobs, ignore_errors, cached_status)
+                                except self._PickleError as error:
+                                    raise RuntimeError(
+                                        "Unable to parallelize execution due to a pickling "
+                                        "error: {}.".format(error))
+                        return list(tqdm(
+                            iterable=results,
+                            desc="Collecting job status info", total=len(jobs), file=err))
+                else:
```
b-butler, line 1643: This means that faulty configuration values will use no parallelism if a faulty value is set. I think we should error in this case.

kidrahahjo: Here I've used else so every other case comes here. I don't think we should raise error unless we just want this to be a two-way procedure, i.e. `Process` and `Thread`.

b-butler: If we have a third option `none` then we can have an `elif` and an error raising `else`. This is better, I think, than just accepting any input besides `thread` and `process` as meaning no parallelization. We should not do something the user might find unexpected and regresses performance.

kidrahahjo: Yes, I think that would be a nice alternative. Do you want me to write tests for it too?

b-butler: That would be good. A test for error raising would also be nice.

kidrahahjo: Pushed the changes which resolve this. You can have a look.

flow/project.py
```diff
@@ -1594,19 +1594,61 @@ def _print_progress(x):
...
+                            desc="Collecting job status info", total=len(jobs), file=err))
+                elif status_parallelization == 'none':
+                    results = map(_get_job_status, jobs)
                     return list(tqdm(
-                        iterable=_map(_get_job_status, jobs),
+                        iterable=results,
                         desc="Collecting job status info", total=len(jobs), file=err))
+                else:
+                    raise RuntimeError("Configuration value status_parallelization is invalid. "
+                                       "You can set it to 'thread', 'parallel', 'none'"
```
b-butler, line 1650:
```suggestion
                                       "You can set it to 'thread', 'parallel', or 'none'"
```

flow/project.py
```diff
@@ -1764,7 +1818,7 @@ def print_status(self, jobs=None, overview=True, overview_max_lines=None,
             ]
 
             with prof(single=False):
-                tmp = self._fetch_status(jobs, err, ignore_errors, no_parallelize)
```
b-butler, line 1767: We still have the argument `no_paralleiize` in `print_status`'s function signature. We should issue a `DeprecationWarning` if it is set to True and change the  configuration value for the parallelization to `none`. We also need to change the docstring to note that the argument is official deprecated.

kidrahahjo: I'll edit the docstring. And will deprecate `no_parallelize` and edit the code

kidrahahjo:
> I noticed one more thing that needs correction to ensure that we don't break anyone's code and issue proper warnings.

@b-butler can you please help me understand which piece of code are you talking about here?

flow/project.py
```diff
@@ -1575,11 +1575,11 @@ def _fetch_scheduler_status(self, jobs=None, file=None, ignore_errors=False):
         else:
             logger.info("Updated job status cache.")
 
-    def _fetch_status(self, jobs, err, ignore_errors, no_parallelize):
+    def _fetch_status(self, jobs, err, ignore_errors, status_parallelization='none'):
```
b-butler, line 1578: no need for the argument to set the value you can just use
````python
self.config['flow']['status_parallelization'] = 'none'. It won't change the underlying configuration file.

File: tests/test_project.py [hunk 8]
```diff
@@ -627,6 +627,32 @@ def test_project_status_homogeneous_schema(self):
                 with redirect_stderr(StringIO()):
                     project.print_status(parameters=parameters, detailed=True)
 
+    def test_concurrent_project_status_homogeneous_schema(self):
```
b-butler, line 630: Technically this is testing serial execution
```suggestion
    def test_serial_project_status_homogeneous_schema(self):
```

flow/project.py
```diff
@@ -1740,6 +1797,14 @@ def print_status(self, jobs=None, overview=True, overview_max_lines=None,
         if eligible_jobs_max_lines is None:
             eligible_jobs_max_lines = flow_config.get_config_value('eligible_jobs_max_lines')
 
+        if no_parallelize:
+            warnings.warn(
+                "The no_parallelize argument is deprecated as of 0.10 "
+                "and will be removed in 0.12. "
+                "Instead, set the status_parallelization configuration value as none.",
```
b-butler, line 1804:
```suggestion
                "Instead, set the flow status_parallelization configuration value as none.",
```

flow/project.py
```diff
@@ -1740,6 +1796,16 @@ def print_status(self, jobs=None, overview=True, overview_max_lines=None,
         if eligible_jobs_max_lines is None:
             eligible_jobs_max_lines = flow_config.get_config_value('eligible_jobs_max_lines')
 
+        if no_parallelize:
+            warnings.warn(
+                "The no_parallelize argument is deprecated as of 0.10 "
+                "and will be removed in 0.12. "
+                "Instead, set the flow status_parallelization configuration value as none.",
```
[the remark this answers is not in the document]
csadorf: This is not addressed yet.

csadorf:
I was just going to merge this when I realized that one change request wasn't addressed yet and that my comment on "how does the user learn about this feature?" has also not been addressed yet. Maybe you can comment on that?

kidrahahjo:
The users will come to know about the changes only when they'll come across a Deprecation warning that says:

> The no_parallelize argument is deprecated as of 0.10 and will be removed in 0.12. Instead, set the status_parallelization configuration value to 'none'.

Although I think that this is not the best solution because we're raising this warning only when the user sets the `no_parallelize` argument equals `True`. I think that raising this message every time a user prints the status will do the trick.

There is just one testing left, which tests the config value setting behaviour using the command `signac config set flow.status_parallalization none`

csadorf:
> I think that raising this message every time a user prints the status will do the trick.

Showing a message every time seems like overkill to me. Maybe we can document this feature as part of the [regular CLI documentation for the status command](https://github.com/glotzerlab/signac-flow/blob/master/flow/project.py#L3489) and additionally as a tip [here](https://docs.signac.io/en/latest/flow-project.html#the-project-status)?

kidrahahjo:
> > I think that raising this message every time a user prints the status will do the trick.
> 
> Showing a message every time seems like overkill to me. Maybe we can document this feature as part of the [regular CLI documentation for the status command](https://github.com/glotzerlab/signac-flow/blob/master/flow/project.py#L3489) and additionally as a tip [here](https://docs.signac.io/en/latest/flow-project.html#the-project-status)?

I think editing the document would be a great solution (for the new users of signac). And for the existing users, we can update the CLI documentation for the status command.

flow/project.py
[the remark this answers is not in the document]
kidrahahjo: This works, I'll update the document.

kidrahahjo:
@csadorf 
I have one doubt, the warning which we are showing [here](https://github.com/kidrahahjo/signac-flow/blob/multiprocessing_feature/flow/project.py#L1798-L1802), should this even be a warning or a print statement?
And I have raised a DeprecationWarning here, this warning is supposed to be a UserWarning right?

Can you please let me know if I'm thinking in the right direction?

csadorf:
> I have one doubt, the warning which we are showing [her](https://github.com/kidrahahjo/signac-flow/blob/multiprocessing_feature/flow/project.py#L1798-L1802)

Using `print(..., file=sys.stderr)` is more suited to ensure that it is visible to the user. In general we use that way to communicate with the user for the CLI. 

One additional proposal: we could recommend to the user to try to switch the status parallelization method to `process` as part of [this warning](https://github.com/kidrahahjo/signac-flow/blob/54b0fab85604e7e33af7f3825bfb3bf2fd8d99a8/flow/project.py#L3372-L3380) in case that's not already the case.

kidrahahjo:
> > I have one doubt, the warning which we are showing [her](https://github.com/kidrahahjo/signac-flow/blob/multiprocessing_feature/flow/project.py#L1798-L1802)
> 
> Using `print(..., file=sys.stderr)` is more suited to ensure that it is visible to the user. In general we use that way to communicate with the user for the CLI.
> 
> One additional proposal: we could recommend to the user to try to switch the status parallelization method to `process` as part of [this warning](https://github.com/kidrahahjo/signac-flow/blob/54b0fab85604e7e33af7f3825bfb3bf2fd8d99a8/flow/project.py#L3372-L3380) in case that's not already the case.

That'd be great, I'll make the changes. Thank you.

flow/project.py
```diff
@@ -2951,7 +2955,13 @@ def _add_print_status_args(cls, parser):
...
             '--no-parallelize',
             action='store_true',
-            help="Do not parallelize the status determination.")
+            help="Do not parallelize the status determination. "
+                 "This feature is deprecated, you should set "
+                 "the flow.status_parallelization config value "
+                 "to 'none', 'thread', 'process'. You can do this by "
+                 "executing `signac config set flow.status_parallelization "
+                 "['none'|'thread'|'process']`."
+            )
```
csadorf, line 2964: Please document the feature as part of the general documentation for the `status` command and then only refer to that here. Otherwise this information will disappear after we completely remove this option.

kidrahahjo: Then I'll update this CLI document after documenting the general documentation.

flow/project.py
```diff
@@ -3574,7 +3572,12 @@ class MyProject(FlowProject):
 
         parser_status = subparsers.add_parser(
             'status',
-            parents=[base_parser])
+            parents=[base_parser],
+            help="You can now specify the parallelization of the status command "
+                 "by seting the the flow.status_parallelization config "
```
csadorf, line 3577:
```suggestion
            help="You can specify the parallelization of the status command "
                 "by setting the the flow.status_parallelization config "
```
Remove the "now" (this is not a changelog) but describes the *status quo* and there was a typo in the second line.

Also, I think we should mention that "thread" is the default value.

flow/project.py
```diff
@@ -3573,9 +3573,9 @@ class MyProject(FlowProject):
         parser_status = subparsers.add_parser(
             'status',
             parents=[base_parser],
-            help="You can now specify the parallelization of the status command "
-                 "by seting the the flow.status_parallelization config "
-                 "value to 'none', 'thread', or 'process'. You can do this by "
+            help="You can specify the parallelization of the status command "
+                 "by setting the the flow.status_parallelization config "
+                 "value to 'thread'(default), 'none', or 'process'. You can do this by "
```
csadorf, line 3578:
```suggestion
                 "by setting the flow.status_parallelization config "
                 "value to 'thread' (default), 'none', or 'process'. You can do this by "
```

PR #273, opened 2020-03-13
Use buffered behavior by default for performance in status checks.

bdice:
## Description
Improves performance in status checks by enabling signac's buffering feature.

## Motivation and Context
Status check was slow, performance profiling with `python -m cProfile -o out.prof project.py submit --pretend` showed that the slowest parts could be improved enabling buffering. Enabling buffering made the slow parts faster.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -1171,12 +1171,6 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
         self._groups = dict()
         self._register_groups()
 
-        # Enable the use of buffered mode for certain functions
-        try:
-            self._use_buffered_mode = self.config['flow'].as_bool('use_buffered_mode')
-        except KeyError:
-            self._use_buffered_mode = False
-
     def _setup_template_environment(self):
         """Setup the jinja2 template environment.
 
@@ -2393,7 +2387,8 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
 
     @contextlib.contextmanager
     def _potentially_buffered(self):
-        if self._use_buffered_mode:
+        """Enable the use of buffered mode for certain functions."""
+        if self.config['flow'].as_bool('use_buffered_mode'):
             logger.debug("Entering buffered mode...")
             with signac.buffered():
                 yield
--- a/changelog.txt
+++ b/changelog.txt
@@ -31,6 +31,7 @@ Changed
 - The command line option ``--cmd`` for ``script`` has been deprecated and will trigger a ``DeprecationWarning`` upon use until removed (#243, #218).
 - Raises ValueError when --job-name passed by the user because that interferes with status checking (#164, #241).
 - Submitting with ``--memory`` no longer assumes a unit of gigabytes on Bridges and Comet clusters (#257).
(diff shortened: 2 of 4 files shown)


bdice:
image
This image shows the code profiling results before this change. The function responsible for the slowdown (the green bar on the left) was just _hammering_ the `signac_project_document.json`, which is a 0.5 MB JSON file that was being parsed and read repeatedly. Buffering solves that.

flow/project.py
```diff
@@ -2393,7 +2387,8 @@ def _verify_group_compatibility(self, groups):
 
     @contextlib.contextmanager
     def _potentially_buffered(self):
-        if self._use_buffered_mode:
+        """Enable the use of buffered mode for certain functions."""
+        if self.config['flow'].as_bool('use_buffered_mode'):
```
bdice: The config is cached, so we don't need to have a separate private attribute `_use_buffered_mode` to store that. That attribute was only used in one place (this function).

flow/project.py
```diff
@@ -1171,12 +1171,6 @@ def __init__(self, config=None, environment=None, entrypoint=None):
         self._groups = dict()
         self._register_groups()
 
-        # Enable the use of buffered mode for certain functions
-        try:
-            self._use_buffered_mode = self.config['flow'].as_bool('use_buffered_mode')
-        except KeyError:
-            self._use_buffered_mode = False
```
csadorf: I think this might create problems if there is no 'flow' config entry.

flow/project.py
csadorf: See previous comment.

flow/project.py
bdice: @csadorf That's not possible in a `FlowProject` because we monkey-patch the config definition. In that redefinition, it always has a `[flow]` section.

csadorf: Ok, then it should be fine.

bdice:
@b-butler @csadorf @pepak13 I updated tests so they would test the unbuffered project (since the default changed). This should be ready for a second pass of review.

vyasr:
@bdice looks like this is waiting on you for a new test. 

@csadorf the obvious failure case of buffering compared to not buffering is if something actually modifies the files on disk, but I think that is explicitly out of scope, right? The implication is that we assume nothing other than the current Python instance will modify the project while the status check is running. 

What kind of race condition could be triggered here that is not already a possibility when we don't buffer? The way we structure the update farms out work on a per-job basis, so there's no central parallel-safe data store required.

bdice:
I don't think a test for that case is needed, and I'm not sure how I would even forcibly trigger such a race condition. I agree with @vyasr that testing that behavior is out of scope. I think this PR will be ready with a second reviewer approval.

csadorf:
> I don't think a test for that case is needed, and I'm not sure how I would even forcibly trigger such a race condition. I agree with @vyasr that testing that behavior is out of scope. I think this PR will be ready with a second reviewer approval.

I don't want to block this and I also couldn't come up with anything. So let's just see if we run into issues. ;)

PR #275, opened 2020-03-19
Fix issue with 'test_with_job_decorator' unit test.

csadorf:
The path comparison was done without 'realpath' wrapper, which causes
issues on Mac OS-X, because the temporary paths are represented
differently depending on context.

### DIFF
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -320,7 +320,7 @@ class TestProjectClass(TestProjectBase):
         @A.operation
         @with_job
         def test_context(job):
-            assert os.getcwd() == job.ws
+            assert os.path.realpath(os.getcwd()) == os.path.realpath(job.ws)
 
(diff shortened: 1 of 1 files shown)


vyasr:
@csadorf looks fine, but could you point me to a case that fails? I tried constructing some weird cases involving nesting symbolic links or links to links, and I didn't find one where `os.cwd() != os.path.realpath(os.getcwd())`. Where does this cause problems, and is it specifically a Mac issue?

csadorf:
> @csadorf looks fine, but could you point me to a case that fails? I tried constructing some weird cases involving nesting symbolic links or links to links, and I didn't find one where `os.cwd() != os.path.realpath(os.getcwd())`. Where does this cause problems, and is it specifically a Mac issue?

It's a specific mac issue.

Issue #270, opened 2020-03-11
Parallelism within / between groups


bdice:
### Feature description

tl;dr: We need a way to control parallelism within and between groups. Parallel operation within a group would be "intra-group" and parallel operation between groups would be "inter-group." This behavior would be controlled by the `--parallel` flag.

Copied from Slack, adapted for brevity:

@bdice: I have two operations, `equilibrate` and `sample`, in a group called `simulate`. Currently only the `equilibrate` jobs are eligible to run. I want to run `equilibrate` and then `sample`, parallelized across jobs (not parallel over both operations at the same time). That is, all jobs run `equilibrate` and then run `simulate` when that's done. On submission, it is requesting enough resources to run `equilibrate` and `sample` simultaneously, instead of sequentially (while still being parallel over jobs). How do I parallelize across groups but operate sequentially within groups?

@b-butler: The current implementation by design parallelizes across both jobs and operations since given only one flag there was not a way to specify only parallelizing part of the submission. We could implement more mutually exclusive flags or make `--parallel` have a default value.

### Proposed solution

Proposal for API from @csadorf:
`--parallel=none`: No parallel execution, default when no option is provided.
`--parallel=inter`: Parallel execution across, but not within groups; default when only --parallel is provided.
`--parallel=intra`: Parallel execution within groups, but not across.
`--parallel=all`: Parallel execution within and across groups.

csadorf:
I would like to update my proposal:

- `--parallel=none`: No parallel execution, default when no option is provided.
- `--parallel=inter-groups`: Parallel execution between, but not within groups; default when only --parallel is provided.
- `--parallel=intra-groups`: Parallel execution within groups, but not between.
- `--parallel=full`: Parallel execution between and within groups.

ac-optimus:
hi, I would like to work on this issue?

bdice:
I have talked with @ac-optimus on Slack about this issue and there are some good first steps that could be taken. Some suggestions for implementation:
- The parallelism has to be handled by both the "submit" logic (which handles parallelism between groups) and the "run" logic (which handles parallelism within a group).
[...]

Issue #265, opened 2020-03-03
Make directives module or class


b-butler:
### Feature description
Directives are the core of being able to computationally difficult jobs whether the problem is approached with MPI, OPENMP, or other parallelization schemes. Currently this logic is primarily in `FlowGroup`. Moving this to a module or just helper functions that take lists of directives or a single operation's directive, and computes correct aggregation, properly applies callable directives, etc. would be useful.

Some ideas for functionality (additions and critiques welcome):
* Aggregate directives in parallel and serial
* Check common typos (eg `ngpu` v `ngpus`)
* Get a set of unused directives with whitelisting and blacklisting handled by the function/class

### Proposed solution

I am not sure whether a class or function approach would be better. I am open to either, but I think this would make the code more modular, simple, and extensible.

kidrahahjo:
@csadorf @b-butler 
Can I try to resolve this issue?

b-butler:
@kidrahahjo I appreciate the enthusiasm, but this task will requires familiarity with cluster submission. I think that it may be best to tackle another issue.

kidrahahjo:
@b-butler  
Okay.
I think I'll follow the complete pull request which fixes this issue in order to gain more knowledge on cluster submission so that in the near future I can myself solve these kinds of issues.
Thank you

b-butler:
At a recent dev meeting we discussed the outline for what this should look like. We have concluded that directives should be their own class. In general, this class should be a mapping from string keys (e.g. `nranks`, `ngpu`, `executable`) to sensible values for each particular key (to match the previous example `int`, `int`, `str`). We also need logic to combine multiple directive instances into a single instance for both parallel and serial execution. After some discussion @vyasr and I have concluded that having a structure like what follows makes sense for our needs (our conversation did not discuss all the following details but the gist is there).

````python
class DirectivesItem:
    def __init__(self, name, type, serial_agg_method, parallel_agg_method, validation_func, ...):
        pass

class Directives:
    def __init__(self, directives_item_list):
        pass

# ...

class ComputeEnvironment:
    def _get_directives_instance(self):
        return Directives([DirectiveItems('ngpu', int, max, sum, lambda x: x > 0)])
````

[...]

Issue #250, opened 2020-02-17
Stampede failure on parallel submission with bundling


b-butler:
### Description

Due to submission scripts always using run in the groups branch, Stampede job submissions where bundling is used will fail. There is currently no way for signac to share process information to another process, so offsets can't be known within each call to `flow.run`.

### To reproduce

Bundle jobs and submit to Stampede.

### Fix
@bdice, @vyasr, and I have talked about using something like,
````python
FLOW_STAMPEDE_OFFSET=4 path/to/executable project.py run -o group -j job
FLOW_STAMPEDE_OFFSET=8 path/to/executable project.py run -o group -j job
````
and giving the Stampede environment the logic necessary to look to the environmental variable.

vyasr:
Resolved by #298.

Issue #237, opened 2020-02-03
Refactor Operation and Group registration with FlowProject


b-butler:
### Feature description
The internals of `FlowProject`'s registation of `BaseFlowOperations` and `FlowGroups` have multiple inconsistencies. Directives are stored as object attributes that are added to the callable operation. However, pre and post conditions are stored as list under `dict`s with callables as keys in the class attribute `_OPERATION_PRE_CONDITIONS`. The internal `_FlowProjectClass` metaclass says that operations are stored in dictionaries, but the actual storage is an associated list. Multiple class attributes provided by the metaclass are also unused throughout the project.

Refactoring this logic into a coherent structure that allows for extending the capabilities easily while reducing the complexity or at least centralizing it would be worth the effort.
### Proposed solution
I believe that the creation of `Registry` objects that hold lists of `RegistryItems` would be an abstract and extendible way to complete this task (the names are just for the example). Doing this would prevent us from having to add attributes to external functions, classes, and objects and allow a single point for new features and finding bugs. Ways of handling aggregation instances from the class hierarchy and converting `RegistryItems` to the final type upon instantiation a  `FlowProject` object could also be put into the `Registry` class.

Another more modest solution would just be to create the idea of `RegistryItems` contained in standard python containers (specifically `dict`s). The aggregation and conversion would then still be handled by the `FlowProject`.

b-butler:
Part of this (namely the duplication in registry objects) is handled by #251.

csadorf:
@b-butler I am having a bit trouble understanding your proposed solution. Could you maybe break this down a bit further and/or create some visual like a class diagram or so? While the current approach is certainly inconsistent, one thing to keep in mind is that certain attributes are attached to the function on purpose (like directives) because they were (at least used to be) tightly coupled to the operation (function) itself, while others only make sense in the context of a `FlowProject` class.

b-butler:
@csadorf I was thinking of something like this structure where the only relationship is ownership (not inheritance). Names are just placeholders, and not meant to be the final names necessarily.
````
FlowProject ----- FlowOperationRegistry ------ FlowOperationRegistryItem
           | 
            ----- FlowGroupRegistry ----- FlowGroupRegistryItem
````
[...]

PR #279, opened 2020-03-27
Show expected exec commands in submission script.

### LINKED ISSUE #252: Execute run with pretend in template testing
With the addition of groups, template testing is no longer as useful because all the scripts just use `python project.py run...`, which prevents us from verifying that the `mpirun`, `OMP_NUM_THREADS` and other command prefixes are correctly handled. We should add a `project.run(..., pretend=True)` called to the template test output so that we can test this behavior in the future.

### LINKED ISSUE #296: Cannot see executable in pretend mode
### Description

The pretend mode in submit does not show the actual command that get called to run an operation. I think this is because it's showing the `run` command, but that does not show the user the executable that ultimately ends up being called (especially, e.g., if you're using a container).

### To reproduce

Here is what I see with pretend:

```
 $ python project.py submit -w 20 -b 2 -p --template great-lakes.sh --pretend
Query scheduler...
Fetching operation status: 100%|█████████████████████████████████████| 8/8 [00:00<00:00, 3274.88it/s]
Submitting cluster job 'projecte/bundle/d3dd91d2bdbc0d709054fa15b120b6c437df9741':
 - Group: sample(c0e5b2b05ec0b8b2355dea01193c22c2)
 - Group: sample(5e9ec3790f463a475257c0c96772e205)
WARNING:flow.project:Some of the keys provided as part of the directives were not used by the template script, including: nranks, omp_num_threads
#!/bin/bash
#SBATCH --job-name="project/bundle/d3dd91d2bdbc0d709054fa15b120b6c437df9741"
#SBATCH --partition=standard
#SBATCH --mem=4g
#SBATCH -t 20:00:00
#SBATCH --nodes=1
#SBATCH --ntasks-per-node=8
#SBATCH --account=acct_name

#SBATCH -o output-files/--parall-c0e5b2b05ec0b8b2355dea01193c22c2.o%j

set -e
set -u

cd /path/to/project
module load singularity
export CT=`date +%s`
export WT=72000
export HOOMD_WALLTIME_STOP=`echo "$CT+$WT-60" | bc`

# sample(c0e5b2b05ec0b8b2355dea01193c22c2)
/path/to/bin/python /path/to/project/project.py run -o sample -j c0e5b2b05ec0b8b2355dea01193c22c2 --parallel &

# sample(5e9ec3790f463a475257c0c96772e205)
/path/to/bin/python /path/to/project/project.py run -o sample -j 5e9ec3790f463a475257c0c96772e205 --parallel &

wait

```

I think because of the groups update, it's showing the run command, but this is misleading, especially since I am using a container and I want to ensure that the proper executable is being called. I decorated the operation with `@directives(executable='singularity exec software-greatlakes.simg python3')`, so it would be useful to be able to see that somewhere.

### System configuration

This is on Great Lakes. 

```
$ python -c 'import platform; print(platform.platform()); import sys; print(sys.version); import signac; print(signac.__version__); import flow; print(flow.__version__)'
Linux-3.10.0-1062.9.1.el7.x86_64-x86_64-with-glibc2.10
3.8.3 | packaged by conda-forge | (default, Jun  1 2020, 17:43:00) 
[GCC 7.5.0]
1.4.0
```

bdice:
## Description
We currently do not have a way to predict what exec commands will be run at submission time. This hurts our ability to test execution behavior. I added some logic that will show the expected "exec" commands in generated submission scripts.

## Motivation and Context
Resolves #252. Resolves #296.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


bdice:
@b-butler @vyasr Is this going in the right direction? I would like some preliminary feedback. Currently this is just hacking the run operations into the `JobOperation` class, which is ugly. Do we want to make a subclass like `SubmitJobOperation` or use some other mechanism for forwarding that information along? I chose to include the run operations in the submit operation because I wanted to avoid changing from one list to two lists, which would require refactoring later steps with bundling.

vyasr:
Yes, I think the cleanest solution is to have another class that encapsulates the submission command. I also don't want to start passing around tuples. The only other reasonable alternative I can see is allowing the submit command to contain multiple `uneval_cmd`s, i.e. a submit command and a set of run commands. If you make a SubmitJobOperation, I'm not sure if it needs to contain a bunch of JobOperations or if it would be sufficient to just contain a bunch of additional unevaluated commands, but I think the latter might suffice.

b-butler:
I think in this case, we should create a new subclass as mentioned above. I also don't think we need the full `JobOperation`s for the run cmds, the strings should be fine. However, I would warn that this will miss operations that aren't currently available (i.e. only listing currently available). With some google summer of code work, maybe we can change this to be all operations that will run with some degree of certainty.

vyasr:
Hmm well if you submit a group the submission essentially assumes that all things in the group will be executed in serial and will eventually run, right? So wouldn't you just include all those commands?

If you needed something more intelligent, I think you could do is take advantage of graph detection to predict things that might run eventually. I'd hold off on that until we know that it's worthwhile, though.

b-butler:
@vyasr, I am not sure what the right solution is because we aren't guaranteeing that they all will run. The current implementation would just show the ones that are available at submission time based on the setting for `ignore_conditions_on_execution`. We would have to set the `ignore_conditions` to all.

csadorf:
Maybe it would be better to slightly revise the current way of testing:

- Add tests that specifically check the generated commands using exec mode (we can use the condition override if necessary).
- Revise the current submission tests to only test the (aggregation of) directives.

vyasr:
But there isn't any exec mode any more, is there? However, I agree that one option would be to only submit singleton groups for submission (to ensure that at least the original cases don't fail), and then separately test aggregation of directives. That would give us reasonable confidence in the correctness of our submission process.

b-butler:
We currently test "one test, one case" directives aggregation in serial and parallel. With the new directives structure, we will likely be able to test individual directives (e.g. `np`, `nranks`, etc.) much more thoroughly and easily. Also, singleton groups will still submit with the run style syntax; currently we have an unified model of `submit` -> `run` -> (`exec` or given cmd string). I agree we need more testing of the run command generation. I think that the solution is to be confident in each leg of the process I have outlined (where currently the `run` leg is the weakest with it now respecting directives).

bdice:
In an offline discussion with @csadorf @vyasr and @b-butler, we decided on some next steps.
1. Make a subclass `SubmitJobOperation` that can carry some information about its "run" operations.
2. It's not really possible to know what is going to run because eligibility of operations in a group can change during that group's execution, so the best we can do is to categorize the operations by "eligible to run now", "unmet preconditions" (which is ineligible), and "met postconditions" (which is ineligible because it's complete).
3. For testing, it would be good to have tests for the "run" logic of handling directives to ensure that the right commands are generated. This would be separate from the integration tests for submission scripts.

In this PR, I plan to do (1) and (2). I am going to leave the other tests out for now, and we can create an issue for that task.

bdice:
@b-butler @vyasr I think this is ready for you to take a look. I'm not totally happy with it -- it still feels "hacky" to me. I think there's particular room for improvement in the usage of `_create_run_job_operations`. Currently it's checking the eligibility of all job operations three times (once with the provided `IgnoreConditions`, once with the provided `IgnoreConditions` plus `PRE`, and once with the provided plus `POST`). This is then used to remove the eligible job operations and leave only those that have unmet preconditions or all met postconditions. The most obvious alternative would be to refactor `_create_run_job_operations` in a way that returns the eligibility in a way that is sort of like the inverse of `IgnoreConditions`. This shows the behavior I intended but I think another discussion is warranted before performing a refactor of that sort.

flow/project.py
```diff
@@ -261,7 +261,7 @@ class JobOperation(object):
         :class:`dict`
     """
 
-    def __init__(self, id, name, job, cmd, directives=None):
+    def __init__(self, id, name, job, cmd, directives=None, run_job_operations=None):
```
b-butler, line 264: Is there a reason the default `JobOperation` needs a `run_job_operations` list?

flow/project.py
```diff
@@ -365,6 +370,60 @@ def get_status(self):
...
+
+    :param \*args:
+        Passed to the constructor of :py:class:`JobOperation`.
+    :param eligible_operations:
+        A list of :py:class:`JobOperation` that will be executed when this
+        submitted job is executed.
+    :type eligible_operations:
+        list
+    :param operations_with_unmet_preconditions:
+        A list of :py:class:`JobOperation` that will not be executed due to
```
b-butler, line 388:
```suggestion
        A list of :py:class:`JobOperation` that may not be executed due to
```

flow/project.py
```diff
@@ -365,6 +370,60 @@ def get_status(self):
...
+        unmet preconditions.
+    :type operations_with_unmet_preconditions:
+        list
+    :param operations_with_met_postconditions:
+        A list of :py:class:`JobOperation` that will not be executed due to
```
b-butler, line 393:
```suggestion
        A list of :py:class:`JobOperation` that may not be executed due to
```

vyasr:
I would suggest creating a test project on great lakes with a large (>1000) number of jobs, submitting some subset of them with the `--hold` flag, then see how different the runtime is if you try to run the submit command with the old version vs the new one. I agree with Brandon that refactoring that function solely for this purpose doesn't make sense to me. However, I'm hopeful that the performance difference won't be that large since I expect status checks to be the slow part of submission.

We should keep in mind that the point of #252 was that we can't test even the pre-groups level of submission behavior any more. The "operations with unmet preconditions" and "operations with all postconditions met" sections would be very nice to have, but if running `_create_run_job_operation` multiple times the way it's being done is unacceptable for some reason (e.g. performance), I would vote to only do enough to resolve #252 for the purposes of testing in this PR. That means (for now) removing the additional sections from the pretend view and just keeping the "eligible operations" section if the extra calls to `_create_run_job_operations` are unacceptably slow. *If* that becomes necessary, we can always revisit when some of the other ongoing and proposed refactorings of signac-flow are done. I don't want this PR to become a blocker for downstream development.

vyasr:
@bdice @b-butler I just tried out my suggested test on Great Lakes. I created a project with 1000 jobs and a 5 operation sequence in a group, ran a limited subset of the operations for each job, then did a pretend submission of the group. There was hardly any noticeable performance difference between the two cases, suggesting that this part of the code is not currently the bottleneck for submission. As a result, I think that worrying about how efficient this branch of the code is right now is premature optimization. 

From the perspective of cleaning code, I agree that some refactoring could help in reducing the clutter; however, I think that many of the proposed follow-ups to this PR will take us in the right direction for that. We unfortunately dug ourselves into a bit of hole of technical debt with the groups PR, and that includes both issues with bugs and code in need of refactoring. Based on my benchmarks, the changes in this PR are not slow enough to substantially affect usability, so I would strongly recommend accepting these changes as is for the purpose of enabling proper testing gain so that we can fix some of the other associated issues (#270, #250) in the short term (aiming for the next minor release of flow). In the medium term, some of the other proposed refactoring (#265, #237) that is waiting on this PR should make it easier to achieve the sort of refactoring @bdice proposed to optimize this code path.

flow/project.py
vyasr: @bdice I don't think there's a need for this, is there? I think it's safe to remove.

vyasr:
@bdice @b-butler I did a little bit of cleanup on this branch and cherry-picked various changes onto separate PRs to simplify the review process. Can you take a quick look (and address the one remaining discussion item)? After that, I'll update and review the templates so that tests pass, and I think we can move forward with this as is.

flow/project.py
```diff
@@ -370,13 +370,16 @@ class SubmissionJobOperation(JobOperation):
     :type eligible_operations:
         list
     :param operations_with_unmet_preconditions:
-        A list of :py:class:`JobOperation` that will not be executed due to
-        unmet preconditions.
+        A list of :py:class:`JobOperation` that will not be executed in the
+        first pass of :meth:`FlowProject.run` due to unmet preconditions. These
```
bdice, line 374:
```suggestion
        current iteration of the :meth:`FlowProject.run` loop due to unmet preconditions. These
```

flow/project.py
```diff
@@ -370,13 +370,16 @@ class SubmissionJobOperation(JobOperation):
...
+        operations may be executed in subsequent iterations of the run loop.
     :type operations_with_unmet_preconditions:
         list
     :param operations_with_met_postconditions:
-        A list of :py:class:`JobOperation` that will not be executed due to
-        all postconditions being met.
+        A list of :py:class:`JobOperation` that will not be executed in the
+        first pass of :meth:`FlowProject.run` because all postconditions are
```
bdice, line 380:
```suggestion
        current iteration of the :meth:`FlowProject.run` loop because all postconditions are
```

flow/project.py
```diff
@@ -956,40 +959,36 @@ def _create_submission_job_operation(self, entrypoint, default_directives, job,
         uneval_cmd = functools.partial(self._submit_cmd, entrypoint=entrypoint, job=job,
                                        ignore_conditions=ignore_conditions_on_execution,
                                        parallel=parallel)
+
+        def _get_run_ops(ignore_ops, additional_ignores_flag=IgnoreConditions.NONE):
```
bdice, line 963:
```suggestion
        def _get_run_ops(ignore_ops=None, additional_ignores_flag=IgnoreConditions.NONE):
```

flow/project.py
```diff
@@ -956,40 +959,36 @@ def _create_submission_job_operation(self, entrypoint, default_directives, job,
...
+            """Get operations that match the combination of the conditions required by
+            _create_submission_job_operation and the ignored flags, and remove operations
+            in the ignore_ops list."""
+            return list(
```
bdice, line 967:
```suggestion
            if ignore_ops is None:
                ignore_ops = []
            return list(
```

flow/project.py
```diff
@@ -956,40 +959,36 @@ def _create_submission_job_operation(self, entrypoint, default_directives, job,
...
-        operations_with_met_postconditions = self._create_run_job_operations(
-            entrypoint=entrypoint,
-            default_directives=default_directives,
-            job=job,
-            ignore_conditions=IgnoreConditions(
-                ignore_conditions_on_execution | IgnoreConditions.POST)
-        )
-        operations_with_met_postconditions = \
-            set(operations_with_met_postconditions) - set(eligible_operations)
+        eligible_operations = _get_run_ops([])
```
bdice, line 977:
```suggestion
        eligible_operations = _get_run_ops()
```

flow/project.py
```diff
@@ -956,40 +959,36 @@ def _create_submission_job_operation(self, entrypoint, default_directives, job,
...
+        operations_with_unmet_preconditions = _get_run_ops(
+            eligible_operations, IgnoreConditions.PRE)
```
bdice, line 979:
```suggestion
            ignore_ops=eligible_operations,
            additional_ignores_flag=IgnoreConditions.PRE)
```

flow/project.py
```diff
@@ -956,40 +959,36 @@ def _create_submission_job_operation(self, entrypoint, default_directives, job,
...
+        operations_with_met_postconditions = _get_run_ops(
+            eligible_operations, IgnoreConditions.POST)
```
bdice, line 981:
```suggestion
            ignore_ops=eligible_operations,
            additional_ignores_flag=IgnoreConditions.POST)
```

flow/project.py
bdice: I think so. 🗑️

flow/project.py
```diff
@@ -300,6 +300,11 @@ def __init__(self, id, name, job, cmd, directives=None):
             {key: value for key, value in directives.items()})
         self.directives._keys_set_by_user = keys_set_by_user
 
+        if run_job_operations is None:
```
bdice, line 303: This block of code should be deleted, see above.

flow/project.py
bdice: We should have defaults for both arguments, or neither argument.

flow/project.py
vyasr: These operations (unmet preconditions/met postconditions) are only seen on construction of the `SubmissionJobOperation`, so they are only relevant to the status of the data space on the first pass. After the first pass they are invalidated, and we depend entirely on the `run` loop to figure out what to run next, so the original version of this docstring is correct.

flow/project.py
vyasr: See response to the comment for unmet preconditions.

vyasr:
@bdice @b-butler I've updated the template reference data and validated that it looks reasonable. The template test project doesn't currently include any actual groups, which we should add (in a separate PR). Also, currently the output for Stampede will not include the extra output this PR adds, but I think that should be addressed as part of #250

vyasr:
Actually, adding a simple test for groups wasn't hard, and I think it's relevant enough that I went ahead and added it here. Once we resolve #270 we will want to verify that inter-group parallelism results in the appropriate resource requests etc.

Issue #238, opened 2020-02-03
Refactor Execution Flow of FlowProject methods


b-butler:
### Feature description

Currently, the execution flow for `FlowProject.main` is heterogeneous. Sometimes the `FlowProject._main_{op}` only parses the command-line arguments and calls the python API function `FlowProject._main_run`, but others don't `FlowProject._main_submit`. Having one execution path that is adhered to will simplify the project and reduce redundancies in code. This also always ensures that our python and command line APIs always match or the python API is a superset of the command line API.

### Proposed solution

The execution should go something like this.

`FlowProject.main` -> `FlowProject._main_{op}` -> `FlowProject.{op}` -> any helper functions

PR #286, opened 2020-04-20
refactoring submit execution flow to include submit() function

atravitz:
partially addresses #238 

## Description

Refactoring the `submit` execution to be:

FlowProject.main() -> FlowProject._main_submit() -> FlowProject.submit() -> submit_operations()

## Motivation and Context
See #238. Currently the `submit()` function was not being called by `_main_submit()`, and `_main_submit()` was replicating the logic.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -2607,20 +2607,18 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
                 "must be a member of class IgnoreConditions")
 
         # Gather all pending operations.
-        default_directives = self._get_default_directives()
         with self._potentially_buffered():
-            operations = self._get_submission_operations(jobs, default_directives, names, parallel,
-                                                         ignore_conditions,
+            default_directives = self._get_default_directives()
+            operations = self._get_submission_operations(jobs, default_directives, names,
+                                                         parallel, ignore_conditions,
                                                          ignore_conditions_on_execution)
         if num is not None:
             operations = list(islice(operations, num))
 
         # Bundle them up and submit.
         for bundle in _make_bundles(operations, bundle_size):
-            status = self.submit_operations(
-                operations=bundle, env=env, parallel=parallel,
-                force=force, walltime=walltime, **kwargs)
-
+            status = self.submit_operations(operations=bundle, env=env, parallel=parallel,
+                                            force=force, walltime=walltime, **kwargs)
             if status is not None:  # operations were submitted, store status
                 for operation in bundle:
                     operation.set_status(status)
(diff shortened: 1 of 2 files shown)


flow/project.py
```diff
@@ -2608,6 +2608,7 @@ def submit(self, bundle_size=1, jobs=None, names=None, num=None, parallel=False,
 
         # Gather all pending operations.
         default_directives = self._get_default_directives()
+        names = args.operation_name if args.operation_name else None
```
vyasr, line 2611: Why are you using `args` here? The names are already unpacked in the function parameter list.

flow/project.py
```diff
@@ -2622,8 +2623,8 @@ def submit(self, bundle_size=1, jobs=None, names=None, num=None, parallel=False,
                 force=force, walltime=walltime, **kwargs)
 
             if status is not None:  # operations were submitted, store status
-                for op in bundle:
-                    op.set_status(status)
+                for operations in bundle:
+                    operations.set_status(status)
```
vyasr, line 2628:
```suggestion
                for operation in bundle:
                    operation.set_status(status)

```

flow/project.py
vyasr: I see where this came from in `_main_submit`. You should just be able to drop this entirely, it's getting unpacked by passing the args through to `submit`.

flow/project.py
```diff
@@ -3377,20 +3379,7 @@ def _main_submit(self, args):
...
-                                                         args.parallel, args.ignore_conditions,
-                                                         args.ignore_conditions_on_execution)
-        operations = list(islice(operations, args.num))
-        # Bundle operations up, generate the script, and submit to scheduler.
-        for bundle in make_bundles(operations, args.bundle_size):
-            status = self.submit_operations(operations=bundle, **kwargs)
-            if status is not None:
-                for operations in bundle:
-                    operations.set_status(status)
+        submit(jobs=jobs, **kwargs)
```
vyasr, line 3382:
```suggestion
        self.submit(jobs=jobs, **kwargs)
```

flow/project.py
atravitz: oh okay, this is what I was asking about the other day. thanks for clarifying!

vyasr:
@atravitz it looks like you didn't actually apply all my changes, in particular you didn't apply the change to call `self.submit` so it's still failing because there is no known `submit` function.

flow/project.py
atravitz: i moved this back to _main_submit. names has to be explicitly passed in because it's named `operation_name` in args

mikemhenry:
@vyasr I think this is ready for another look

Issue #266, opened 2020-03-03
Job Aggregation for Operations


b-butler:
This is just to include in our issues, a link to the discussion for job aggregation from a closed pull request. All discussion has be had thus far in #52  

The general idea is to allow operations to operate on multiple jobs at once.

vyasr:
I believe that the PR you want is #52, right?

csadorf:
> I believe that the PR you want is #52, right?

I edited the first comment.

b-butler:
Yes it is. I just thought a link in our issues would be helpful to show the discussion thus far on the matter @vyasr.

bdice:
Resolved by #464.

PR #289, opened 2020-05-10
WIP: Enable Aggregate Operations

kidrahahjo:
## Description

This pull request adds a new feature which allows the users to operate over a group of jobs at once
## Motivation and Context

This pull request addresses issue #266 
Users will be able to use this feature using the method below
```
def groupper_function(jobs):
    ## Returns jobs in a specified manner

aggregation_example = aggregation(grouper=groupper_function, sort='b', reverse=True)

@select(filterby=lambda job: job.sp.a > 25)
@aggregate_example
@FlowProject.operation
def example_operation(jobs):
    for job in jobs:
        print(jobs)
```
## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Work to be done:
- [x] Support for `aggregate` and `select` classes.
- [x] Introduce functionality where all operations act as aggregate operations.
- [x] Support for all pre/post conditions.
- [x]  Support for aggregate-groups, where you cannot add aggregate operations having different aggregate parameters in a single group.
- [x] Enable execution of operations and groups using `run`, `exec`.
- [x] Enable checking of jobs to be executed for any operation using `next`.
- [x] Support for `FlowCmdOperation`
- [x] Enable submission of operations
- [x] Preventing from resubmission
- [ ] Ensuring the jobs are aggregated in the same manner every time we run or submit operations
- [x] Enable status printing of operations
- [ ] **Write tests**
## Checklist:

If necessary:


flow/project.py
```diff
@@ -513,7 +513,7 @@ class FlowGroupEntry(object):
     :type name:
         str
     :param options:
-        The :py:meth:`FlowProject.run` options to pass when submitting the group.
+        The :py:meth:`FlowProject.run` options to pass when sub_flow_groupsmitting the group.
```
kidrahahjo, line 516: This is a just a typo.

csadorf:
@kidrahahjo Make sure to explicitly ping us when you want feedback/ input on this. Otherwise I'm just going to assume that this is pure drafting/ WIP and ignore related notifications.

kidrahahjo:
> @kidrahahjo Make sure to explicitly ping us when you want feedback/ input on this. Otherwise I'm just going to assume that this is pure drafting/ WIP and ignore related notifications.

@csadorf Yes, I'll make sure to do that.

kidrahahjo:
An implementation that enables the execution of aggregate operations has been achieved for now. The current iteration of do not differentiate between normal operation and aggregate operation. Every operation is an aggregate operation but in order to maintain consistency for the users, unpacking of jobs, where the length of jobs is equal to 1, is done so they can directly refer to the statepoints rather that performing indexing (`jobs[0]`)

kidrahahjo:
Few problems remain.

1. While the operations are executed as expected, I am facing an unexpected problem while executing groups. 
If I run `python project.py run -o some_operation`, where the grouper function of that operation is the pairwise function, the code runs as expected. But when I run `python project.py run -o group_name` then only the first operation of that group runs as expected.
I checked the problem and it was the list of jobs (according to the grouper function) that were created in the `_create_run_job_operations` method.
Generally, the structure for groups of N jobs is `[[a1, a2, a3, .. aN], [b1, b2, b3 ... bN], ....[...] ]`
But when executed for groups, the list is created for the operations (except the first one) `[[[a1, a2, a3, .. aN], [b1, b2, b3 ... bN], ....[...]]]`. (I am not sure why this extra nesting is taking place here)

2. While running jobs in parallel using the `Pool`, there is a pickling problem for the grouper function along with the operation function.

flow/project.py
```diff
@@ -803,18 +897,30 @@ def isdisjoint(self, group):
...
+        for job in jobs:
+            if length_not_defined:
+                upto = len(str(job.id))
+            jobstr+=str(job.id)[:upto]+'-'
+        return jobstr.strip('-')
+
+    def _generate_id(self, jobs, operation_name=None, index=0):
         "Return an id, which identifies this group with respect to this job."
-        project = job._project
+        project = jobs[0]._project
```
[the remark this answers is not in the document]
kidrahahjo: We're assuming that the jobs are from the same project. When we're done executing this, we filter the JobOperation instance and there we check whether every job is in the same project or not. Though it seems to me that I should test whether the jobs are in the same project or not here also.

flow/project.py
```diff
@@ -345,7 +422,7 @@ def set_status(self, value):
     def get_status(self):
         "Retrieve the operation's last known status."
         try:
-            return JobStatus(self.job._project.document['_status'][self.id])
+            return JobStatus(self.jobs._project.document['_status'][self.id])
```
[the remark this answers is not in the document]
kidrahahjo: This method was not used in the execution part, hence I haven't changed it. Though, self.jobs won't work here as jobs of the type `list`.

flow/project.py
```diff
@@ -933,19 +1040,32 @@ def _create_run_job_operations(self, entrypoint, default_directives, job,
...
-                # are different.
-                job_op = JobOperation(self._generate_id(job, name, index=index), name, job,
-                                      cmd=uneval_cmd, directives=deepcopy(directives))
-                yield job_op
+            jobs = self.flow_aggregate[name](jobs)
+            jobs=deeplist(list(jobs))
+            for job in jobs:
+                eligible = False
+                for j in job:
+                    if not type(j) is signac.contrib.job.Job:
```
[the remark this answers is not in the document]
kidrahahjo: I made jobs to be a nested list.
If we wanted to create a group of 2 using the grouper function then we will eventually have to iterate over something like this: `[[Job1, Job2], [Job3, Job4], [Job5, Job6], ... ]`
Hence this must be of the type `signac.contrib.job.job`

flow/project.py
```diff
@@ -2138,11 +2258,11 @@ class _PickleError(Exception):
 
     @staticmethod
     def _dumps_op(op):
-        return (op.id, op.name, op.job._id, op.cmd, op.directives)
+        return (op.id, op.name, op.jobs, op.cmd, op.directives)
```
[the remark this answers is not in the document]
kidrahahjo: Oh okay, I have a question though. Is this the reason I was facing a pickling error?

flow/project.py
```diff
@@ -2319,8 +2440,13 @@ def log(msg, lvl=logging.INFO):
         reached_execution_limit = Event()
 
         def select(operation):
-            if operation.job not in self:
-                log("Job '{}' is no longer part of the project.".format(operation.job))
+            job_in_project = True
+            for job in operation.jobs:
+                if not job in self:
+                    job_in_project = False
+                    break
```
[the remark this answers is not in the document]
kidrahahjo: I'll raise an error here.

flow/project.py
```diff
@@ -301,13 +372,19 @@ def __init__(self, id, name, job, cmd, directives=None):
         self.directives._keys_set_by_user = keys_set_by_user
 
     def __str__(self):
-        return "{}({})".format(self.name, self.job)
+        jobstr = ""
+        for job in self.jobs:
+            jobstr+="{} ".format(job.id)
+        return "{}({})".format(self.name, jobstr.strip())
```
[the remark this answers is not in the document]
kidrahahjo: Oh yes, that's a nice way to do this. Thanks! I'll make the change.

flow/project.py
```diff
@@ -301,13 +372,19 @@ def __init__(self, id, name, job, cmd, directives=None):
...
     def __repr__(self):
-        return "{type}(name='{name}', job='{job}', cmd={cmd}, directives={directives})".format(
+        jobstr = ""
+        for job in self.jobs:
+            jobstr+="{} ".format(job.id)
+        return "{type}(name='{name}', job(s)='{job}', cmd={cmd}, directives={directives})".format(
                    type=type(self).__name__,
                    name=self.name,
-                   job=str(self.job),
+                   job=jobstr.strip(),
```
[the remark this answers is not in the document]
kidrahahjo: I'll make the change

flow/project.py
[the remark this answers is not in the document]
csadorf: Might be good to make that assumption explicit with an assert. This could be a significant performance drag, but now is not the time for optimization, now is the time for code safety.

flow/project.py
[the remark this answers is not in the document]
csadorf: If it must be of that type, no need to check.

flow/project.py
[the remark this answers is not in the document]
csadorf: Probably.

flow/project.py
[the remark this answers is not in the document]
kidrahahjo: That's true, will insert assertion

flow/project.py
```diff
@@ -299,7 +303,7 @@ def keyfunction(job):
 
         def grouper(jobs):
             for key, group in groupby(sorted(jobs, key=keyfunction), key=keyfunction):
-                yield group
+                yield list(group)
```
csadorf, line 306: I think it would be much better to not rely on the fact that this is a list. Users might yield iterables as well. You can always create a list where you call the generator function.

flow/project.py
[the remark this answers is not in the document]
kidrahahjo: This worked, thank you!

flow/project.py
kidrahahjo: That's valid, I'll make this compatible.

flow/project.py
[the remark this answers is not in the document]
kidrahahjo: @csadorf I forgot to mention that this also helps when we handle `None Type` inside a group.
For instance, `[ [j1, j2, j3], [j4, j5, j6], ... [j10, None None] ]`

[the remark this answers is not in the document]
csadorf: Then it is better to explicitly check with `is None`.

flow/project.py
csadorf: Also worth noting that you should only create a list, when you really need a list. There could be literally millions of jobs in that list which would take up a significant amount of memory.

kidrahahjo: This means that there's also a need of optimization in my code because there are places where I created a list where it's not even necessary.

csadorf: Ok! Just be conscious of that, no need to optimize the code right now.

flow/project.py
```diff
@@ -236,6 +242,76 @@ def make_bundles(operations, size=None):
...
+
+    def __init__(self, grouper=None):
+        if grouper is None:
+            def grouper(jobs):
+                yield jobs
+
+        self.grouper = grouper
+
+    @classmethod
+    def groupsof(cls, num=1, fillvalue=None):
```
kidrahahjo, line 270: How should the `fillvalue` attribute be treated by the users? What all values should it take? Should we check pre/post conditions for that particular value?

kidrahahjo: I think it's best to skip these.
For instance, if I want a group of 3 and I get [joba, None, None] then it's was probably supposed to happen. Hence we shouldn't interfere much when a use asks for "user specified" arguments

flow/project.py
```diff
@@ -38,6 +38,12 @@
 from multiprocessing import Event
 import jinja2
 from jinja2 import TemplateNotFound as Jinja2TemplateNotFound
+import six
+from six.moves import zip_longest
+if six.PY2:
+    from collections import Iterable
```
b-butler, line 44: We no longer support Python 2, so six should be unnecessary.

flow/project.py
```diff
@@ -236,6 +251,119 @@ def make_bundles(operations, size=None):
...
+            def grouper(jobs):
+                yield jobs
+
+        def key_sort(job):
+            try:
+                return job.sp[sort]
+            except KeyError:
+                raise KeyError("The key '{}' was not found in statepoint "
+                               "parameters of the job {}.".format(sort, job))
+        self.grouper = grouper
```
b-butler, line 290: `grouper` should be a private member.

flow/project.py
```diff
@@ -236,6 +251,119 @@ def make_bundles(operations, size=None):
...
+        self.sort = None if sort is None else functools.partial(sorted,
```
b-butler, line 291: Likewise, here. In general, if the user shouldn't change it, it should be underscored.

flow/project.py
```diff
@@ -236,6 +251,119 @@ def make_bundles(operations, size=None):
...
+        # copied from: https://docs.python.org/3/library/itertools.html#itertools.zip_longest
+        def grouper(jobs):
+            args = [iter(jobs)] * num
+            return zip_longest(*args)
+
+        return cls(grouper, sort, reverse)
+
+    @classmethod
+    def groupby(cls, key=None, default=None, sort=None, reverse=False):
+        if isinstance(key, six.string_types):
```
b-butler, line 306: We don't need to use six here. We should use `str`

flow/project.py
```diff
@@ -236,6 +251,119 @@ def make_bundles(operations, size=None):
...
+            except KeyError:
+                raise KeyError("The key '{}' was not found in statepoint "
+                               "parameters of the job {}.".format(sort, job))
+        self.grouper = grouper
+        self.sort = None if sort is None else functools.partial(sorted,
+                                                                key=key_sort,
+                                                                reverse=reverse)
+
+    @classmethod
+    def groupsof(cls, num=1, sort=None, reverse=False):
```
b-butler, line 296: The arguments `sort` and `reverse` are never used.

flow/project.py
```diff
@@ -236,6 +251,119 @@ def make_bundles(operations, size=None):
...
+                def keyfunction(job):
+                    return job.sp.get(key, default)
+
+        elif isinstance(key, Iterable):
+            if default is None:
+                def keyfunction(job):
+                    return tuple(job.sp[k] for k in key)
+            else:
+                def keyfunction(job):
+                    return tuple(job.sp.get(k, default) for k in key)
```
b-butler, line 320: If the user, uses a generator, we will consume it on the first call of the function. We need to convert to a list first to prevent that from happening.

flow/project.py
```diff
@@ -236,6 +251,119 @@ def make_bundles(operations, size=None):
...
+    def groupsof(cls, num=1, sort=None, reverse=False):
+        # copied from: https://docs.python.org/3/library/itertools.html#itertools.zip_longest
+        def grouper(jobs):
+            args = [iter(jobs)] * num
+            return zip_longest(*args)
+
+        return cls(grouper, sort, reverse)
+
+    @classmethod
+    def groupby(cls, key=None, default=None, sort=None, reverse=False):
```
b-butler, line 305: I am assuming the reason we are not using the `JobCursor` implementation here is due to us wanting to use a list of Jobs instead?

flow/project.py
```diff
@@ -301,13 +428,13 @@ def __init__(self, id, name, job, cmd, directives=None):
         self.directives._keys_set_by_user = keys_set_by_user
 
     def __str__(self):
-        return "{}({})".format(self.name, self.job)
+        return "{}({})".format(self.name, " ".join(map(str, self.jobs)))
```
b-butler, line 431: This is fine for a first pass, but we may want to concatenate this in the case that there are many jobs.

flow/project.py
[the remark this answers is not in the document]
b-butler: You should have it work with at least one job in this PR.

flow/project.py
```diff
@@ -358,23 +485,25 @@ class FlowCondition(object):
     This can be used to build a graph of conditions and operations.
 
     :param callback:
-        A callable with one positional argument (the job).
+        A callable with one or multiple positional arguments (job or jobs).
```
b-butler, line 488: The way you have that currently

flow/project.py
```diff
@@ -358,23 +485,25 @@ class FlowCondition(object):
...
     :type callback:
         callable
     """
 
     def __init__(self, callback):
         self._callback = callback
 
-    def __call__(self, job):
+    def __call__(self, jobs):
```
b-butler, line 496: Are we internally just using the list, because this will fail if more than one positional argument is passed.

flow/project.py
```diff
@@ -446,18 +575,18 @@ def eligible(self, job, ignore_conditions=IgnoreConditions.NONE):
                 "must be a member of class IgnoreConditions")
         # len(self._prereqs) check for speed optimization
         pre = (not len(self._prereqs)) or (ignore_conditions & IgnoreConditions.PRE) \
-            or all(cond(job) for cond in self._prereqs)
+            or all(cond(jobs) for cond in self._prereqs)
```
b-butler, line 578:
```suggestion
            or all(cond(*jobs) for cond in self._prereqs)
```

flow/project.py
```diff
@@ -446,18 +575,18 @@ def eligible(self, job, ignore_conditions=IgnoreConditions.NONE):
...
         if pre and len(self._postconds):
             post = (ignore_conditions & IgnoreConditions.POST) \
-                or any(not cond(job) for cond in self._postconds)
+                or any(not cond(jobs) for cond in self._postconds)
```
b-butler, line 581:
```suggestion
                or any(not cond(*jobs) for cond in self._postconds)
```

flow/project.py
```diff
@@ -446,18 +575,18 @@ def eligible(self, job, ignore_conditions=IgnoreConditions.NONE):
...
         else:
             post = True
         return pre and post
 
-    def complete(self, job):
+    def complete(self, jobs):
         "True when all post-conditions are met."
         if len(self._postconds):
-            return all(cond(job) for cond in self._postconds)
+            return all(cond(jobs) for cond in self._postconds)
```
b-butler, line 589:
```suggestion
            return all(cond(*jobs) for cond in self._postconds)
```

flow/project.py
```diff
@@ -638,15 +767,17 @@ def op2(job):
 
     MAX_LEN_ID = 100
 
-    def __init__(self, name, operations=None, operation_directives=None,
-                 options=""):
+    def __init__(self, name, flow_aggregate=None, operations=None,
```
b-butler, line 770: Ideally, the position of existing arguments shouldn't change. `flow_aggregate` should be last or second to last.

flow/project.py
```diff
@@ -715,32 +846,42 @@ def _submit_cmd(self, entrypoint, ignore_conditions, parallel, job=None):
         else:
             return cmd.strip()
 
-    def _run_cmd(self, entrypoint, operation_name, operation, directives, job):
+    def _run_cmd(self, entrypoint, operation_name, operation, directives, jobs):
+        jobstr = deepcopy(jobs)
+        if not isinstance(jobs, signac.contrib.job.Job):
+            if not isinstance(jobs, list):
```
b-butler, line 852: When would we have a collection that is neither a list or a job?

flow/project.py
```diff
@@ -803,18 +950,28 @@ def isdisjoint(self, group):
         """
         return set(self).isdisjoint(set(group))
 
-    def _generate_id(self, job, operation_name=None, index=0):
+    def get_aggregate_hash(self, jobs):
```
b-butler, line 953: Is this user facing?

flow/project.py
```diff
@@ -803,18 +950,28 @@ def isdisjoint(self, group):
...
+        return calc_id('-'.join(map(str, jobs)))
+
+    def job_ids_seperated(self, jobs, upto=None):
```
b-butler, line 956: Is this user facing?

flow/project.py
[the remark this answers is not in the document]
b-butler: Eventually, I think this can be covered by tests, but for now safety is the right approach.

flow/project.py
```diff
@@ -803,18 +950,28 @@ def isdisjoint(self, group):
...
+        return '-'.join(map(str, jobs))
+
+    def _generate_id(self, jobs, operation_name=None, index=0):
         "Return an id, which identifies this group with respect to this job."
-        project = job._project
+        project = jobs[0]._project
+
+        for job in jobs:
+            assert project == job._project
```
b-butler, line 964: We should probably do this before we index into jobs.

flow/project.py
```diff
@@ -829,11 +986,12 @@ def _generate_id(self, job, operation_name=None, index=0):
         if max_len < len(job_op_id):
             raise ValueError("Value for MAX_LEN_ID is too small ({}).".format(self.MAX_LEN_ID))
 
+        aggregate_hash = self.get_aggregate_hash(jobs)
```
b-butler, line 989: There is really not a need for this function. We use the concatenated string in `full_name` for a unique hash.

flow/project.py
```diff
@@ -829,11 +986,12 @@ def _generate_id(self, job, operation_name=None, index=0):
...
         separator = getattr(project._environment, 'JOB_ID_SEPARATOR', '/')
         readable_name = '{project}{sep}{job}{sep}{op_string}{sep}{index:04d}{sep}'.format(
                     sep=separator,
                     project=str(project)[:12],
-                    job=str(job)[:8],
```
b-butler, line 836: This can remain as is with the change to `job_op_id`.

flow/project.py
```diff
@@ -933,19 +1091,34 @@ def _create_run_job_operations(self, entrypoint, default_directives, job,
...
-                                               operation_name=name, operation=op,
-                                               directives=directives, job=job)
-                # Uses a different id than the groups direct id. Do not use this for submitting
-                # jobs as current implementation prevents checking for resubmission in this case.
-                # The different ids allow for checking whether JobOperations created to run directly
-                # are different.
-                job_op = JobOperation(self._generate_id(job, name, index=index), name, job,
-                                      cmd=uneval_cmd, directives=deepcopy(directives))
-                yield job_op
+            filter = self.flow_select[name]
```
b-butler, line 1096: This doesn't always exist right, looking at the `__init__` function at least. If we allow users to create a `FlowGroup` without explicitly using aggregate features, we shouldn't fail here if they didn't specify those features.

flow/project.py
```diff
@@ -933,19 +1091,34 @@ def _create_run_job_operations(self, entrypoint, default_directives, job,
...
+            grouper, sort = self.flow_aggregate[name]
```
b-butler, line 1097: Isn't the grouper also responsible for sorting. This feels redundant.

flow/project.py
```diff
@@ -933,19 +1091,34 @@ def _create_run_job_operations(self, entrypoint, default_directives, job,
...
+            jobs_list = filter(jobs)
+            if sort is not None:
+                jobs_list = sort(jobs_list)
```
b-butler, line 1100: Why is this necessary?

flow/project.py
```diff
@@ -933,19 +1091,34 @@ def _create_run_job_operations(self, entrypoint, default_directives, job,
...
+            jobs_list = grouper([job for job in jobs_list])
+            jobs_list = [[j for j in job] for job in jobs_list]
```
b-butler, line 1102: From my understanding, the first line returns a generator over groups of jobs correct? if so why do we need to bother explicitly making it a list here? Also, grouper should work if passed the filter object correct?

flow/project.py
b-butler, line 1100: We should unify this sorting and grouping behavior into a single function/class, instead of relying on the users of the grouping feature to do this.

flow/project.py
```diff
@@ -933,19 +1091,34 @@ def _create_run_job_operations(self, entrypoint, default_directives, job,
...
+            for job_list in jobs_list:
+                for i, job in enumerate(job_list):
+                    if not isinstance(job, signac.contrib.job.Job):
+                        del job_list[i:]
+                        break
```
b-butler, line 1107: I would rather error out if an unexpected object is in the place of a job. If we are allowing None as valid, then explicitly check for that.

b-butler:
Also, locally for me, almost every test is failing.

flow/project.py
kidrahahjo: Okay!

flow/project.py
kidrahahjo: I agree with the second comment. The functionality here is same as the idea we all developed.

kidrahahjo: Thank you for pointing this out, I'll edit the code.

flow/project.py
kidrahahjo: Not really

flow/project.py
kidrahahjo: No

flow/project.py
kidrahahjo: I'll use the appropriate underscores.

flow/project.py
kidrahahjo: I'll use the appropriate underscores.

flow/project.py
kidrahahjo: okay!

flow/project.py
kidrahahjo: Users can create `FlowGroup` instances by using the `make_group` method of the `FlowProject` class (right?).
I have inserted statements when we register group to include these attributes.

flow/project.py
kidrahahjo: This was discussed [here](https://signac.slack.com/archives/CPCMN12EL/p1590959069395700)
If this is not what the comments meant, I'll revert my changes

flow/project.py
kidrahahjo: We're just ensuring list here. To optimize this, I can ensure the list inside the loop rather than creating a list as a whole.

kidrahahjo:
I think the first iteration of aggregation ready for a user review.
I request @glotzerlab/signac-gsoc-mentors and @csadorf to please review this.
Once I get the reviews, I'll start working on the optimization and refactoring part.
As agreed by @vyasr, I'll now remove the draft tag from this pull request.

b-butler:
@kidrahahjo just looking at the CI testing output; it seems that some change to the code is making it much slower, to the point that tests fail for taking to long in `test-3.8`. The others were killed by the CI OS/container. This could be from excessive memory usage or other conditions. I am going to try to find some areas to improve this, but this problem definitely needs to be fixed.

Running the performance test, I found that my system used at its peak about 4 GB of memory, it also ran  for well over a minute when the benchmark is less than a second per run (it runs 10 times so total run time should be under 10 seconds).

On master, my peak memory usage (this was harder to measure since the test ran quick fast) was around 77 MB, and ran in about 3 seconds.

To be clear, you do not need to optimize the code now, but such a degradation in performance is likely due to some hanging references or N^2 algorithms both of which will need to be pruned. For the performance test at least, on a normal machine with a few applications open, it could lock up the computer from memory use. In our CI testing it prevents the tests from even running since it either freezes or causes the OS to kill the process. For now though, it may be beneficial to skip that test so CI can show the other tests' status.

----

Also, locally to test to see what was happening excluding the performance test, my testing output was
> 156 failed, 56 passed, 1 deselected in 63.75s (0:01:03)

Most tests are still failing. I think we are all fine with reviewing a PR even if it isn't complete to help guide the code along, but this is certainly not ready for optimization and refactoring. We need to have passing tests to be confident about the behavior in a way that playing around locally with a branch cannot give.

It would be very helpful to us as reviewers if you used these tests for guidance on areas of the code that still have bugs. We can give high level guidance even with buggy code, but it isn't the best use of our time to search for the bugs ourselves when they are covered by existing tests.

flow/operations.py
```diff
@@ -90,6 +90,15 @@ def hello(job):
...
     """
+    signature = inspect.signature(func)
+    if len(signature.parameters) > 1:
+        raise ValueError("Only a single 'positional argument' is allowed "
+                         "when @with_job decorator is used")
+    for v in signature.parameters.values():
+        if v.kind is inspect.Parameter.VAR_POSITIONAL:
+            raise ValueError("Only a single 'positional argument' is allowed "
+                             "when @with_job decorator is used")
+
```
b-butler, line 101: This doesn't work for the case `foo(*jobs)` you need to check the single parameter's `kind` attribute. If it is a variable number of positional arguments `_ParameterKind.VAR_POSITIONAL`, then we should raise a `ValueError` as well.

flow/templates/base_status.jinja
```diff
@@ -26,14 +26,17 @@
 {% endfor %}
 {% endblock%}
 
-{% block operation_summary %}
-| operation | number of eligible jobs |
-| --------- | ----------------------- |
-{% for op, n_jobs in op_counter %}
-| {{ op }} | {{ n_jobs }} |
+{% block aggregate_summary %}
+| operation | number of eligible aggregates |
```
b-butler, line 30: I might just use the word operations here. I think it is less confusing and all operations are aggregates so it isn't false.

flow/templates/base_status.jinja
```diff
@@ -26,14 +26,17 @@
...
+| --------- | ----------------------------- |
+{% for ops, n_aggs in aggregate_counter %}
+| {{ ops }} | {{ n_aggs }} |
 {% endfor %}
 {% endblock %}
 {% endif %}
+{{ 'All operations are now aggregate operations hence you will '
+   'now see number of eligible aggregates instead of eligible jobs.'
+}}
```
b-butler, line 39: If we change the name to operations, we likely don't need this message.

flow/templates/base_status.jinja
```diff
@@ -53,6 +56,19 @@
 {% endif %}
 {% endfor %}
 {% endfor %}
+
+{{ 'Detailed Aggregate View: \n' }}
+| operation | aggregate |
+| --------- | --------- |
```
b-butler, line 62: This has two sections and the table has three this leads to incorrect output. We need something like `|operation | aggregate | status`

flow/project.py
```diff
@@ -221,8 +230,11 @@ def never(cls, func):
 
     @classmethod
     def not_(cls, condition):
-        "Returns ``not condition(job)`` for the provided condition function."
-        return cls(lambda job: not condition(job),
+        "Returns ``not condition(job)`` for the provided condition function "
+        "for every job in the aggregate"
+        def _not(*jobs):
+            return not any(condition(job) for job in jobs)
```
b-butler, line 236: I am thinking we should change how this works. All other conditions that a user provides will now need to handle potentially multiple jobs or they will fail if called with a larger aggregate. The not here is designed to work with a standard condition, so the behavior should be the same.

```suggestion
            return not condition(*jobs)
```

flow/project.py
```diff
@@ -221,8 +230,11 @@ def never(cls, func):
 
     @classmethod
     def not_(cls, condition):
-        "Returns ``not condition(job)`` for the provided condition function."
-        return cls(lambda job: not condition(job),
+        "Returns ``not condition(job)`` for the provided condition function "
+        "for every job in the aggregate"
```
b-butler, line 234: Change doc string accordingly.

flow/project.py
```diff
@@ -265,8 +277,135 @@ def make_bundles(operations, size=None):
...
+    .. code-block:: python
+        example_aggregate = aggregate()
+
+        @example_aggregate
+        @FlowProject.operation
+        def foo(jobs):
+            return len(jobs)
+
+    :param grouper:
+        Information on how to aggregate jobs.
```
b-butler, line 292: Needs more documentation about the type of callable accepted. Should it take in a list or a variable number of positional parameters for instance. Does it need to return a list of lists or can it just yield lists or single jobs instead.

flow/project.py
```diff
@@ -265,8 +277,135 @@ def make_bundles(operations, size=None):
...
+    :type grouper:
+        callable
+    :param sort:
+        Sort jobs by a statepoint parameter.
```
b-butler, line 296: When does this sort? Before or after grouping? These are the questions that people will come to the documentation with.

flow/project.py
```diff
@@ -265,8 +277,135 @@ def make_bundles(operations, size=None):
     return _make_bundles(operations, size)
 
 
+class aggregate(object):
+    """Decorator for operation functions that needs to be aggregated.
```
b-butler, line 281: It is helpful in documentation to state the default behavior of a parameter. In this case, everything is optional and the default behavior is just to pass all jobs as a single aggregate. This should be documented in the body of the documentation and the individual default behaviors mentioned with the parameters.

flow/project.py
```diff
@@ -265,8 +277,135 @@ def make_bundles(operations, size=None):
...
+        if isinstance(key, str):
+            if default is None:
+                def keyfunction(job):
+                    return job.sp[key]
+            else:
+                def keyfunction(job):
+                    return job.sp.get(key, default)
+
+        elif isinstance(key, Iterable):
+            key = list(key)
```
b-butler, line 351: We should probably use `keys` here for clarity.

flow/project.py
```diff
@@ -265,8 +277,135 @@ def make_bundles(operations, size=None):
...
+                def keyfunction(job):
+                    return [job.sp[k] for k in key]
+            else:
+                def keyfunction(job):
+                    return [job.sp.get(k, default) for k in key]
+
+        elif key is None:
+            # Must return a type that can be ordered with <, >
+            def keyfunction(job):
+                return str(job)
```
b-butler, line 362: Can't we just make `groupby` require a key argument? This behavior doesn't seem useful.

flow/project.py
```diff
@@ -265,8 +277,135 @@ def make_bundles(operations, size=None):
...
+
+        else:
+            keyfunction = key
```
b-butler, line 365: We should do a callable check here

b-butler, line 365: We should do a callable check here

flow/project.py
```diff
@@ -265,8 +277,135 @@ def make_bundles(operations, size=None):
...
+
+    .. code-block:: python
+
+        @select(filter=lambda job: job.sp.a>=5)
+        @FlowProject.operation
+        def foo(jobs):
+            return len(jobs)
+
+    :param filterby:
+        Condition for filtering jobs.
```
b-butler, line 390: Does it operate on a single job or multiple? The example code makes it clear, but it would be helpful to document it in words too.

flow/project.py
```diff
@@ -265,8 +277,135 @@ def make_bundles(operations, size=None):
...
+                yield group
+
+        return cls(grouper, sort, reverse)
+
+    def __call__(self, func):
+        setattr(func, '_flow_aggregate', (self._grouper, self._sort))
+        return func
+
+
+class select(object):
```
b-butler, line 378: Unless we are offering more options, this does not need to be a class like `aggregate`.

flow/project.py
```diff
@@ -330,13 +468,18 @@ def __init__(self, id, name, job, cmd, directives=None):
         self.directives._keys_set_by_user = keys_set_by_user
 
     def __str__(self):
-        return "{}({})".format(self.name, self.job)
+        if len(self.jobs) == 1:
+            return "{}({})".format(self.name, str(self.jobs[0]))
+        else:
+            return "{}({}...{}...{})".format(
+                self.name, str(self.jobs[0])[0:8], len(self.jobs), str(self.jobs[-1])[0:8]
```
b-butler, line 475: The format is not bad, but I would personally rather `name-job-number-ids`. We should see others opinions first though.

flow/project.py
```diff
@@ -330,13 +468,18 @@ def __init__(self, id, name, job, cmd, directives=None):
...
+            )
 
     def __repr__(self):
-        return "{type}(name='{name}', job='{job}', cmd={cmd}, directives={directives})".format(
+        return "{type}(name='{name}', job(s)='{job}', cmd={cmd}, directives={directives})".format(
```
b-butler, line 479:
```suggestion
        return "{type}(name='{name}', jobs='{jobs}', cmd={cmd}, directives={directives})".format(
```

flow/project.py
```diff
@@ -330,13 +468,18 @@ def __init__(self, id, name, job, cmd, directives=None):
...
                    type=type(self).__name__,
                    name=self.name,
-                   job=str(self.job),
+                   job=" ".join(map(str, self.jobs)),
```
b-butler, line 482:
```suggestion
                   jobs=" ".join(map(str, self.jobs)),
```
These changes are more in line with what `repr` is supposed to do versus `str`.

flow/project.py
```diff
@@ -369,12 +512,12 @@ def cmd(self):
 
     def set_status(self, value):
         "Store the operation's status."
-        self.job._project.document.setdefault('_status', dict())[self.id] = int(value)
+        self.jobs[0]._project.document.setdefault('_status', dict())[self.id] = int(value)
```
b-butler, line 515: This will definitely allow for submission time errors.

flow/project.py
```diff
@@ -369,12 +512,12 @@ def cmd(self):
...
 
     def get_status(self):
         "Retrieve the operation's last known status."
         try:
-            return JobStatus(self.job._project.document['_status'][self.id])
+            return JobStatus(self.jobs[0]._project.document['_status'][self.id])
```
b-butler, line 520: See comment about `set_status`

flow/project.py
```diff
@@ -470,14 +615,15 @@ def __eq__(self, other):
...
+    job or aggregate of jobs.
 
     Every BaseFlowOperation is associated with a specific command.
 
     Pre-requirements (pre) and post-conditions (post) can be used to
     trigger an operation only when certain conditions are met. Conditions are unary
-    callables, which expect an instance of job as their first and only positional
-    argument and return either True or False.
+    callables, which expect an instance of job or aggregate of jobs as their first
+    and only positional argument and return either True or False.
```
b-butler, line 626: This isn't true though given the other code, we expect the signature to accept multiple positional arguments for aggregates correct?

flow/project.py
```diff
@@ -552,13 +698,14 @@ class FlowCmdOperation(BaseFlowOperation):
     """A BaseFlowOperation that holds a shell executable command.
 
     When an operation has the ``@cmd`` directive specified, it is instantiated
-    as a FlowCmdOperation. The operation should be a function of
+    as a FlowCmdOperation. The operation should be a function of single or multiple
     :py:class:`~signac.contrib.job.Job`. The command (cmd) may
-    either be a unary callable that expects an instance of
-    :class:`~signac.contrib.job.Job` as its only positional argument and returns
+    either be a unary callable that expects an instance of single or multiple
```
b-butler, line 703: it isn't a unary callable if it can take multiple arguments.

flow/project.py
```diff
@@ -552,13 +698,14 @@ class FlowCmdOperation(BaseFlowOperation):
...
+    :class:`~signac.contrib.job.Job` as its positional argument and returns
```
b-butler, line 704:
```suggestion
    :class:`~signac.contrib.job.Job` as its positional argument(s) and returns
```

flow/project.py
```diff
@@ -571,11 +718,15 @@ def __init__(self, cmd, pre=None, post=None):
     def __str__(self):
         return "{type}(cmd='{cmd}')".format(type=type(self).__name__, cmd=self._cmd)
 
-    def __call__(self, job=None):
+    def __call__(self, jobs=None):
         if callable(self._cmd):
-            return self._cmd(job).format(job=job)
+            return self._cmd(*jobs).format(jobs=' '.join(map(str, jobs)))
```
b-butler, line 723:
```suggestion
            return self._cmd(*jobs).format(*jobs))
```
This is likely what is needed here. The string with spaces may not be the format needed by the end user, so we should not force that on him.

flow/project.py
```diff
@@ -571,11 +718,15 @@ def __init__(self, cmd, pre=None, post=None):
...
         else:
-            return self._cmd.format(job=job)
+            return self._cmd.format(jobs=' '.join(map(str, jobs)))
```
b-butler, line 725:
```suggestion
            return self._cmd.format(*jobs)
```

flow/project.py
```diff
@@ -675,7 +831,7 @@ class FlowGroup(object):
     """A FlowGroup represents a subset of a workflow for a project.
 
     Any :py:class:`FlowGroup` is associated with one or more instances of
-    :py:class:`BaseFlowOperation`.
+    :py:class:`BaseFlowOperation` having similar aggregations.
```
b-butler, line 834:
```suggestion
    :py:class:`BaseFlowOperation` having identical aggregations.
```

flow/project.py
```diff
@@ -616,10 +767,15 @@ class FlowGroupEntry(object):
         commands to execute.
     :type options:
         str
+    :param aggregator:
+        Aggregation parameters :py:class:`FlowGroup` to be created.
+    :type aggregator:
+        tuple
     """
-    def __init__(self, name, options=""):
+    def __init__(self, name, options="", flow_aggregate=(None, None)):
```
b-butler, line 775: What parameters here? Since we do not have an actual `Aggregate` object that knows how do do all of the aggregation and sorting, what gets passed here? In addition, it is not clear in the documentation or the code here what is represented by the members of the tuple.

I recommend we create an `Aggregate` class that after jobs have been filtered, yields the appropriate aggregates.

flow/project.py
```diff
@@ -720,55 +876,68 @@ def op2(job):
         This lets options like ``--num_passes`` to be given to a group.
     :type options:
         str
+    :param flow_aggregate:
+        The operation specific aggregate parameters specified
+        using the :py:class:`aggregate`
+    :type flow_aggregate:
+        tuple
```
b-butler, line 883: The fact that the user is passing in a tuple here is a good sign I think to have an `Aggregate` class that handles all aspects of the creating aggregates from selected jobs.

flow/project.py
```diff
@@ -783,49 +952,54 @@ def np_callable(job):
         # Evaluate strings and callables for job:
         def evaluate(value):
             if value and callable(value):
-                return value(job)
+                return value(*jobs)
             elif isinstance(value, str):
-                return value.format(job=job)
+                return value.format(jobs=' '.join(map(str, jobs)))
```
b-butler, line 957:
```suggestion
                return value.format(jobs)
```

flow/project.py
atravitz: What is the purpose of this, and would it make more sense just to have `name-number` since the job ids are not actually very helpful in this case?

flow/project.py
```diff
@@ -783,49 +952,54 @@ def np_callable(job):
...
                    name=self.name,
                    operations=' '.join(list(self.operations)),
                    directives=self.operation_directives,
-                   options=self.options)
+                   options=self.options,
+                   flow_aggregate=self.flow_aggregate,
+                   flow_select=self.flow_select)
 
-    def eligible(self, job, ignore_conditions=IgnoreConditions.NONE):
+    def eligible(self, jobs, ignore_conditions=IgnoreConditions.NONE):
```
b-butler, line 996: This is a great example of a user facing function that takes a list of jobs rather than any number of jobs. This can/will be very confusing for users. We need to make sure that we find a consistent API for users with this. This may mean hiding many more functions/methods.

flow/project.py
```diff
@@ -836,24 +1010,25 @@ def eligible(self, job, ignore_conditions=IgnoreConditions.NONE):
...
             complete).
         :rtype:
             bool
         """
-        return all(op.complete(job) for op in self)
+        return all(op.complete(jobs) for op in self)
 
-    def add_operation(self, name, operation, directives=None):
+    def add_operation(self, name, operation, directives=None,
+                      flow_aggregate=None, flow_select=None):
```
b-butler, line 1031: Users can no longer select `flow_aggregate` or `flow_select` for individual operations, correct? If that is the case it should be removed.

flow/project.py
```diff
@@ -868,11 +1043,151 @@ def add_operation(self, name, operation, directives=None):
...
+        :param flow_aggregate:
+            The operation specific aggregate parameters specified
+            using the :py:class:`aggregate`
+        :type flow_aggregate:
+            tuple
+        :param flow_select:
+            The operation specific filter parameters specified
+            using the :py:class:`select`
+        :type flow_select:
+            callable
```
b-butler, line 1055: See comment on function signature above.

flow/project.py
```diff
@@ -720,55 +876,68 @@ def op2(job):
         This lets options like ``--num_passes`` to be given to a group.
     :type options:
         str
+    :param flow_aggregate:
```
b-butler, line 879: We should just use `aggregate` and `select`. No need for the flow prefix.

flow/project.py
```diff
@@ -868,11 +1043,151 @@ def add_operation(self, name, operation, directives=None):
...
+                    flow_select.args[0].__code__.co_freevars
+                ):
+                    raise ValueError("Can't add the operation '{}' to the group '{}' "
+                                     "due to different select parameters.".
+                                     format(name, self.name))
+            except Exception:
+                if not (self.flow_select.args[0] == flow_select.args[0]):
+                    raise ValueError("Can't add the operation '{}' to the group '{}' "
+                                     "due to different select parameters.".
+                                     format(name, self.name))
```
b-butler, line 1189: Why are we in a try block to raise an exception we immediately catch and reraise the same error? Does the conditional occasionally raise errors?

flow/project.py
```diff
@@ -868,11 +1043,151 @@ def add_operation(self, name, operation, directives=None):
...
+            callable
         """
         self.operations[name] = operation
         if directives is not None:
             self.operation_directives[name] = directives
 
+        if self.flow_aggregate == (None, None):
+            self.flow_aggregate = flow_aggregate
+            self.aggregate_per_group = False
+        elif not self.aggregate_per_group:
```
b-butler, line 1064: Is this logic necessary? An operation could theoretically support multiple modes of aggregation/selecting. We should just enforce the aggregation and selecting the group was registered with. If it isn't the same as the singleton group, then that is a problem with the user's script.

flow/project.py
b-butler: Still not addressed.

flow/project.py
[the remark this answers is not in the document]
atravitz: does this mean that if two aggregates have the same job first in their lists that their ids would be identical?

flow/project.py
```diff
@@ -888,18 +1203,22 @@ def isdisjoint(self, group):
...
         # The full name is designed to be truly unique for each job-group.
         if operation_name is None:
             op_string = ''.join(sorted(list(self.operations)))
         else:
             op_string = operation_name
 
+        job_id = '-'.join(map(str, jobs))
         full_name = '{}%{}%{}%{}'.format(project.root_directory(),
-                                         job.get_id(),
+                                         job_id,
```
b-butler, line 1221: It would be simpler just to join the job ides in the format function.

flow/project.py
```diff
@@ -1039,21 +1395,32 @@ def _create_run_job_operations(self, entrypoint, default_directives, job,
...
-                # are different.
-                job_op = JobOperation(self._generate_id(job, name, index=index), name, job,
-                                      cmd=uneval_cmd, directives=deepcopy(directives))
-                yield job_op
-
-    def _get_submission_directives(self, default_directives, job):
+            # For every operation, firstly jobs are filtered according to some valid function
+            # whose information is given in the :class:`~.select`, then grouped according to
+            # the valid grouper functions and eventually sorted according to some state-point
+            # parameter whose information is given in the :class:`~.aggregate`.
```
atravitz, line 1406:
```suggestion
            # For every operation, first jobs are filtered according to the function
            # defined in the :class:`~.select`, then are grouped according to
            # the grouper functions also defined in the :class:`~select`, then are sorted according to some state point
            # parameter whose information is given in the :class:`~.aggregate`.
```

flow/project.py
b-butler, line 520: See comment above for `set_status`. The problem comes in that if the number of jobs change in between status checks, the jobs grouped together will change, and this will result in different ids.

flow/project.py
```diff
@@ -914,25 +1233,61 @@ def _generate_id(self, job, operation_name=None, index=0):
...
         # By appending the unique job_op_id, we ensure that each id is truly unique.
         return readable_name + job_op_id
 
-    def _get_status(self, job):
-        """For a given job check the groups submission status."""
+    def _get_status(self, jobs):
+        """For given jobs check the groups submission status."""
         try:
-            return JobStatus(job._project.document['_status'][self._generate_id(job)])
+            return JobStatus(jobs[0]._project.document['_status'][self._generate_id(jobs)])
```
b-butler, line 1257: If jobs have changed between the submission of a group and the status check, then this will likely miss if a job is submitted for two aggregates.

flow/project.py
```diff
@@ -914,25 +1233,61 @@ def _generate_id(self, job, operation_name=None, index=0):
...
+            jobs_list = sort(jobs_list)
+        jobs_list = grouper([job for job in jobs_list])
+        jobs_list = [[j for j in jobs] for jobs in jobs_list]
+        if not len(jobs_list):
+            return None
+        for i, job in enumerate(jobs_list[-1]):
+            if job is None:
+                del jobs_list[-1][i:]
+                break
+        return jobs_list
```
b-butler, line 1277: This should be done largely using an `Aggregate` class or an external function at least. The filtering is fine.

flow/project.py
```diff
@@ -914,25 +1233,61 @@ def _generate_id(self, job, operation_name=None, index=0):
...
         try:
-            return JobStatus(job._project.document['_status'][self._generate_id(job)])
+            return JobStatus(jobs[0]._project.document['_status'][self._generate_id(jobs)])
         except KeyError:
             return JobStatus.unknown
 
-    def _create_submission_job_operation(self, entrypoint, default_directives, job,
+    def _get_filtered_jobs(self, jobs):
+        "Returns filtered jobs for this group"
+        jobs = list(jobs)
```
b-butler, line 1263: Why are we doing this here, is the input a `JobsCursor` object?

flow/project.py
```diff
@@ -914,25 +1233,61 @@ def _generate_id(self, job, operation_name=None, index=0):
...
+        filter = self.flow_select
+        grouper, sort = self.flow_aggregate
+        jobs_list = filter(jobs)
+        if sort is not None:
+            jobs_list = sort(jobs_list)
+        jobs_list = grouper([job for job in jobs_list])
+        jobs_list = [[j for j in jobs] for jobs in jobs_list]
+        if not len(jobs_list):
+            return None
```
b-butler, line 1272: Shouldn't we just return `[]` which `jobs_list` already is here?

flow/project.py
```diff
@@ -914,25 +1233,61 @@ def _generate_id(self, job, operation_name=None, index=0):
...
+        for i, job in enumerate(jobs_list[-1]):
+            if job is None:
+                del jobs_list[-1][i:]
+                break
```
b-butler, line 1276: Why do we have to check for `None` here? Shouldn't we just have shorter lists (e.g. `[[1, 2], [3, 4], [5]]`)?

flow/project.py
```diff
@@ -914,25 +1233,61 @@ def _generate_id(self, job, operation_name=None, index=0):
...
+        return jobs_list
+
+    def _get_filtered_job_id_per_group(self, jobs, index=0):
+        """For any group, create filtered job lists.
+
+        Jobs are filtered using the attributes flow_select and flow_aggregate.
+        """
+        jobs_list = self._get_filtered_jobs(jobs)
+        if jobs_list is None:
+            return None
```
b-butler, line 1286: If we use `[]` then we don't need this.

flow/project.py
```diff
@@ -1603,16 +1972,35 @@ def _get_operations_status(self, job, cached_status):
...
+            scheduler_status = cached_status.get(group._generate_id(jobs),
                                                  JobStatus.unknown)
             for operation in group.operations:
                 if scheduler_status >= status_dict[operation]['scheduler_status']:
                     status_dict[operation] = {
                             'scheduler_status': scheduler_status,
                             'eligible': eligible,
-                            'completed': completed
+                            'completed': completed,
+                            'aggregate_jobs': '{}...{}...{}'.format(
```
b-butler, line 2000: See other comment about my suggested way of representing aggregates.

flow/project.py
```diff
@@ -1603,16 +1972,35 @@ def _get_operations_status(self, job, cached_status):
         starting_dict = functools.partial(dict, scheduler_status=JobStatus.unknown)
         status_dict = defaultdict(starting_dict)
         for group in self._groups.values():
-            completed = group.complete(job)
-            eligible = False if completed else group.eligible(job)
-            scheduler_status = cached_status.get(group._generate_id(job),
+            jobs_list = group._get_filtered_jobs(self)
```
b-butler, line 1975: Shouldn't this job operate on a `job_list` parameter now, given the signature with a `job` parameter.

flow/project.py
```diff
@@ -1663,11 +2051,11 @@ def _fetch_scheduler_status(self, jobs=None, file=None, ignore_errors=False):
             scheduler_info = {sjob.name(): sjob.status() for sjob in self.scheduler_jobs(scheduler)}
             status = dict()
             print("Query scheduler...", file=file)
-            for job in tqdm(jobs,
-                            desc="Fetching operation status",
-                            total=len(jobs), file=file):
-                for group in self._groups.values():
-                    _id = group._generate_id(job)
+            for group in self._groups.values():
```
b-butler, line 2054: Have we completely removed the progress bar?

flow/project.py
```diff
@@ -1663,11 +2051,11 @@ def _fetch_scheduler_status(self, jobs=None, file=None, ignore_errors=False):
...
+                _ids = group._get_filtered_job_id_per_group(jobs)
```
b-butler, line 2055: What happens if a subset of jobs is passed and then the aggregates created are not the ones used by default?

flow/project.py
```diff
@@ -2532,21 +2946,25 @@ def _get_submission_operations(self, jobs, default_directives, names=None,
...
                                    ignore_conditions_on_execution=IgnoreConditions.NONE):
         """Grabs JobOperations that are eligible to run from FlowGroups."""
-        for job in jobs:
-            for group in self._gather_flow_groups(names):
-                if group.eligible(job, ignore_conditions) and self._eligible_for_submission(group,
-                                                                                            job):
+        for group in self._gather_flow_groups(names):
+            jobs_list = group._get_filtered_jobs(jobs)
+            if jobs_list is None:
+                continue
```
b-butler, line 2952: If we use an empty jobs_list we don't need this.

flow/project.py
```diff
@@ -3070,24 +3488,45 @@ def labels(self, job):
 
         See also: :meth:`~.label`
         """
+        aggregates = []
```
b-butler, line 3491: Should we ask for the labels of a job now or a group of jobs? I am inclined to think the later, aggregate labels shouldn't show up for single jobs.

flow/project.py
```diff
@@ -3184,38 +3623,39 @@ def classify(self, job):
         for _label in self.labels(job):
             yield _label
 
-    def completed_operations(self, job):
-        """Determine which operations have been completed for job.
+    def completed_operations(self, jobs):
```
b-butler, line 3626: The docstring still says job instead of jobs.

flow/project.py
```diff
@@ -3184,38 +3623,39 @@ def classify(self, job):
...
                                                         index=0)
 
-    def next_operations(self, *jobs, ignore_conditions=IgnoreConditions.NONE):
+    def next_operations(self, jobs, ignore_conditions=IgnoreConditions.NONE):
         """Determine the next eligible operations for jobs.
 
         :param jobs:
-            The signac job handles.
+            Sequence of signac job handles.
         :type job:
```
b-butler, line 3657:
```suggestion
        :type jobs:
```

flow/project.py
```diff
@@ -3184,38 +3623,39 @@ def classify(self, job):
         for _label in self.labels(job):
             yield _label
 
-    def completed_operations(self, job):
-        """Determine which operations have been completed for job.
+    def completed_operations(self, jobs):
+        """Determine which operations have been completed for
+        specified jobs.
 
         :param job:
```
b-butler, line 3630:
```suggestion
        :param jobs:
```

flow/project.py
```diff
@@ -3184,38 +3623,39 @@ def classify(self, job):
...
-            The signac job handle.
+            Sequence of signac job handles.
         :type job:
```
b-butler, line 3632:
```suggestion
        :type jobs:
```

flow/project.py
```diff
@@ -3184,38 +3623,39 @@ def classify(self, job):
...
-            :class:`~signac.contrib.job.Job`
+            Iterable
```
b-butler, line 3633: Should list iterable of what?

flow/project.py
```diff
@@ -3184,38 +3623,39 @@ def classify(self, job):
...
         for name in self.operations:
             group = self._groups[name]
-            yield from group._create_run_job_operations(entrypoint=self._entrypoint, job=job,
+            yield from group._create_run_job_operations(entrypoint=self._entrypoint, jobs=jobs,
                                                         default_directives=dict(),
                                                         ignore_conditions=ignore_conditions,
                                                         index=0)
 
-    def next_operations(self, *jobs, ignore_conditions=IgnoreConditions.NONE):
+    def next_operations(self, jobs, ignore_conditions=IgnoreConditions.NONE):
```
b-butler, line 3652: It seems like this function is identical to `_job_operations` now. Do we need `_job_operations` any more?

flow/project.py
```diff
@@ -3280,11 +3720,15 @@ def hello(job):
...
-            if i and v.default is inspect.Parameter.empty:
+            if (
+                i != total_params - 1 and
+                v.kind is inspect.Parameter.VAR_POSITIONAL
+            ):
                 raise ValueError(
-                    "Only the first argument in an operation argument may not have "
-                    "a default value! ({})".format(name))
+                    "The variable parameter '{}' in the operation '{}' should be "
+                    "the last parameter!".format(k, name))
```
b-butler, line 3731: This is not technically true they could have a function `foo(a, b, *c, d=False)` and this is a valid signature that would work. We will never use the keyword only parameter `d`, but the user internally could.

flow/project.py
```diff
@@ -3369,13 +3822,37 @@ def foo(job):
             A string to append to submissions can be any valid :meth:`FlowOperation.run` option.
         :type options:
             str
+        :param aggregator:
+            Aggregator function which aggregates jobs for this group.
+        :type aggregator:
+            callable
```
b-butler, line 3828: We call it aggregator here but grouper elsewhere. We should be consistent in our names.

flow/project.py
```diff
@@ -3369,13 +3822,37 @@ def foo(job):
...
+            selector = None
+        else:
+            def key_sort(job):
+                try:
+                    return job.sp[sort]
+                except KeyError:
+                    raise KeyError("The key '{}' was not found in statepoint "
+                                   "parameters of the job {}.".format(sort, job))
+
+            selector = functools.partial(sorted, key=key_sort, reverse=reverse)
```
b-butler, line 3853: The selector isn't for sorting is it? It is for filtering?

flow/project.py
```diff
@@ -3369,13 +3822,37 @@ def foo(job):
...
+            bool
         """
         if name in cls._GROUP_NAMES:
             raise ValueError("Repeat definition of group with name '{}'.".format(name))
         else:
             cls._GROUP_NAMES.add(name)
 
-        group_entry = FlowGroupEntry(name, options)
+        if sort is None:
+            selector = None
```
b-butler, line 3844: Filter will accept None as its first argument and just return the list.

flow/project.py
```diff
@@ -3385,10 +3862,19 @@ def _register_groups(self):
         # Gather all groups from class and parent classes.
         for cls in type(self).__mro__:
             group_entries.extend(getattr(cls, '_GROUPS', []))
-
+        flow_aggregate = dict()
+        flow_select = dict()
+        for op_name, op in self._operations.items():
```
b-butler, line 3867: Do we need the flow prefixes here? I don't think it adds any clarity.

flow/project.py
```diff
@@ -3401,7 +3887,8 @@ def _register_groups(self):
                 operation_directives = getattr(func, '_flow_group_operation_directives', dict())
                 for group_name in func._flow_groups:
                     self._groups[group_name].add_operation(
-                        op_name, op, operation_directives.get(group_name, None))
+                        op_name, op, operation_directives.get(group_name, None),
+                        flow_aggregate[op_name], flow_select[op_name])
```
b-butler, line 3891: See comment in `FlowGroup.add_operation`. I don't see a need for `flow_aggregate` or `flow_select` being parameters.

flow/project.py
```diff
@@ -3385,10 +3862,19 @@ def _register_groups(self):
...
-
+        flow_aggregate = dict()
+        flow_select = dict()
+        for op_name, op in self._operations.items():
+            try:
+                flow_aggregate[op_name] = self._operation_functions[op_name]._flow_aggregate
+                flow_select[op_name] = self._operation_functions[op_name]._flow_select
+            except KeyError:
+                flow_aggregate[op_name] = op.cmd._flow_aggregate
+                flow_select[op_name] = op.cmd._flow_select
```
b-butler, line 3873: This code likely isn't necessary. This can be done in the loop below that uses `(op_name, op)` and the `func` variable.

flow/project.py
```diff
@@ -3401,7 +3887,8 @@ def _register_groups(self):
                 operation_directives = getattr(func, '_flow_group_operation_directives', dict())
                 for group_name in func._flow_groups:
                     self._groups[group_name].add_operation(
-                        op_name, op, operation_directives.get(group_name, None))
+                        op_name, op, operation_directives.get(group_name, None),
+                        flow_aggregate[op_name], flow_select[op_name])
 
             # For singleton groups add directives
             self._groups[op_name].operation_directives[op_name] = getattr(func,
```
b-butler, line 3894: We need to set the singleton groups aggregation and selection here.

flow/project.py
```diff
@@ -3417,20 +3904,20 @@ def operations(self):
     def groups(self):
         return self._groups
 
-    def _eligible_for_submission(self, flow_group, job):
-        """Determine if a flow_group is eligible for submission with a given job.
+    def _eligible_for_submission(self, flow_group, jobs):
+        """Determine if a flow_group is eligible for submission with the given jobs.
```
b-butler, line 3908: We will need to consider that the ordering of jobs can change.

flow/project.py
```diff
@@ -3491,9 +3978,12 @@ def _main_status(self, args):
...
         "Determine the jobs that are eligible for a specific operation."
-        for job in self:
-            if args.name in {op.name for op in self.next_operations(job)}:
-                print(job)
+        for op in self.next_operations(self):
+            if(args.name in op.name):
+                print("Eligible aggregates: ", end=" ")
+                for job in op.jobs:
+                    print(job, end=" ")
+                print()
```
b-butler, line 3986:
```suggestion
                print(" ".join(str(job) for job in op.jobs)

```

b-butler, line 3986:
```suggestion
                print("Eligible aggregates:", " ".join(str(job) for job in op.jobs)
```

flow/project.py
```diff
@@ -3575,18 +4065,33 @@ def _main_exec(self, args):
             except KeyError:
                 operation = self._operations[args.operation]
 
-                def operation_function(job):
-                    cmd = operation(job).format(job=job)
+                def operation_function(*jobs):
+                    cmd = operation(jobs).format(jobs=jobs)
```
b-butler, line 4069: There is no need for the format here, it exists in `FlowCmdOperation`.

flow/project.py
```diff
@@ -3575,18 +4065,33 @@ def _main_exec(self, args):
...
-            operation_function(jobs)
-        else:
-            for job in jobs:
-                operation_function(job)
+        try:
+            filter = operation_function._flow_select
+            grouper, sort = operation_function._flow_aggregate
+        except AttributeError:
+            filter = operation.cmd._flow_select
+            grouper, sort = operation.cmd._flow_aggregate
```
b-butler, line 4080: This seems to be checking indirectly the operation class of the given operation. We should just check that directly then. Also, I think we want to allow for the group to determine the aggregation and selection behavior. This does not allow for that since we cannot override that in exec.

flow/project.py
```diff
@@ -3575,18 +4065,33 @@ def _main_exec(self, args):
...
+        if sort is not None:
+            jobs_list = sort(jobs_list)
+        jobs_list = grouper([job for job in jobs_list])
+
+        for job_list in jobs_list:
+            job_list = list(job_list)
+            for i, job in enumerate(job_list):
+                if job is None:
+                    del job_list[i:]
+                    break
```
b-butler, line 4093: This code is repeated frequently. We should combine this logic into a function or class.

flow/project.py
b-butler: I meant `name-number-ids`, but Alyssa's suggestion works too, though it will be impossible to distinguish between aggregates with the same number. We are already working with that difficulty in general since the name with the ids would get much too long quickly.

flow/project.py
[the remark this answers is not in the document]
b-butler: It is only using the `job` to get to the project. Assuming that all jobs come from the same project this is fine. The problem more comes in with changing aggregates if a project adds or removes jobs, or the aggregation is not deterministic.

flow/project.py
```diff
@@ -265,8 +277,135 @@ def make_bundles(operations, size=None):
...
+
+    .. code-block:: python
+        example_aggregate = aggregate()
+
+        @example_aggregate
+        @FlowProject.operation
+        def foo(jobs):
+            return len(jobs)
+
+    :param grouper:
```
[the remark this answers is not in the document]
kidrahahjo: I'll change the name to aggregator

flow/project.py
```diff
@@ -286,10 +425,10 @@ class JobOperation(object):
...
     :type name:
         str
-    :param job:
-        The job instance associated with this operation.
-    :type job:
-        :py:class:`signac.Job`.
+    :param jobs:
+        A list of job(s) instance associated with this operation.
+    :type jobs:
+        list
```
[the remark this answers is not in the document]
kidrahahjo: That is nice point. Thank you for pointing out. I'll see what I can do here.

flow/project.py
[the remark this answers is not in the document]
kidrahahjo: @atravitz 
This can be a problem if we run in something like this `python project.py run -o op_group_of_2 -j J1 J1 J1 J1`
This behaviour seems fair to me because even if we generate similar ids, we'll just be overwriting the `id` in the project document.

flow/project.py
kidrahahjo: We're using `zip_longest` to make groups of numbers. It was designed to always generate tuples of equal length.
Hence, it automatically fills a value `None` when it doesn't have anything else to fill.

flow/project.py
kidrahahjo: @b-butler We're now dealing with aggregates. We can introduce a new progress bar which it indicates the "fetching of group status"

flow/project.py
kidrahahjo: This concept can get confusing because there are many possibilities for this one. 
If a user does as you said, then the user should expect that the aggregates that will be formed won't be the same as expected.
I will soon come up with an efficient way of dealing with this.

flow/project.py
kidrahahjo: Since labels will be having an aggregator function associated with it, hence the implementation that I'll be going forward with will also incline towards te latter part.

flow/project.py
kidrahahjo: No we don't. I'll remove `_job_operations`

flow/project.py
kidrahahjo: `aggregator` is the right choice to go with. Will edit elsewhere.

flow/project.py
kidrahahjo: I'm not sure what do you mean by this. Can you please explain?

flow/project.py
kidrahahjo: I'll merge this concept with the `aggregate` class.

flow/templates/base_status.jinja
```diff
@@ -26,14 +26,17 @@
...
+{% block aggregate_summary %}
+| operation | number of eligible aggregates |
+| --------- | ----------------------------- |
+{% for ops, n_aggs in aggregate_counter %}
+| {{ ops }} | {{ n_aggs }} |
 {% endfor %}
 {% endblock %}
 {% endif %}
+{{ 'All operations are now aggregate operations hence you will '
+   'now see number of eligible aggregates instead of eligible jobs.'
```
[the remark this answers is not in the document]
kidrahahjo: Actually I decided to go with it after the slack conversation.
https://signac.slack.com/archives/CPCMN12EL/p1593002786143600

flow/project.py
kidrahahjo: I think then there's no point limiting users by raising some kind of error. I'll remove the parameter check.

flow/operations.py
kidrahahjo:
```
@with_job
@FlowProject.operation
def op_1(*jobs):
    print('op_1')
```
@b-butler  Executing this raises an error in my case.

flow/project.py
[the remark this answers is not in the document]
kidrahahjo: @atravitz We assume here that all the jobs in the `JobOperation` instance are from the same project hence we just fetch the project from a job.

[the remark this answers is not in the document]
kidrahahjo: @b-butler, For instance, we submit the following operation
```
@aggregate.groupsof(2)
@FlowProject.operation
def foo(*jobs):
    import time
    time.sleep(5)
    print('slept well')
```

And in the meanwhile we delete a job which is present in the aggregated which is not yet submitted.
This will throw the following error which says the following for that particular aggregate.

> Using environment configuration: DefaultSlurmEnvironment
> ERROR: Encountered error during program execution: 'Did not find job with id '670d5659b2965b80171e24047ff66a40'.'
> 
> Execute with '--show-traceback' or '--debug' to get more information.
>

kidrahahjo:
This pull request has served its purpose of guiding me through my project as #336 is now ready for a review.
@b-butler @csadorf @atravitz @bdice I'm now closing this pull request.
After #336 gets merged into `feature/enable-aggregation` our final step will be to merge `feature/enable-aggregation` into master after resolving merge conflicts.

PR #290, opened 2020-05-28
Styling according to flake8 fixed

kidrahahjo:
## Description

Resolve styling erro

## Motivation and Context

Fixes this error produced by flake8

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/util/misc.py
+++ b/flow/util/misc.py
@@ -163,11 +163,11 @@ class _hashable_dict(dict):
         return hash(tuple(sorted(self.items())))
 
 
-def to_hashable(l):
+def to_hashable(line):
     # if isinstance(l, Sequence):
-    if type(l) == list:
-        return tuple(to_hashable(_) for _ in l)
-    elif type(l) == dict:
-        return _hashable_dict(l)
+    if type(line) == list:
+        return tuple(to_hashable(_) for _ in line)
+    elif type(line) == dict:
+        return _hashable_dict(line)
     else:
-        return l
+        return line
--- a/flow/util/misc.py
+++ b/flow/util/misc.py
@@ -163,11 +163,11 @@ class _hashable_dict(dict):
         return hash(tuple(sorted(self.items())))
 
 
-def to_hashable(line):
+def to_hashable(obj):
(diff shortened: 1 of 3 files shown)


kidrahahjo:
> ./tests/test_project.py:1270:20: F523 '...'.format(...) has unused arguments at position(s): 0
>assert '--num-passes=2'.format(job) in script2
This error still remains.

flow/util/misc.py
```diff
@@ -163,11 +163,11 @@ def __hash__(self):
         return hash(tuple(sorted(self.items())))
 
 
-def to_hashable(l):
+def to_hashable(line):
```
csadorf, line 166: Why did you name this argument `line`?

kidrahahjo: There's no specific reason, I was waiting for the reviews.

vyasr: I assume that flake8 is complaining because a single letter variable, but it won't fail commit hooks unless you actually try to commit a change in the line in the code where the error is.

@kidrahahjo when you change a variable name like this, it's important to give it a name that is meaningful, not just something that prevents flake8 from throwing errors. In this case, I would suggest something generic but still descriptive like `obj` since the input is an object that *may* be a list, a dict, or something else.

kidrahahjo: @vyasr okay, I'll keep that in mind.

vyasr:
> > ./tests/test_project.py:1270:20: F523 '...'.format(...) has unused arguments at position(s): 0
> > assert '--num-passes=2'.format(job) in script2
> 
> This error still remains.

The reason for this error is that the corresponding test line
```
assert '--num-passes=2'.format(job) in script2
```
has a string "--num-passes=2" where `format` tries to replace something with `job` but doesn't find anything to replace. In general, `format` expects strings to have `{}` somewhere in there to replace. In this case, the formatting seems completely superfluous since we do just want to check for `--num-passes=2`, so I would just remove the call to `format`.

kidrahahjo:
@vyasr I have pushed the changes.

PR #297, opened 2020-06-15
Remove parallel option for FlowGroup submission

b-butler:
## Description
Removes the option to run `FlowGroup`s in parallel this does not effect inter-group parallelization.

## Motivation and Context
This removes the potentially unexpected behavior of parallelizing both within and across groups when
the `parallel` flag is used #270. A more robust implementation that supports both kinds of
parallelism needs to be created for the future, but in order to release flow this should provide the
safe option (inter-group) parallelism.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

Note that I don't think this should be on the changelog since we have not released `FlowGroup`s yet,
and the current change log does not mention parallelization.

## Checklist:

If necessary:


b-butler:
Merge after #279.

b-butler:
@bdice @pepak13 this is ready for review.

flow/project.py
```diff
@@ -2804,9 +2789,10 @@ def submit(self, bundle_size=1, jobs=None, names=None, num=None, parallel=False,
 
         # Gather all pending operations.
         with self._potentially_buffered():
+
```
bdice, line 2792: Remove extra newline.
```suggestion
```

bdice:
@b-butler ⚠️ This will require a corresponding PR to signac-docs to fix this section: https://docs.signac.io/en/latest/flow-group.html?highlight=flowgroup#running-groups-in-parallel

I don't think we need to update the changelog since we're altering an unreleased feature, right?

b-butler:
@bdice that was my thoughts regarding the changelog. I just wanted to mention that this PR shouldn't have a changelog entry. I will check your PR assuming, I don't see anything merge both.

PR #305, opened 2020-06-27
Drop Support for Python 3.5

bdice:
## Description
This PR drops support for Python 3.5 and adopts NEP 29.

## Types of Changes

- [x] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,9 +1,21 @@
-=======
+======
 Changes
 =======
 
 The **signac-flow** package follows `semantic versioning <https://semver.org/>`_.
 The numbers in brackets denote the related GitHub issue and/or pull request.
+
+next
+====
+
+next
+----
+
+Removed
++++++++
+
+- Dropped support for Python 3.5 (#305). The signac project will follow the `NEP 29 deprecation policy <https://numpy.org/neps/nep-0029-deprecation_policy.html>`_ going forward.
+
 
 Version 0.10
 ============
--- a/changelog.txt
+++ b/changelog.txt
@@ -14,7 +14,7 @@ next
 Removed
 +++++++
 
-- Dropped support for Python 3.5 (#305). The signac project will follow the `NEP 29 deprecation policy <https://numpy.org/neps/nep-0029-deprecation_policy.html>`_ going forward.
+- Drop support for Python 3.5 (#305). The signac project will follow the `NEP 29 deprecation policy <https://numpy.org/neps/nep-0029-deprecation_policy.html>`_ going forward.
 
 
 Version 0.10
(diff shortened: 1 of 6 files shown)


b-butler:
This partially depends on if we want to support testing on other Python implementations (besides CPython), but we could also remove `OrderedDict` in this PR. That should give us a speed up without much work, a free lunch if you will.

changelog.txt
```diff
@@ -1,10 +1,22 @@
...
+======
 Changes
 =======
 
 The **signac-flow** package follows `semantic versioning <https://semver.org/>`_.
 The numbers in brackets denote the related GitHub issue and/or pull request.
 
+next
+====
+
```
csadorf, line 10: This should not be there.
```suggestion
```

changelog.txt
```diff
@@ -1,10 +1,22 @@
...
+next
+----
+
+Removed
++++++++
+
+- Dropped support for Python 3.5 (#305). The signac project will follow the `NEP 29 deprecation policy <https://numpy.org/neps/nep-0029-deprecation_policy.html>`_ going forward.
```
csadorf, line 17:
```suggestion
- Drop support for Python 3.5 (#305). The signac project will follow the `NEP 29 deprecation policy <https://numpy.org/neps/nep-0029-deprecation_policy.html>`_ going forward.
```
Changelog entries should be in present tense, see other entries (mostly consistent).

changelog.txt
bdice: This is necessary to have twice, otherwise the RST is incorrectly parsed. Notice that we have two levels of indentation, one for minor versions and one for bugfix versions.

bdice:
> 
> 
> This partially depends on if we want to support testing on other Python implementations (besides CPython), but we could also remove `OrderedDict` in this PR. That should give us a speed up without much work, a free lunch if you will.

The specification for dicts being ordered by default is not until 3.7 (it's only an implementation detail in 3.6). @b-butler Could you make an issue for this (to be resolved upon dropping 3.6 support) since `OrderedDict` is reportedly [implemented in pure Python and 10x slower than `dict`](https://stackoverflow.com/questions/18422995/why-is-ordereddict-10x-slower-than-dict-and-list)? I don't think this is a blocking issue (or one we can deal with yet) and I will merge this PR.

b-butler:
@bdice I am aware that the specification is only for 3.7, and in 3.6 is an implementation detail. I just was not aware that the ordering was an API guarantee (I don't think we make one about it anywhere), it is just a requirement for our tests to run properly. If we only test on 3.6 and above using CPython we should be fine.

PR #306, opened 2020-06-30
WIP: Add Aggregation Feature to Flow

kidrahahjo:
## Description
Implementation of helper class for aggregation, `Aggregate`.
This class includes features like:
1.  _Aggregating by numbers_
2. _Aggregating by a condition/statepoint-param_ 
3. _Sorting by a standpoint parameter_
4. _Filtering jobs using the `select` parameter_

## Motivation and Context

The current [PR](https://github.com/glotzerlab/signac-flow/pull/289) for aggregation is too big. Now that PR will track my project and I'll be splitting the work into several small PRs.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


flow/aggregate.py
```diff
@@ -0,0 +1,153 @@
...
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+from collections.abc import Iterable
+from itertools import groupby
+from itertools import zip_longest
+from functools import partial
+
+
+class _aggregate:
+    """Decorator for operation functions that needs to be aggregated.
```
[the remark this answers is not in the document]
bdice: *are*, because "functions" is plural.
```suggestion
    """Decorator for operation functions that are to be aggregated.
```

flow/aggregate.py
```diff
@@ -0,0 +1,153 @@
+# Copyright (c) 2018 The Regents of the University of Michigan
```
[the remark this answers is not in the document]
kidrahahjo: Oh yes, sorry.

flow/aggregate.py
```diff
@@ -0,0 +1,183 @@
...
+import signac
+from signac.contrib.project import Project, JobsCursor
+
+
+class aggregate:
+    """Decorator for operation functions that are to be aggregated.
+
+    If this class is used for aggregation then by-default, if the aggregator
+    parameter is not passed, an aggregate of all jobs will be created.
+
```
atravitz, line 19:
```suggestion
    By default, if the aggregator parameter is not passed, 
    an aggregate of all jobs will be created.

```

flow/aggregate.py
```diff
@@ -0,0 +1,183 @@
...
+from copy import deepcopy
+from itertools import groupby
+from itertools import zip_longest
+from functools import partial
+
+import signac
+from signac.contrib.project import Project, JobsCursor
+
+
+class aggregate:
```
atravitz, line 14:
```suggestion
class Aggregate:
```

should this be capitalized because it's a class?

flow/aggregate.py
```diff
@@ -0,0 +1,183 @@
...
+        example_aggregate = aggregate()
+        @example_aggregate
+        @FlowProject.operation
+        def foo(*jobs):
+            print(len(jobs))
+
+    :param aggregator:
+        Information on how to aggregate jobs. Takes in a list of
+        jobs and can return or yield lists or single jobs instead.
+        The default behavior is creating an aggregate of all jobs.
```
atravitz, line 31:
```suggestion
        Information describing how to aggregate jobs. Takes in a list of
        jobs and can return or yield lists or single jobs instead.
        The default behavior is creating an aggregate of all jobs.
```

flow/aggregate.py
```diff
@@ -0,0 +1,183 @@
...
+    :type aggregator:
+        callable
+    :param sort:
+        Before aggregating, sort the jobs given by a statepoint parameter.
+        The default value is None.
```
atravitz, line 36:
```suggestion
        Before aggregating, sort the jobs by a given state point parameter.
        The default value is None.
```

flow/aggregate.py
```diff
@@ -0,0 +1,183 @@
...
+    :type sort:
+        str or NoneType
+    :param reverse:
+        States if the jobs are to be sorted in reverse order.
+        The default value is False.
+    :type reverse:
+        bool
+    :param select:
+        Condition for filtering jobs. This operates on a single job.
```
atravitz, line 45:
```suggestion
        Condition for filtering jobs.
```

Why is it important to specify that this operates on a single job?

kidrahahjo: Actually the point I was trying to make here is the sequence in which we currently go.
1. select
2. sort
3. aggregate

The `select` parameter should not be something like
```
def _select(job1, job2):
    return job1 > job2
```

flow/aggregate.py
b-butler:
```suggestion
        Information describing how to aggregate jobs. Is a callable that
        takes in a list of jobs and can return or yield subsets of jobs as list
         or single jobs. The default behavior is creating a single aggregate
         of all jobs.
```
Also, I would recommend that the function should always return list or tuples of subsets of jobs. The yielded or returned value should be an iterable.

flow/aggregate.py
b-butler: Rather than state the default value, we should state the default behavior here. An user won't necessarily know what `None` means in this context (i.e. no sorting).

flow/aggregate.py
b-butler:
```suggestion
        Condition for filtering individual jobs. This is passed as the callable argument
        to `filter`.
```

flow/aggregate.py
```diff
@@ -0,0 +1,183 @@
...
+        The default value is None
```
b-butler, line 46: Here as well don't state the default value, state the default behavior.

flow/aggregate.py
```diff
@@ -0,0 +1,183 @@
...
+        callable or NoneType
+    """
+
+    def __init__(self, aggregator=None, sort=None, reverse=False, select=None):
+        if aggregator is None:
+            def aggregator(jobs):
+                return jobs
+
+        if not callable(aggregator):
+            raise TypeError("Expected callable aggregator function, got {}"
```
b-butler, line 57:
```suggestion
            raise TypeError("Expected callable for aggregator, got {}"
```

flow/aggregate.py
```diff
@@ -0,0 +1,183 @@
...
+                            "".format(type(aggregator)))
+
+        if sort is not None and not isinstance(sort, str):
+            raise TypeError("Expected string sort parameter, got {}".format(type(sort)))
+
+        if not isinstance(reverse, bool):
+            raise TypeError("Expected bool reverse parameter got {}".format(type(reverse)))
```
b-butler, line 64: Here I would be lenient we could just do `reverse = bool(reverse)`, so 0 or 1 would work.

flow/aggregate.py
```diff
@@ -0,0 +1,183 @@
...
+
+        if select is not None and not callable(select):
+            raise TypeError("Expected callable select parameter, got {}".format(type(select)))
```
b-butler, line 67:
```suggestion
            raise TypeError("Expected callable for select, got {}".format(type(select)))
```

flow/aggregate.py
```diff
@@ -0,0 +1,183 @@
...
+
+        elif isinstance(key, Iterable):
+            keys = list(key)
+
+            if default is None:
+                def keyfunction(job):
+                    return [job.sp[key] for key in keys]
+            else:
+                def keyfunction(job):
+                    return [job.sp.get(key, default) for key in keys]
```
b-butler, line 109: We should allow for `default` to be a list of `len(keys)` here.

flow/aggregate.py
```diff
@@ -0,0 +1,183 @@
...
+            for key, group in groupby(sorted(jobs, key=keyfunction), key=keyfunction):
+                yield group
+
+        return cls(aggregator, sort, reverse, select)
+
+    def __call__(self, obj=None):
+        if callable(obj):
+            setattr(obj, '_flow_aggregate', self)
+            return obj
+        elif isinstance(obj, (list, Project, JobsCursor)):
```
b-butler, line 128: This conditional makes me thing we should be dealing with two classes, one for registering the aggregates and another for performing the logic associated with them.

flow/aggregate.py
```diff
@@ -0,0 +1,183 @@
...
+            if not len(aggregated_jobs):
+                return []
+            for i, job in enumerate(aggregated_jobs[-1]):
+                if job is None:
+                    del aggregated_jobs[-1][i:]
+                    break
+            return aggregated_jobs
+        elif isinstance(obj, signac.contrib.job.Job):
+            # For backwards compatibility
+            return [obj]
```
b-butler, line 145: Why is this needed exactly? Ideally, we would prevent this from ever being called on by a single job internally.

flow/aggregate.py
```diff
@@ -0,0 +1,183 @@
...
+        else:
+            raise TypeError('Invalid argument passed while calling '
+                            'the aggregate instance. Expected a callable '
+                            'or an Iterable of Jobs, got {}.'.format(type(obj)))
+
+    def _create_nested_aggregate_list(self, aggregated_jobs):
```
b-butler, line 151: This function is complicated quite a bit by allowing for iterators that return single jobs. I think we should not allow that behavior.

kidrahahjo:
For now, I'm going to remove the review request from @b-butler @atravitz @csadorf because I'm converting this to a draft.
I will re-request my review after completing this PR

kidrahahjo:
This is being moved to #336

PR #312, opened 2020-06-30
Remove deprecated methods

kidrahahjo:
## Description

This PR removes the methods that were deprecated in flow version 0.9 and were to be removed in 0.11
## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -20,7 +20,7 @@ Removed
 +++++++
 
 - Drop support for Python 3.5 (#305). The signac project will follow the `NEP 29 deprecation policy <https://numpy.org/neps/nep-0029-deprecation_policy.html>`_ going forward.
-- Remove the command line option ``--cmd`` for ``script`` and deprecated methods ``always``, ``make_bundles``, and ``JobOperation.get_id`` (#312).
+- Remove the deprecated methods ``always``, ``make_bundles``, and ``JobOperation.get_id`` (#312).
 
 Version 0.10
 ============
--- a/changelog.txt
+++ b/changelog.txt
@@ -5,6 +5,14 @@ Changes
 The **signac-flow** package follows `semantic versioning <https://semver.org/>`_.
 The numbers in brackets denote the related GitHub issue and/or pull request.
 
+next
+----
+
+Removed
++++++++
+
+- Remove the deprecated methods ``always``, ``make_bundles``, and ``JobOperation.get_id`` (#312).
+
 Version 0.10
 ============
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -15,7 +15,7 @@ Removed
 +++++++
 
 - Drop support for Python 3.5 (#305). The signac project will follow the `NEP 29 deprecation policy <https://numpy.org/neps/nep-0029-deprecation_policy.html>`_ going forward.
-- Remove the deprecated methods ``always``, ``make_bundles``, and ``JobOperation.get_id`` (#312).
+- Remove the command line option ``--cmd`` for ``script`` and deprecated methods ``always``, ``make_bundles``, and ``JobOperation.get_id`` (#312).
 
 Version 0.10
 ============
--- a/flow/project.py
+++ b/flow/project.py
@@ -2448,14 +2448,6 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
             self.run_operations(operations, pretend=pretend,
                                 np=np, timeout=timeout, progress=progress)
 
-    def _generate_operations(self, cmd, jobs, requires=None):
-        "Generate job-operations for a given 'direct' command."
-        for job in jobs:
-            if requires and set(requires).difference(self.labels(job)):
-                continue
-            cmd_ = cmd.format(job=job)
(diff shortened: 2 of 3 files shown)


kidrahahjo:
https://github.com/glotzerlab/signac-flow/blob/master/flow/project.py#L173-L176

I'm not sure how to phrase this comment

vyasr:
@kidrahahjo can't you just rephrase to remove references to `always`?

kidrahahjo:
> @kidrahahjo can't you just rephrase to remove references to `always`?

The comment says

> the value should be used as the tag for that condition to ensure that **no pair of "always" and "never" conditions** are found to be equal by the graph detection algorithm

I can't remove the reference of always directly here (can I?)

vyasr:
This comment should be sufficient.

> the value should be used as the tag for that condition to ensure that no pair of "never" conditions are found to be equal by the graph detection algorithm

kidrahahjo:
> Can you remove the references to `always` in the docs?

@b-butler I'll raise another pull request for that

b-butler:
@kidrahahjo, that is good, but in the API docs (in this repo) there are also references to `always`.

kidrahahjo:
> @kidrahahjo, that is good, but in the API docs (in this repo) there are also references to `always`.

@b-butler oh okay, yes I'll edit. Sorry I didn't notice those.

changelog.txt
```diff
@@ -5,12 +5,16 @@ Changes
 The **signac-flow** package follows `semantic versioning <https://semver.org/>`_.
 The numbers in brackets denote the related GitHub issue and/or pull request.
 
+next
+====
+
 next
 ----
```
b-butler: Why do we have next twice now?

kidrahahjo:
@b-butler I forgot to remove a deprecated command-line option `--cmd`.
I have removed it and also made sure that we don't have more deprecated methods/arguments.

There is a problem that we released this deprecation in 0.10 and we state that we did it in 0.9. Please comment on this whether should this be removed or whether the removal of this argument should get extended?

flow/project.py
```diff
@@ -3435,28 +3385,16 @@ def _main_run(self, args):
...
         # Select jobs:
         jobs = self._select_jobs_from_args(args)
 
         # Gather all pending operations or generate them based on a direct command...
         with self._potentially_buffered():
-            if args.cmd:
-                warnings.warn("The --cmd option for script is deprecated as of "
-                              "0.9 and will be removed in 0.11.",
-                              DeprecationWarning)
-                operations = self._generate_operations(args.cmd, jobs, args.requires)
```
b-butler, line 3453: We should delete `FlowProject._generate_operations` as it is used no where else.

b-butler:
@kidrahahjo we can remove it now since we released on `flow` version 0.10. The `master` branch can be considered an 0.11 alpha (still adding features), which means we can remove items that are scheduled to remove in 0.11.

flow/project.py
kidrahahjo: @b-butler I have deleted this method. I think we can go ahead.

b-butler: Thanks for pointing that out. I missed it in my review.

kidrahahjo:
#243 states that the `--cmd` argument was deprecated in `flow` _version 0.9_ but I can guess by seeing the changelog that it is actually deprecated in `flow` _version 0.10_.
Generally, we remove the deprecated object after one release. But in this case, we'll be removing it right after version in which it was deprecated.

vyasr:
@kidrahahjo I see. You are correct. I would undo that deprecation, and instead change the deprecation warning to reflect the correct versions. Then we can merge this PR.

kidrahahjo:
@vyasr 
Instead of merging it in this PR. I'd prefer if I made another PR, make a minor bug-fix release, and then merge this PR.
Please suggest.

vyasr:
That's a good idea. That way we will be able to make a very small release that just contains this fix. When you make this change, make sure to branch off of the last release (tag v0.10.0) so that if we make a small bugfix release for this it doesn't contain any of the other changes that have been incorporated into master since we finalized 0.10.0. @bdice let me know if you are OK with this. We could in theory bump the minor version instead, but if we did that right now the release would contain nothing except a deprecation and dropping 3.5. I think we can hold off until a few more of the deprecations are merged, but since you're managing releases I'll let you make a final decision.

@kidrahahjo As you said, though, the method appears to have only been deprecated in 0.10, so you would still need to update this PR to not remove the `--cmd` argument.

bdice:
Personally, I’m okay with making an exception to the deprecation policy if a mistake was made in the deprecation notice. I am fine with adopting whichever (the changelog version or the deprecation decorator version) you all want to follow here. I don't think releasing a bugfix version for such a minor change is warranted. Feel free to change the deprecations however you think is appropriate in this case. I'm fine with rapid removal for minor functionality like this.

vyasr:
@kidrahahjo that was good work making that PR, but after discussing yesterday @bdice, @b-butler and I decided that this minor breaking of our deprecation policy wouldn't be worth the effort of making a bugfix release, so you can go ahead and just fix the deprecation warning here.

PR #315, opened 2020-06-30
Make FlowCondition class private

### LINKED ISSUE #307: Deprecate FlowCondition
The FlowCondition class is currently part of the API, but there is no user-facing method that requires it or returns it, nor is there any reason for there to be one.

Moreover, I don't think there is a strong reason to maintain the class at all. It is a very thin wrapper around a callable, and converting to and from the class adds unnecessary overhead. We should change the `_prereqs` and `_postconds` of `FlowOperation` to directly store and use the callables. If desired, we can add the error handling of the callable in places where it's useful (which I think would be more readable in any case).

kidrahahjo:
## Description

This pull request makes the `FlowCondition` class private.

## Motivation and Context

Since the `FlowCondition` class is not user-facing hence it needs to be private.
Resolves #307 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -11,7 +11,7 @@ next
 Changed
 +++++++
 
-- Deprecate the use of ``FlowCondition`` class (#307, #315).
+- Make ``FlowCondition`` class private (#307, #315).
 
 Version 0.10
 ============
--- a/changelog.txt
+++ b/changelog.txt
@@ -5,6 +5,14 @@ Changes
 The **signac-flow** package follows `semantic versioning <https://semver.org/>`_.
 The numbers in brackets denote the related GitHub issue and/or pull request.
 
+next
+----
+
+Changed
++++++++
+
+- Deprecate the use of ``FlowCondition`` class (#307, #315).
+
 Version 0.10
 ============
 
--- a/flow/project.py
+++ b/flow/project.py
@@ -449,8 +449,11 @@ class FlowCondition(object):
         callable
     """
 
+    @deprecated(deprecated_in="0.11", removed_in="0.13", current_version=__version__)
     def __init__(self, callback):
         self._callback = callback
+        warnings.warn("The FlowCondition class is deprecated as of 0.11 "
+                      "and will be removed in 0.13", DeprecationWarning)
 
     def __call__(self, job):
         if self._callback is None:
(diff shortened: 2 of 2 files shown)


vyasr:
Hi @kidrahahjo, sorry if we did a poor job explaining, but I don't think you quite understood the goal here. The deprecation itself is fine, since we do want to remove `FlowCondition` from public view. However, the internal refactoring that @b-butler and I suggested might be worthwhile was not to simply create some internal replacement the way you have done, but rather to remove that layer entirely, i.e. to make `BaseFlowOperation._prereqs` into a list of user-provided callables rather than a list of `FlowCondition` objects constructed from user provided callables. The question we wanted to ask was whether the `FlowCondition` class provided any benefit over just storing the raw callables directly. Currently, there is some additional overhead whenever we want to work directly with the callables; for instance, if you look at `FlowProject.detect_operation_graph`, it has to pull the callables out of each `FlowCondition` so that it can work with them directly. Currently that is the only place where these lists are used, so we haven't noticed more cumbersome issues elsewhere yet.

Having said that, @csadorf may be correct that this won't be more readable after the changes. I think we _can_ make it much more readable by storing the callables directly, but then we lose the custom error message, which I think is helpful enough to merit maintaining the small extra class. In an ideal world, I would like `FlowCondition` to be a completely transparent wrapper for callables that forwards all methods to the underlying function, but unfortunately I don't think that's possible since, for instance, `hash` bypasses `__getattribute__` entirely so you would still need some boilerplate to enforce the pass through. We could override `__getattr__` in FlowCondition to make it _mostly_ behave like this, but I'm not sure that's merited right now since we use the conditions in so few places. Unless one of the reviewers has a better idea or a different opinion here, I think for now we can reduce the scope of this PR to just deprecating the class from the public API (and removing the `if cond is None` block from `__call__` since we should just error on a null condition).

kidrahahjo:
@vyasr @csadorf I now understand what you meant. Yes, the changes I made don't look convincing over using the class `FlowCondition`.
As @csadorf  said, since the class is not "public" (user-facing) and only its name is not started with an underscore. So instead of deprecating it,  we could make this class private (or maybe I can just replace it with a callable).
The class handles the conditions well, although it adds another layer. For now, if not directly replacing it with a callable, we should make this private. I'll do as you guys suggest.

csadorf:
@vyasr Can you point out other spots at which this class creates additional overhead besides https://github.com/glotzerlab/signac-flow/blob/6fc1ad8e1a11910c2a19b897fec5483a89b8fe72/flow/project.py#L1464-L1466 ?

Because if that's the only spot, I'd suggest to just make the class private and otherwise close this issue.

vyasr:
@csadorf no, that's the only place. Sorry if I didn't make that clear in my previous comment. So yes, @kidrahahjo for now I recommend following Simon's suggestion to simply make the class private. Also remove the one superfluous error check that I mentioned.

b-butler:
Are we skipping the deprecation cycle on this since we don't expose `FlowCondition` objects in our API?

kidrahahjo:
@b-butler the class is not exposed to the users in any way. Hence I think there's no need of depreciating this class.

Issue #278, opened 2020-03-25
Convert all docstrings to numpy style


vyasr:
As part of our overall docs overhaul (see glotzerlab/signac-docs#64), we want to convert our docstrings to numpy style (as decided in glotzerlab/signac-docs#74). The best automated tool I'm familiar with for this task is [pyment](https://github.com/dadadel/pyment). In addition to converting docstrings, it will also generate docstrings for functions, classes, etc that are missing docstrings entirely. However, the conversion will require significant manual review to ensure that all docstrings are converted correctly.

vyasr:
@CryoZEUS could you comment on this issue so that I can assign you?

CryoZEUS:
Yeah, I'd love to work on it.

b-butler:
We should revisit this after GSoC. We will likely need someone else to move #316 to completion.

bdice:
Resolved by #392.

PR #316, opened 2020-07-04
Update docstrings to Numpy style 

CryoZEUS:
## Description

Converts module docstrings to numpy style.
## Motivation and Context

Issue #278 
## Types of Changes

- [x] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/labels.py
+++ b/flow/labels.py
@@ -11,16 +11,17 @@ of a FlowProject class definition.
 
 class label(object):
     """Decorate a :class:`~.FlowProject` class function as a label function.
-
+    
     For example:
-
+    
     .. code-block:: python
-
+    
         class MyProject(FlowProject):
-
+    
             @label()
             def foo(self, job):
                 return True
+
     """
 
     def __init__(self, name=None):
@@ -36,7 +37,15 @@ class label(object):
 class staticlabel(label):
     """A label decorator for staticmethods.
 
+    Parameters
+    ----------
+    label :
+
+    Notes
+    -----
+    
     This decorator implies "staticmethod"!
+
     """
 
     def __call__(self, func):
@@ -46,7 +55,15 @@ class staticlabel(label):
 class classlabel(label):
     """A label decorator for classmethods.
 
+    Parameters
+    ----------
+    label :
+
+    Notes
+    -----
+    
     This decorator implies "classmethod"!
+
     """
 
     def __call__(self, func):
@@ -54,4 +71,12 @@ class classlabel(label):
 
 
 def _is_label_func(func):
+    """Return func if func is a label
+
+    Parameters
+    ----------
+    func :
+        FlowProject class label function
+
+    """
     return getattr(getattr(func, '__func__', func), '_label', False)
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,6 +12,7 @@ python-tag = py3
 
 [pydocstyle]
 match = template_filters.py
+match = flow/labels.py
 
 [flake8]
 max-line-length = 100


flow/labels.py
```diff
@@ -36,7 +37,15 @@ def __call__(self, func):
 class staticlabel(label):
     """A label decorator for staticmethods.
 
+    Parameters
+    ----------
+    label :
```
tcmoore3, line 42: This needs to be filled in.

flow/labels.py
```diff
@@ -46,12 +55,28 @@ def __call__(self, func):
 class classlabel(label):
     """A label decorator for classmethods.
 
+    Parameters
+    ----------
+    label :
```
tcmoore3, line 60: This needs to be filled in.

flow/labels.py
```diff
@@ -36,7 +37,15 @@ def __call__(self, func):
 class staticlabel(label):
     """A label decorator for staticmethods.
 
+    Parameters
+    ----------
+    label :
+
+    Notes
+    -----
+
```
tcmoore3, line 46: I don't think there should be a blank line here.

flow/labels.py
```diff
@@ -46,12 +55,28 @@ def __call__(self, func):
...
     def __call__(self, func):
         return classmethod(super(classlabel, self).__call__(func))
 
 
 def _is_label_func(func):
+    """Return func if func is a label
+
+    Parameters
+    ----------
+    func :
```
tcmoore3, line 78: Type is `callable` I believe

flow/labels.py
```diff
@@ -11,16 +11,17 @@
 
 class label(object):
     """Decorate a :class:`~.FlowProject` class function as a label function.
-
+
```
bdice, line 14: @CryoZEUS Can remove these extra spaces on otherwise blank lines? I recommend configuring your editor to trim trailing whitespace when you save a file.

setup.cfg
```diff
@@ -12,6 +12,7 @@ python-tag = py3
 
 [pydocstyle]
 match = template_filters.py
+match = flow/labels.py
```
bdice, line 15: Please verify whether this overrides the previous "match" or if it checks both files. I'm not sure what behavior to expect from specifying it twice. In signac we used a complicated regular expression to match all desired filenames, examples below.

Initially we started with a list of allowed (files):
https://github.com/glotzerlab/signac/blob/e8209473103f562fe89c54bea1f5d1ba82b9632a/setup.cfg#L17-L19

Then later, once most of the code was compliant, we transitioned to a list of disallowed / skipped files:
https://github.com/glotzerlab/signac/blob/5186bb27621a1707b9d7493e14e551fe6426737d/setup.cfg#L17-L25

flow/labels.py
```diff
@@ -11,16 +11,17 @@
...
+    
     .. code-block:: python
-
+    
         class MyProject(FlowProject):
-
+    
             @label()
             def foo(self, job):
                 return True
```
bdice, line 23: This should be replaced by an Examples block with a doctest-style code snippet. Look at the "Examples" here: https://numpydoc.readthedocs.io/en/latest/format.html

Something like:
```
Examples
--------
>>> from flow import FlowProject, label
>>> class MyProject(FlowProject):
...     @label
...     def foo(self, job):
...         return True
```

vyasr:
> Left one more comment in addition to the previous reviews. Also, the numpy docstring guide suggests that `the length of docstring lines should be kept to 75 characters to facilitate reading the docstrings in text terminals.` Not sure if `flake8` can somehow moderate this.

It is possible to specify a maximum line length for [pydocstyle](https://pycodestyle.pycqa.org/en/latest/intro.html#configuration) distinct from the one we use for code. However, I'm not sure if there's much benefit to that; the reasoning in the numpy docstring guide matches the typical reasoning for keeping line lengths to a specific number, and aside from the practical annoyances with having two different line lengths I don't think it actually improves readability. I'll leave it to the reviewers to make a final call, though.

flow/labels.py
CryoZEUS: Okay, will do that

b-butler:
We should hold up on this PR until after GSoC, and then someone will need to take over pushing this PR to completion.

atravitz:
this is no longer blocked if someone wants to pick this up

bdice:
Replaced by #392.

PR #322, opened 2020-07-08
Update tested signac versions to 1.3.0 and 1.4.0.

bdice:
We aim to test the two most recent minor versions and one major version of signac with signac-flow. This updates from 1.0.0, 1.2.0, and 1.3.0 to 1.0.0, 1.3.0, and 1.4.0.

Merge this PR when CI passes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,14 +80,14 @@ jobs:
     <<: *test-template
     environment:
       SIGNAC_VERSION: "v1.0.0"
-  test-3.8-signac-1.2.0:
-    <<: *test-template
-    environment:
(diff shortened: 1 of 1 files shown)


csadorf:
@bdice Should we update our [signac dependency](https://github.com/glotzerlab/signac-flow/blob/master/setup.py#L10) as well then?

bdice:
> 
> 
> Should we update our [signac dependency](https://github.com/glotzerlab/signac-flow/blob/master/setup.py#L10) as well then?

I don't think we need to change it since nothing backwards-incompatible has been done here. We still test version 1.0.0, so that or any newer version will suffice.

Issue #313, opened 2020-06-30
Making JobOperation Private


b-butler:
The `JobOperation` class is currently exposed to users for the primary purpose of using with `submit_operations` and `run_operations` #310. The only way to obtain `JobOperation` objects through the user interface is through the `FlowProject.next_operations`. The use case for this class is small, so I purpose making `JobOperation` a private class. The current public methods of `JobOperation` involve the id, command string, and querying the status. None of these, I believe, to be a compelling use case for keeping the class public.

Making `JobOperation` private would also more easily allow future refactoring.

kidrahahjo:
@b-butler should I work on making `JobOperation`, `SubmissionJobOperation`, and `next_operations` private?

vyasr:
@kidrahahjo I would hold off on this task for now. I'd like to hear feedback from some more of @glotzerlab/signac-committers before pushing forward too quickly.

csadorf:
The original intention to expose some of this to the public API was to enable users to more easily develop their own execution and submission workflows, completely independent of the flow supported by the command line interface. I have so far not seen any real-world use cases for this. So with the goal to simplify things, I think it is totally fine to make this class (and all functions that directly use it) private.

kidrahahjo:
@csadorf @vyasr @b-butler 
Deprecation of these methods along with `JobOperation` class is scheduled for this issue to get resolved.

1. `next_operations`
2. `run_operations`
3. `submit_operations`
4. `script` (This is because it takes in list of `JobOperation` instances and we are deprecating it now so this should also be deprecated)
5. `completed_operations` (This is just lying in the API without any internal use, I'm not sure why is this so)

I also feel like there's a need for deprecation `get_job_status` because of change in the way we fetch status for aggregation. After that implementation, we won't be using it for status update.

PR #283, opened 2020-04-13
Add a class for Directives

kidrahahjo:
By - @b-butler 

Adds two classes `Directives` and `DirectivesItem` that serve as a smart mapping for the environment and user-specified directives and a specification for environmental directives respectively. `FlowProject` and `FlowGroup` have been changed accordingly.
## Motivation and Context

This resolves #265 and helps to centralize logic for directives.
This pull request is a necessary follow-up for #282.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change<sup>1</sup>

<sup>1</sup>The change breaks (or has the potential to break) existing functionality.

## Tasks to accomplish:
- [x] Test the individual DirectivesItems.
- [x] Test the Directives class.
- [x] Determine and add environment specific DirectivesItem and correct their get_default_directives function (this at least at the decision level will likely involve a discussion between multiple people).
- [x] Some probably large degree of code refactoring (I was just initially trying to get the outline working and there).
- [ ] Opening a PR in signac-docs to update documentation.
- [x] Going through docstrings and ensuring they are complete, grammatically correct, and helpful. Multiple methods will need these as well.
- [x] Add tracking of user specified directives. Previously TrackGetItemDict found in flow/util/misc.py was used. We can use this internally or create our own fix.
- [x] Fix code so tests pass

## Checklist:

- [x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac-flow/blob/master/ContributorAgreement.md).
- [x] My name is on the [list of contributors](https://github.com/glotzerlab/signac-flow/blob/master/contributors.yaml).
- [x] My code follows the [code style guideline](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md#code-style) of this project.
- [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.

If necessary:
- [ ] I have updated the API documentation as part of the package doc-strings.
- [ ] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
- [x] I have updated the [changelog](https://github.com/glotzerlab/signac-flow/blob/master/changelog.txt).

PR #324, opened 2020-07-16
Enable aggregate logic in flow

kidrahahjo:
Enabling aggregate logic in `flow` (enabling internal handling of lists) without any change in the API.

## Description

The original PR for enabling aggregate operations #289  is too large.
This PR reduces the scope of that PR by handling only aggregates of 1.

## Motivation and Context

Suggested by @glotzerlab/signac-gsoc-mentors 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


kidrahahjo:
The tests passed locally and will hopefully pass here too.
Given this, I think we'd want to deprecate the `get_job_status` method or maybe refactor it. Internally `get_job_status` is not used now, after this implementation, this method will just float in the API.
Also, I think that this is ready for review.

flow/project.py
```diff
@@ -491,13 +516,13 @@ def __init__(self, pre=None, post=None):
...
 
-    def eligible(self, job, ignore_conditions=IgnoreConditions.NONE):
+    def eligible(self, *jobs, ignore_conditions=IgnoreConditions.NONE):
         """Eligible, when all pre-conditions are true and at least one post-condition is false,
         or corresponding conditions are ignored.
-        :param job:
-            The signac job handles.
-        :type job:
-            :class:`~signac.contrib.job.Job`
+        :param jobs:
```
bdice, line 522: The parameter is actually `*jobs`, not `jobs`. However, if you just add an asterisk, Sphinx will interpret the asterisk as "beginning bold text." This is annoying, but you can fix it by changing this to a raw string (`r"""`) and escaping the asterisk like this:

```
r"""Docstring header.

:param \*jobs:
    A list of signac job handles.
:type \*jobs:
    list of :class:`~signac.contrib.job.Job`
"""
```
Please use that pattern in all docstrings accepting `*jobs`.

flow/project.py
```diff
@@ -334,6 +353,10 @@ def get_id(self):
     def id(self):
         return self._id
 
+    @property
+    def job(self):
+        return self.jobs[0]
```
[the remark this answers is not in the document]
kidrahahjo: This is there only for backward compatibility. We might have to warn users to use `jobs` instead.

flow/aggregate.py
```diff
@@ -0,0 +1,75 @@
...
+
+        self._aggregator = aggregator
+        self._select = None if select is None else partial(filter, select)
+
+    @classmethod
+    def groupsof(cls, num=1, select=None):
+        # copied from: https://docs.python.org/3/library/itertools.html#itertools.zip_longest
+        try:
+            num = int(num)
+            if num < 0:
```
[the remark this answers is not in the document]
kidrahahjo: It's not supposed to be 0. I will change the condition

flow/project.py
```diff
@@ -349,12 +372,12 @@ def cmd(self):
 
     def set_status(self, value):
         "Store the operation's status."
-        self.job._project.document.setdefault('_status', dict())[self.id] = int(value)
+        self.jobs[0]._project.document.setdefault('_status', dict())[self.id] = int(value)
```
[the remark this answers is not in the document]
kidrahahjo: For now (if we allow only one job in an aggregate) I think it's complete. If you wan't I can start implementing the behaviour we discussion (storing the ids for status check)

flow/project.py
```diff
@@ -310,13 +325,17 @@ def __init__(self, id, name, job, cmd, directives=None):
         self.directives._keys_set_by_user = keys_set_by_user
 
     def __str__(self):
-        return "{}({})".format(self.name, self.job)
+        if len(self.jobs) == 1:
+            return "{}({})".format(self.name, str(self.jobs[0]))
+        else:
+            return "{}-{}: {}...{})".format(
+                self.name, len(self.jobs), str(self.jobs[0])[0:8], str(self.jobs[-1])[0:8])
```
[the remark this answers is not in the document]
kidrahahjo: Thanks for the link

flow/project.py
```diff
@@ -201,8 +212,11 @@ def never(cls, func):
 
     @classmethod
     def not_(cls, condition):
-        "Returns ``not condition(job)`` for the provided condition function."
-        return cls(lambda job: not condition(job),
+        """Evaluates to True if the provided condition function returns
+        False for all the jobs in the aggregate passed at once."""
```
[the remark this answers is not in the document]
kidrahahjo: Actually this is a documentation confusion.

>    Evaluates to True if the provided condition function returns
False when the jobs in the aggregate passed to the condition function.

I'll edit the doc string

flow/project.py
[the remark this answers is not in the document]
bdice: I think it's fine to leave this for the current PR since it's 100% backwards compatible. We just need to track this carefully so that we change it in a future PR.

flow/project.py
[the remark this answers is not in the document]
bdice: Even though this is for backwards compatibility, we might avoid user errors like this:
```suggestion
        if len(self.jobs) == 1:
            return self.jobs[0]
        else:
            raise AttributeError("This object holds multiple jobs. Use the jobs attribute instead.")
```

flow/project.py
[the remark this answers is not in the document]
kidrahahjo: Yes, I'll definitely track this carefully for future work.

flow/project.py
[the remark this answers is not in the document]
kidrahahjo: @bdice I think we should also raise a deprecation warning here. I'll make the changes after your approval for the warning.

flow/aggregate.py
```diff
@@ -0,0 +1,81 @@
...
+
+    :param _aggregate:
+        Aggregate object associated with an operation function
+    :type _aggregate:
+        :py:class:`Aggregate`
+    """
+    def __init__(self, _aggregate=Aggregate()):
+        self._aggregate = _aggregate
+
+    def __call__(self, obj):
```
b-butler, line 70: can we be more specific than the `obj` parameter.

flow/aggregate.py
```diff
@@ -0,0 +1,81 @@
...
+        "Return aggregated jobs"
+        aggregated_jobs = list(obj)
+        aggregated_jobs = self._aggregate._aggregator([job for job in aggregated_jobs])
+        aggregated_jobs = [[job for job in aggregate] for aggregate in aggregated_jobs]
+        if not len(aggregated_jobs):
+            return []
```
b-butler, line 76:
```suggestion
        if len(aggregated_jobs) == 0:
            return []
```

This is a bit more explicit.

flow/project.py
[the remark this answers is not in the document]
b-butler: @bdice this PR is designed to only change the internals. I would vote to keep this property for now, and then we can decide on what to do with this class later. We have an issue #313 about making this class private in general.

flow/project.py
```diff
@@ -282,10 +298,10 @@ class JobOperation(object):
         :class:`dict`
     """
 
-    def __init__(self, id, name, job, cmd, directives=None):
+    def __init__(self, id, name, jobs, cmd, directives=None):
         self._id = id
         self.name = name
-        self.job = job
+        self.jobs = jobs
```
b-butler, line 304: Make this private for now.

b-butler: Also, since we internally use this class check if the `job` parameter is a list, so we can store it appropriately. I know this is a bit more of a hassle for now.

flow/project.py
```diff
@@ -310,13 +326,19 @@ def __init__(self, id, name, job, cmd, directives=None):
         self.directives._keys_set_by_user = keys_set_by_user
 
     def __str__(self):
-        return "{}({})".format(self.name, self.job)
+        if len(self.jobs) == 1:
+            return "{}({})".format(self.name, str(self.jobs[0]))
+        else:
+            minimum = self.jobs[0]._project.min_len_unique_id()
+            return "{}-{}: {}-{})".format(
```
b-butler, line 333:
```suggestion
            return "{}-{}: ({}-{})".format(
```

flow/project.py
```diff
@@ -310,13 +326,19 @@ def __init__(self, id, name, job, cmd, directives=None):
...
 
     def __repr__(self):
-        return "{type}(name='{name}', job='{job}', cmd={cmd}, directives={directives})".format(
+        return "{type}(name='{name}', jobs='{jobs}', cmd={cmd}, directives={directives})".format(
                    type=type(self).__name__,
                    name=self.name,
-                   job=str(self.job),
+                   jobs="[" + " ,".join(map(str, self.jobs)) + "]",
                    cmd=repr(self.cmd),
                    directives=self.directives)
```
b-butler, line 343: This is technically API breaking but that is fine for now.

flow/project.py
```diff
@@ -491,13 +519,14 @@ def __init__(self, pre=None, post=None):
     def __str__(self):
         return "{type}(cmd='{cmd}')".format(type=type(self).__name__, cmd=self._cmd)
 
-    def eligible(self, job, ignore_conditions=IgnoreConditions.NONE):
-        """Eligible, when all pre-conditions are true and at least one post-condition is false,
+    def eligible(self, *jobs, ignore_conditions=IgnoreConditions.NONE):
```
b-butler, line 522: This does change the API of a user facing class and method. In this case, I am not sure what to do, but it is worth noting.

flow/project.py
```diff
@@ -594,10 +624,15 @@ class FlowGroupEntry(object):
         commands to execute.
     :type options:
         str
+    :param aggregate:
+        Aggregator associated with the :py:class:`FlowGroup` to be created.
+    :type aggregate:
+        :py:class`_MakeAggregate`
```
b-butler, line 630: If this is only changing the internals, then this should not be exposed to the user yet.

flow/project.py
```diff
@@ -653,7 +688,7 @@ class FlowGroup(object):
     """A FlowGroup represents a subset of a workflow for a project.
 
     Any :py:class:`FlowGroup` is associated with one or more instances of
-    :py:class:`BaseFlowOperation`.
+    :py:class:`BaseFlowOperation` having identical aggregations.
```
b-butler, line 691: You can change the internal docs accordingly, but we shouldn't change the external docs to include aggregation.

flow/project.py
```diff
@@ -698,12 +733,16 @@ def op2(job):
         This lets options like ``--num_passes`` to be given to a group.
     :type options:
         str
+    :param aggregate:
+        Aggregate object associated with the group.
+    :type aggregate:
+        :py:class`_MakeAggregate`
```
b-butler, line 739: No user facing yet.

flow/project.py
```diff
@@ -761,32 +802,34 @@ def np_callable(job):
...
                 return value
         return {key: evaluate(value) for key, value in directives.items()}
 
-    def _submit_cmd(self, entrypoint, ignore_conditions, job=None):
-        entrypoint = self._determine_entrypoint(entrypoint, dict(), job)
+    def _submit_cmd(self, entrypoint, ignore_conditions, jobs=None):
+        entrypoint = self._determine_entrypoint(entrypoint, dict(), jobs)
         cmd = "{} run -o {}".format(entrypoint, self.name)
-        cmd = cmd if job is None else cmd + ' -j {}'.format(job)
+        cmd = cmd if jobs is None else cmd + ' -j {}'.format(' '.join(map(str, jobs)))
```
b-butler, line 818: This will work, but we may want to think on this before the finish aggregation, as this could lead to a huge command line command depending on number of jobs.

flow/project.py
```diff
@@ -800,13 +843,13 @@ def __repr__(self):
...
+        r"""Eligible, when at least one BaseFlowOperation is eligible.
 
-        :param job:
-            A :class:`signac.Job` from the signac workspace.
-        :type job:
-            :class:`signac.Job`
+        :param \*jobs:
+            The list of jobs that the :class:`~.JobOperation` is based on.
+        :type \*jobs:
+            list of :class:`signac.Job`
```
b-butler, line 852: Is user facing, and the API changes. Once again, this may be unavoidable but noting it.

flow/project.py
```diff
@@ -817,22 +860,22 @@ def eligible(self, job, ignore_conditions=IgnoreConditions.NONE):
...
+        r"""True when all BaseFlowOperation post-conditions are met.
 
-        :param job:
-            A :class:`signac.Job` from the signac workspace.
-        :type job:
-            :class:`signac.Job`
+        :param \*jobs:
+            The list of jobs that the :class:`~.JobOperation` is based on.
+        :type \*jobs:
+            list of :class:`signac.Job`
```
b-butler, line 871: docs don't need to change yet.

flow/project.py
```diff
@@ -854,6 +897,11 @@ def add_operation(self, name, operation, directives=None):
         if directives is not None:
             self.operation_directives[name] = directives
 
+    def _add_aggregate(self, aggregate):
+        """Add :py:class:`_MakeAggregate` object associated to the FlowGroup"""
+        if self.aggregate is None:
+            self.aggregate = aggregate
+
```
b-butler, line 904: Should we just do this at the creation of the `FlowGroup`?

flow/project.py
```diff
@@ -869,9 +917,12 @@ def isdisjoint(self, group):
...
         return set(self).isdisjoint(set(group))
 
-    def _generate_id(self, job, operation_name=None, index=0):
+    def _generate_id(self, jobs, operation_name=None, index=0):
         "Return an id, which identifies this group with respect to this job."
-        project = job._project
+        project = jobs[0]._project
+
+        for job in jobs:
+            assert job._project == project
```
b-butler, line 925: This might be worth being a static method of the class. Something like `_verify_aggregate`.

b-butler: We only do it here and check in other places that require it through `_generate_id`.

flow/project.py
```diff
@@ -895,25 +946,31 @@ def _generate_id(self, job, operation_name=None, index=0):
...
+
         separator = getattr(project._environment, 'JOB_ID_SEPARATOR', '/')
-        readable_name = '{project}{sep}{job}{sep}{op_string}{sep}{index:04d}{sep}'.format(
-                    sep=separator,
-                    project=str(project)[:12],
-                    job=str(job)[:8],
-                    op_string=op_string[:12],
-                    index=index)[:max_len]
+        readable_name = '{project}{sep}{jobs}{sep}{op_string}' \
+                        '{sep}{index:04d}{sep}'.format(
```
b-butler, line 956: Is there a reason this is split into two lines now?

flow/project.py
```diff
@@ -955,7 +1012,12 @@ def _create_submission_job_operation(self, entrypoint, default_directives, job,
         :rtype:
             :py:class:`SubmissionJobOperation`
         """
-        uneval_cmd = functools.partial(self._submit_cmd, entrypoint=entrypoint, job=job,
+        if isinstance(jobs, signac.contrib.job.Job):
+            jobs = [jobs]
```
b-butler, line 1016: As this is an internal function, we should never have this branch of the conditional be true, if we do that is an inconsistency in the internals.

flow/project.py
```diff
@@ -955,7 +1012,12 @@ def _create_submission_job_operation(self, entrypoint, default_directives, job,
...
+        else:
+            jobs = list(jobs)
```
b-butler, line 1018: If it is already supposed to be a list, why do we convert to a list here?

flow/project.py
```diff
@@ -1020,27 +1082,38 @@ def _create_run_job_operations(self, entrypoint, default_directives, job,
         :rtype:
             Iterator[JobOperation]
         """
+        if isinstance(jobs, signac.contrib.job.Job):
+            jobs = [jobs]
+        else:
+            jobs = list(jobs)
```
b-butler, line 1088: See comment on `_create_submission_job_operations`.

flow/project.py
```diff
@@ -1020,27 +1082,38 @@ def _create_run_job_operations(self, entrypoint, default_directives, job,
...
+
+        project = jobs[0]._project
+        for job in jobs:
+            assert project == job._project
```
b-butler, line 1092: Centralizing this would be helpful.

flow/project.py
```diff
@@ -1020,27 +1082,38 @@ def _create_run_job_operations(self, entrypoint, default_directives, job,
...
+
+        aggregated_jobs = self.aggregate(jobs)
```
b-butler, line 1094: Aggregate should be internal not public.

flow/project.py
```diff
@@ -1020,27 +1082,38 @@ def _create_run_job_operations(self, entrypoint, default_directives, job,
...
-    def _get_submission_directives(self, default_directives, job):
+            for aggregate in aggregated_jobs:
+                if op.eligible(*aggregate, ignore_conditions=ignore_conditions):
+                    directives = self._resolve_directives(name, default_directives, aggregate)
+                    cmd = self._run_cmd(entrypoint=entrypoint, operation_name=name,
+                                        operation=op, directives=directives, jobs=aggregate)
+                    # Uses a different id than the groups direct id. Do not use this for submitting
+                    # jobs as current implementation prevents checking for resubmission in this
+                    # case. The different ids allow for checking whether JobOperations created to
+                    # run directly are different.
```
b-butler, line 1104: This isn't related, but I don't think this is true any more.

flow/project.py
kidrahahjo: This logic is to support multiple aggregations associated with an operation (if it also there in some group).
This is not necessary for now, but I think this is the right way to do it)

flow/project.py
kidrahahjo: Will insert an assert statement

flow/project.py
kidrahahjo: I changed the way we printed `id` but then realized that it may not be something we should change in this PR. Forgot to change it back to the original state, will do it now.
Thank you for pointing out

flow/project.py
kidrahahjo: This means then I'd have to edit the tests.

flow/project.py
kidrahahjo: Should I make a separate PR for this?

flow/project.py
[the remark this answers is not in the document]
kidrahahjo: @b-butler I personally think a custom error will be helpful

kidrahahjo:
I have optimized the way we fetch status, now it's much more efficient than the previous implementation.
Given this, I think now the tests will pass.
I have not resolved the conversations which I feel would be helpful in future to keep a track of the changes we introduced.

kidrahahjo:
@glotzerlab/signac-gsoc-mentors 
This is strange, I'm not even sure why the `test_status_performance` is passing in the first place if it's failing for signac-1.0.0

kidrahahjo:
@b-butler @bdice @csadorf @atravitz 
I have now removed the use of aggregate class and ensured that we don't make any change in the API.
Please have a look once you get time.

flow/project.py
```diff
@@ -310,13 +324,19 @@ def __init__(self, id, name, job, cmd, directives=None):
...
+                self.name, len(self._jobs), str(self._jobs[0])[0:minimum],
+                str(self._jobs[-1])[0:minimum])
 
     def __repr__(self):
-        return "{type}(name='{name}', job='{job}', cmd={cmd}, directives={directives})".format(
+        return "{type}(name='{name}', jobs='{jobs}', cmd={cmd}, directives={directives})".format(
                    type=type(self).__name__,
                    name=self.name,
-                   job=str(self.job),
+                   jobs="[" + " ,".join(map(str, self._jobs)) + "]",
```
b-butler, line 339: This is also changing behavior, it is fine since this will be mostly internal, just marking it, since we will need to change it for `JobOperation` not `_JobOperation` per #325

flow/project.py
```diff
@@ -334,6 +354,10 @@ def get_id(self):
     def id(self):
         return self._id
 
+    @property
+    def job(self):
+        return self._jobs[0]
+
```
b-butler, line 360: This will only need to be implemented on `JobOperation` in #325

flow/project.py
```diff
@@ -454,7 +479,7 @@ class BaseFlowOperation(object):
 
     Pre-requirements (pre) and post-conditions (post) can be used to
     trigger an operation only when certain conditions are met. Conditions are unary
-    callables, which expect an instance of job as their first and only positional
+    callables, which expects an instance of job as their first and only positional
```
b-butler, line 482: This is incorrect.

flow/project.py
```diff
@@ -549,7 +575,7 @@ def __init__(self, cmd, pre=None, post=None):
     def __str__(self):
         return "{type}(cmd='{cmd}')".format(type=type(self).__name__, cmd=self._cmd)
 
-    def __call__(self, job=None):
+    def __call__(self, job):
```
b-butler, line 578: This is technically a breaking change. I don't think we use `job=None` anywhere internally though. We do need to change this to work with list of jobs though. Maybe just use `(self, *jobs)`. This allows no jobs to be passed. Then we just do `return self._cmd(*jobs).format(*jobs)`. If we want to catch `None` then we could check the list length.

b-butler: This also may be a good reason to deprecate the call method of `FlowOperation` so as to not need to keep the signature valid between versions.

flow/project.py
```diff
@@ -1023,27 +1059,30 @@ def _create_run_job_operations(self, entrypoint, default_directives, job,
...
-                # to True since we must launch a separate process. Override
-                # the command directly.
-                prefix = job._project._environment.get_prefix(job_op)
-                if prefix != '':
-                    job_op.directives['fork'] = True
-                    job_op._cmd = '{} {}'.format(prefix, job_op.cmd)
-                yield job_op
-
-    def _get_submission_directives(self, default_directives, job):
+        aggregated_jobs = [[job] for job in jobs]
```
b-butler, line 1062: This function should take one aggregate not create them.

flow/project.py
b-butler: It may be worth doing that, and we could just merge right away. No need for a changelog even.

flow/project.py
```diff
@@ -2440,10 +2479,9 @@ def select(operation):
                 with self._potentially_buffered():
                     operations = []
                     for flow_group in flow_groups:
-                        for job in jobs:
-                            operations.extend(
-                                flow_group._create_run_job_operations(
-                                    self._entrypoint, default_directives, job, ignore_conditions))
```
b-butler, line 2446: This for loop should still exist, it should just yield lists of jobs.

flow/project.py
```diff
@@ -2510,14 +2548,16 @@ def _get_submission_operations(self, jobs, default_directives, names=None,
                                    ignore_conditions=IgnoreConditions.NONE,
                                    ignore_conditions_on_execution=IgnoreConditions.NONE):
         """Grabs JobOperations that are eligible to run from FlowGroups."""
-        for job in jobs:
-            for group in self._gather_flow_groups(names):
-                if group.eligible(job, ignore_conditions) and self._eligible_for_submission(group,
-                                                                                            job):
+        for group in self._gather_flow_groups(names):
+            for job in jobs:
```
b-butler, line 2552: `job` should be a list of jobs here.

flow/project.py
```diff
@@ -3153,17 +3200,17 @@ def _job_operations(self, job, ignore_conditions=IgnoreConditions.NONE):
...
 
     def next_operations(self, *jobs, ignore_conditions=IgnoreConditions.NONE):
-        """Determine the next eligible operations for jobs.
+        r"""Determine the next eligible operations for jobs.
 
-        :param jobs:
+        :param \*jobs:
             The signac job handles.
-        :type job:
+        :type \*jobs:
```
b-butler, line 3213: This still has to have the same behavior as before since it is user facing. The changes will exist though in private method that takes over this functionality #325.

tests/test_project.py
```diff
@@ -737,7 +737,7 @@ def test_pending_operations_order(self):
         project = self.mock_project()
         ops = list(project._get_pending_operations(self.project.find_jobs()))
         # The length of the list of operations grouped by job is equal
-        # to the length of its set if and only if the operations are grouped
+        # the length of its set if and only if the operations are grouped
```
b-butler, line 740: This sentence does not make sense any more.

flow/project.py
kidrahahjo: The `aggregate` parameter, when associated with a group, will have that behaviour. 
But for now, I think passing [job] in should be enough.

kidrahahjo:
@b-butler 
I think I am fine with deprecating `eligible` and `complete` methods of `Flow{Operation, Group}` depending on the real-world use case. But I'd prefer if that method remains public and we alert the users that we will change the functionality of this method in the upcoming release.
Currently, we're sure that there will be only one argument passed to `*jobs`. The reason I kept it that way is that if we used `*jobs` instead of job then to specify the `ignore_condition`, users would have pass `ignore_condition` as a keyword argument (which I think they won't be doing like that). Hence, we should definitely raise a `UserWarning` that this method will accept multiple jobs not and in order to specify any `ignore_condition`, the users would have to pass it as a `keyword argument`.

I think we should deprecate the use of `get_job_status` (Hence resulting in the removal of `_get_operation_status` with the removal of `get_job_status`) in this PR because I am not using `get_job_status` to fetch the status of an aggregate.

I think after #325 gets merged, and refactoring this a little, we can merge this without any issue.

flow/project.py
bdice: Addressed in #327. I will merge that PR.

b-butler:
@kidrahahjo if we deprecated, we would go through the normal cycle. The more important thing, I think for this PR is to have these functions work with lists of jobs. We could use duck typing, but in most other places in our API we are using the `*args` syntax which we cannot use here. That is why I suggested the internal methods that handle lists of jobs that `eligible` and `complete` call to under the hood so to speak.

flow/project.py
```diff
@@ -2476,7 +2609,7 @@ def select(operation):
                         for job in jobs:
                             operations.extend(
                                 flow_group._create_run_job_operations(
-                                    self._entrypoint, default_directives, job, ignore_conditions))
+                                    self._entrypoint, default_directives, [job], ignore_conditions))
```
atravitz, line 2612: I dont understand why we need to pass [job] in as a list? Would this be better if handled in `_create_run_job_operations()`?

kidrahahjo: The plan is to pass in a single aggregate (which is a list of jobs) to every internal method. Since we're currently dealing with aggregates of 1 in this PR and all the internal methods (of `FlowGroup`) now take a list of jobs, I thought that to maintain consistency, we should to pass `[job]` instead of `job`.

flow/project.py
```diff
@@ -334,6 +354,12 @@ def get_id(self):
     def id(self):
         return self._id
 
+    @property
+    def job(self):
+        if(len(self._jobs) > 1):
+            raise NotImplementedError
+        return self._jobs[0]
```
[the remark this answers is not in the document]
kidrahahjo: This is meant to be only implemented in `JobOperation` and not `_JobOperation`. I thought if I implemented it in `_JobOperation` then the test wouldn't have to be changed.
As suggested by @b-butler in a previous comment (which I accidentally forgot to address) I should remove this from `_JobOperation` and implement it only in `JobOperation` (Where we only take a single job as an argument).

flow/project.py
```diff
@@ -415,13 +512,14 @@ class _FlowCondition(object):
     def __init__(self, callback):
         self._callback = callback
 
-    def __call__(self, job):
+    def __call__(self, *jobs):
```
b-butler, line 515: This is internal now, so do we need to use `*jobs` here or can we just use the list?

kidrahahjo: A list will be fine.
I'll change this behaviour

b-butler, line 515: This is not user facing, we should use `jobs`.

kidrahahjo:
@b-butler We will create all aggregates right below where we loop over groups.
For example:
```
for flow_group in flow_groups:
   aggregated_jobs = flow_group.aggregate(jobs)
    for aggregate in aggregated_jobs:
       operations.extend(
           flow_group._create_run_job_operations(
               self._entrypoint, default_directives, aggregate, ignore_conditions))
```

Since the `-j` argument, when passed, specifies the jobs for the `jobs` parameter (which is usually fetched using the `_select_jobs_from_args` method). I think we definitely need to warn the users about the new behaviour because every time we `run`/`submit`, we'd pass `jobs` to the aggregator from that particular argument only. But I think this should be addressed in [signac-docs](github.com/glotzerab/signac-docs) while documenting aggregation.

flow/project.py
```diff
@@ -3509,9 +3686,9 @@ def _main_status(self, args):
 
     def _main_next(self, args):
         "Determine the jobs that are eligible for a specific operation."
-        for job in self:
-            if args.name in {op.name for op in self._next_operations(job)}:
-                print(job)
+        for op in self._next_operations(*self):
```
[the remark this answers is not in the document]
kidrahahjo: I think this is related to this PR in some manner. I'll change this in this PR only.
Also, yes, `_next_operation` is used in `_get_pending_operations` and `_main_next`. I think we should keep this method as for the CLA `next` it'll show all the aggregates which are eligible for an operation.

flow/project.py
kidrahahjo: Noted. Thanks for the suggestion.

flow/project.py
```diff
@@ -881,25 +1051,30 @@ def _generate_id(self, job, operation_name=None, index=0):
         if max_len < len(job_op_id):
```
bdice, line 1051: I am having a bit of difficulty determining if this code will work correctly with large aggregates. Please make sure to check that this code behaves as you expect for large aggregates. The part I'm confused about is the section involving `MAX_LEN_ID` and how it interacts with the parts above and below (whether we have the appropriate guarantees on uniqueness of scheduler ids).

flow/project.py
```diff
@@ -993,10 +1168,10 @@ def _create_run_job_operations(self, entrypoint, default_directives, job,
...
         :type default_directives:
             dict
-        :param job:
-            The job that the :class:`~._JobOperation` is based on.
-        :type job:
-            :class:`signac.Job`
+        :param jobs:
+            The list of jobs that the :class:`~._JobOperation` is based on.
+        :type jobs:
+            list of :class:`signac.Job`
```
bdice, line 1174: `signac.Job` is not the correct namespace of the class. It should be `signac.contrib.job.Job`. I think this is a mistake that has spread through the docstrings, so you may have copied it from somewhere else that was also wrong. Please fix this everywhere it occurs in this PR (this is the first case I've caught).
```suggestion
            list of :class:`signac.contrib.job.Job`
```

flow/project.py
```diff
@@ -3600,11 +3785,8 @@ def operation_function(job):
         except KeyError:
             raise KeyError("Unknown operation '{}'.".format(args.operation))
 
-        if getattr(operation_function, '_flow_aggregate', False):
```
bdice, line 3603: What is this replacing? I'm not sure I know what the old code was meant for.

flow/project.py
```diff
@@ -192,8 +200,10 @@ def never(cls, func):
 
     @classmethod
     def not_(cls, condition):
-        "Returns ``not condition(job)`` for the provided condition function."
-        return cls(lambda job: not condition(job),
+        """Returns ``not condition(job)`` for the provided condition function."""
```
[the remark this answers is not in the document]
kidrahahjo: @bdice this is a user-facing function and we decided not to change any use facing method's documentation/behaviour in this PR.

flow/project.py
```diff
@@ -415,13 +512,14 @@ class _FlowCondition(object):
...
-    def __call__(self, job):
+    def __call__(self, jobs):
         try:
-            return self._callback(job)
+            return self._callback(*jobs)
         except Exception as e:
             raise UserConditionError(
                 'An exception was raised while evaluating the condition {name} '
-                'for job {job}.'.format(name=self._callback.__name__, job=job)) from e
+                'for job {jobs}.'.format(name=self._callback.__name__,
```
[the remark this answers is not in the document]
kidrahahjo: This is a user-facing error message and should be edited in #306

flow/project.py
kidrahahjo: I too wasn't sure how to go about that because I didn't find any code that sets the `_flow_aggregate` attribute for an operation.

flow/project.py
```diff
@@ -700,42 +829,43 @@ def __init__(self, name, operations=None, operation_directives=None,
...
         nranks = directives.get('nranks', 1)
         nthreads = directives.get('omp_num_threads', 1)
         if callable(nranks) or callable(nthreads):
-            def np_callable(job):
-                nr = nranks(job) if callable(nranks) else nranks
-                nt = nthreads(job) if callable(nthreads) else nthreads
+            def np_callable(*jobs):
+                nr = nranks(*jobs) if callable(nranks) else nranks
+                nt = nthreads(*jobs) if callable(nthreads) else nthreads
                 return nr*nt
```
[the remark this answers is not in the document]
kidrahahjo: `nranks` as defined [here](https://github.com/glotzerlab/signac-flow/pull/283/files#diff-dd7774bb5ebcb1df3843a90e78c026a7R257) is _The number of MPI ranks to use for this operation._
I thought that it'd always return an Integer (Given any number of jobs we pass if it's a callable)

flow/project.py
kidrahahjo: I thought `job_op_id` is of constant length for any number of jobs in aggregate.
We calculate the job_op_id in this way
```
    full_name = '{}%{}%{}%{}'.format(project.root_directory(),
                                     '-'.join(map(str, jobs)),
                                     op_string,
                                     index)
    job_op_id = calc_id(full_name)
```
But still, I'll check and let you know the results.

flow/project.py
```diff
@@ -263,10 +273,13 @@ class _JobOperation(object):
...
 
-    def __init__(self, id, name, job, cmd, directives=None):
+    def __init__(self, id, name, jobs, cmd, directives=None):
         self._id = id
         self.name = name
-        self.job = job
+
+        assert isinstance(jobs, list)
+
+        self._jobs = jobs
```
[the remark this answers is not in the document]
kidrahahjo: A user will be able to return aggregates or yield them in any type. We internally convert those generators/Iterables to "nested list"
This was discussed earlier too but for now, I think consistency should be the key. For example, internally, if we use lists somewhere then we'll be using list throughout the code.
If @b-butler agrees, I'll be happy to change lists to tuples.

flow/project.py
```diff
@@ -291,13 +304,19 @@ def __init__(self, id, name, job, cmd, directives=None):
...
 
     def __str__(self):
-        return "{}({})".format(self.name, self.job)
+        if len(self._jobs) == 1:
+            return "{}({})".format(self.name, str(self._jobs[0]))
+        else:
+            minimum = self._jobs[0]._project.min_len_unique_id()
+            return "{}-{}: ({}-{})".format(
+                self.name, len(self._jobs), str(self._jobs[0])[0:minimum],
+                str(self._jobs[-1])[0:minimum])
```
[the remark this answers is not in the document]
kidrahahjo: Yes, I agree that this isn't unique.
Should we also provide `_JobOperation.id` so that users could look up in the `.aggregates/` directory to get details of specific aggregate for an operation?

flow/project.py
kidrahahjo: We'll be storing the submission id of the aggregate in the `.aggregates/` directory which is not introduced yet but will be introduced in [#MyFork/2](github.com/kidrahahjo/signac-flow/pull/2).
I think that will be enough for a meta-data?

flow/project.py
kidrahahjo: Moreover, I think this should be addressed in #306 as we're not yet dealing with Aggregates in this PR.

kidrahahjo: It works fine for an aggregate of length 100k.

flow/project.py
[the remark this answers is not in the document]
bdice: If you want an input to be a `list`, it's better to _convert to a list_ than to just _assert that it is a list_. That's why I think `self._jobs = list(jobs)` (conversion, not assertion) is the right approach.

Changing to `tuple` is a separate question. I think it would be the better choice from a data structure perspective -- the aggregate's jobs should be immutable, and as a secondary consideration, `tuple` is almost always faster than `list`.

flow/project.py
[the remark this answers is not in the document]
bdice: Yes, especially if that makes it unique.

flow/project.py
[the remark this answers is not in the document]
bdice: Sorry if this was unclear -- my actual question is "What does `nranks(job1, job2, job3)` return?" Does it return a list of integers (ranks for each job)? If it returns a list, then `nr` is a list. If you have `nr = [4, 2, 8]` and `nt = [1, 1, 1]`, then you have a problem because lists can't be multiplied (NumPy arrays _can_ be multiplied but I don't want to use NumPy here).

flow/project.py
bdice: @csadorf It looks like you wrote this line around version 0.6.0. I don't see any other code referencing it, so I think it must be safe to delete?

flow/project.py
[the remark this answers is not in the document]
kidrahahjo: I don't think a list of MPI ranks should be returned here, it should only return an integer. 
@b-butler please suggest.

flow/project.py
vyasr: I don't know how this code snuck onto master, but it was added as part of the original push for aggregation, see [this line](https://github.com/glotzerlab/signac-flow/pull/52/files#diff-bc465f25a5bed755976338a6fd0e34ceR186) from #52. At the time we were working actively on getting the decorator syntax etc into shape, so this line probably accidentally made its way into one of those PRs.

flow/project.py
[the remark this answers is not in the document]
kidrahahjo: @bdice I gave this a thought and I don't think that we'll need this to be unique.
Since `_JobOperation` is private hence users will never be able to print it by themselves.
Moreover, this is mainly used in scripts and we already print the script in this manner

```
Submitting cluster job 'playing-with/ee5ca9ab-6704a2f9/50/compute_sum/0000/c7033c1762ed0658882bc75c23dda4d5':
 - Group: compute_sum-50: (ee5-670)
```
 Which makes a submission of `_JobOperation` readable as well as unique.

flow/project.py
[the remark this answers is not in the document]
csadorf: I agree with @bdice that a `tuple` is likely the preferred data structure, but I don't agree that *converting is better than asserting* since the former triggers a potential copy operation. Since this is an internal class we can make certain assumptions about the passed arguments and the way to explicitly declare those assumptions would be to use an `assert` as was done here.

flow/project.py
```diff
@@ -291,13 +304,19 @@ def __init__(self, id, name, job, cmd, directives=None):
...
 
     def __str__(self):
-        return "{}({})".format(self.name, self.job)
+        if len(self._jobs) == 1:
+            return "{}({})".format(self.name, str(self._jobs[0]))
+        else:
+            minimum = self._jobs[0]._project.min_len_unique_id()
+            return "{}-{}: ({}-{})".format(
+                self.name, len(self._jobs), str(self._jobs[0])[:minimum],
+                str(self._jobs[-1])[:minimum])
```
csadorf, line 313: I'd prefer to use f-strings whenever possible (readability + performance).

flow/project.py
```diff
@@ -291,13 +304,19 @@ def __init__(self, id, name, job, cmd, directives=None):
         self.directives._keys_set_by_user = keys_set_by_user
 
     def __str__(self):
-        return "{}({})".format(self.name, self.job)
+        if len(self._jobs) == 1:
+            return "{}({})".format(self.name, str(self._jobs[0]))
+        else:
```
csadorf, line 309: I keep repeating that I'd like to treat all operations as similar as possible and that the "classic single-argument" operations are simply aggregate operations of size 1. This should be reflected here as well and I would very much prefer to use the same representation of the class for size 1 and larger. Having a fundamentally different representation for size 1 is unexpected and will likely cause a lot of confusion.

This is not a trivial issue, ensuring that users (and developers!) understand that there is nothing fundamentally different about an aggregate operation of size 1 is crucial for a good understanding of the code logic and ultimately usage.

flow/project.py
```diff
@@ -326,20 +345,98 @@ def cmd(self):
 
     def set_status(self, value):
         "Store the operation's status."
-        self.job._project.document.setdefault('_status', dict())[self.id] = int(value)
+        self._jobs[0]._project.document.setdefault('_status', dict())[self.id] = int(value)
```
csadorf, line 348: Are the jobs sorted in some way? What if the order changes? Are we still able to retrieve this? I think I'd prefer to store this in the project-doc.

flow/project.py
```diff
@@ -326,20 +345,98 @@ def cmd(self):
...
 
     def get_status(self):
         "Retrieve the operation's last known status."
         try:
-            return JobStatus(self.job._project.document['_status'][self.id])
+            return JobStatus(self._jobs[0]._project.document['_status'][self.id])
```
csadorf, line 353: Same question as above.

flow/project.py
```diff
@@ -326,20 +345,98 @@ def cmd(self):
...
             return JobStatus.unknown
 
 
 @deprecated(
     deprecated_in="0.11", removed_in="0.13", current_version=__version__)
 class JobOperation(_JobOperation):
-    pass
+    """This class represents the information needed to execute one group for one job.
+
+    The execution or submission of a :py:class:`FlowGroup` uses a passed in command
```
csadorf, line 363:
```suggestion
    The execution or submission of a :py:class:`FlowGroup` uses a passed-in command
```
?

flow/project.py
```diff
@@ -326,20 +345,98 @@ def cmd(self):
...
+        A dictionary of additional parameters that provide instructions on how
+        to execute this operation, e.g., specifically required resources.
+    :type directives:
+        :class:`dict`
+    """
+    def __init__(self, id, name, job, cmd, directives=None):
+        self._id = id
+        self.name = name
+
+        assert isinstance(job, signac.contrib.job.Job)
```
csadorf, line 400: Why? No duck-typing allowed?

flow/project.py
```diff
@@ -326,20 +345,98 @@ def cmd(self):
...
+        self.directives = TrackGetItemDict(
+            {key: value for key, value in directives.items()})
+        self.directives._keys_set_by_user = keys_set_by_user
+
+    @property
+    def job(self):
+        # The length of `_jobs` will never be more than one due to the
+        # instance check we're doing while initializing an instance of
+        # JobOperation. Hence return only the first element of `self._jobs`
+        return self._jobs[0]
```
csadorf, line 431: Both the function itself as well as the comment appear outdated?

flow/project.py
```diff
@@ -326,20 +345,98 @@ def cmd(self):
...
+
+    def __repr__(self):
+        return "{type}(name='{name}', job='{job}', cmd={cmd}, directives={directives})".format(
+                   type=type(self).__name__,
+                   name=self.name,
+                   job=str(self._jobs[0]),
```
csadorf, line 437: I am a bit confused by this PR. I thought the whole point was to internally refactor to use lists instead of single job arguments in preparation for aggregation, but then some classes are somehow supporting only a single job? Can you briefly elaborate on this?

flow/project.py
```diff
@@ -493,21 +592,43 @@ def eligible(self, job, ignore_conditions=IgnoreConditions.NONE):
                 "must be a member of class IgnoreConditions")
         # len(self._prereqs) check for speed optimization
         pre = (not len(self._prereqs)) or (ignore_conditions & IgnoreConditions.PRE) \
-            or all(cond(job) for cond in self._prereqs)
+            or all(cond(jobs) for cond in self._prereqs)
```
csadorf, line 595: I thought we had agreed that condition functions have signature `cond(*jobs)`?

flow/project.py
```diff
@@ -493,21 +592,43 @@ def eligible(self, job, ignore_conditions=IgnoreConditions.NONE):
...
         if pre and len(self._postconds):
             post = (ignore_conditions & IgnoreConditions.POST) \
-                or any(not cond(job) for cond in self._postconds)
+                or any(not cond(jobs) for cond in self._postconds)
```
csadorf, line 598: Same as above.

flow/project.py
```diff
@@ -493,21 +592,43 @@ def eligible(self, job, ignore_conditions=IgnoreConditions.NONE):
...
         else:
             post = True
         return pre and post
 
-    def complete(self, job):
+    def _complete(self, jobs):
         "True when all post-conditions are met."
         if len(self._postconds):
-            return all(cond(job) for cond in self._postconds)
+            return all(cond(jobs) for cond in self._postconds)
```
csadorf, line 606: Same as above.

flow/project.py
```diff
@@ -532,11 +653,19 @@ def __init__(self, cmd, pre=None, post=None):
     def __str__(self):
         return "{type}(cmd='{cmd}')".format(type=type(self).__name__, cmd=self._cmd)
 
-    def __call__(self, job=None):
+    def __call__(self, *jobs):
+        if len(jobs) == 0:
```
csadorf, line 657: I think you want something like this:
```suggestion
    def __call__(self, *jobs, **kwargs):
        job = kwargs.pop('job', None)
        if kwargs:
            raise ValueError(f"Invalid key-word arguments: {', '.join(kwargs)}")
        if job is not None:
            warnings.warn(...)
```

flow/project.py
```diff
@@ -700,42 +829,43 @@ def __init__(self, name, operations=None, operation_directives=None,
...
         """
         entrypoint = entrypoint.copy()
-        self._set_entrypoint_item(entrypoint, directives, 'executable', sys.executable, job)
+        self._set_entrypoint_item(entrypoint, directives, 'executable', sys.executable, jobs)
 
         # If a path is not provided, default to the path to the file where the
         # FlowProject (subclass) is defined.
-        default_path = inspect.getfile(job._project.__class__)
-        self._set_entrypoint_item(entrypoint, directives, 'path', default_path, job)
+        # We are assuming that all the jobs belong from the same project
```
csadorf, line 853:
```suggestion
        # We are assuming that all the jobs belong to the same project
```

flow/project.py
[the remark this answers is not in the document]
csadorf: I think it should be an integer.

flow/project.py
```diff
@@ -747,32 +877,34 @@ def np_callable(job):
...
         if isinstance(operation, FlowCmdOperation):
-            return operation(job).lstrip()
+            return operation(*jobs).lstrip()
         else:
-            entrypoint = self._determine_entrypoint(entrypoint, directives, job)
-            return '{} exec {} {}'.format(entrypoint, operation_name, job).lstrip()
+            entrypoint = self._determine_entrypoint(entrypoint, directives, jobs)
+            return '{} exec {} {}'.format(
+                        entrypoint, operation_name, ' '.join(map(str, jobs))
+                    ).lstrip()
```
csadorf, line 907: Please use f-strings.

flow/project.py
```diff
@@ -881,25 +1051,30 @@ def _generate_id(self, job, operation_name=None, index=0):
         if max_len < len(job_op_id):
             raise ValueError("Value for MAX_LEN_ID is too small ({}).".format(self.MAX_LEN_ID))
 
+        if len(jobs) > 1:
+            aggregate_id = str(jobs[0])[:8]+'-'+str(jobs[-1])[:8]
+        else:
+            aggregate_id = str(jobs[0])[:8]
```
csadorf, line 1057: This id does not seem sufficiently unique to me.

flow/project.py
csadorf: Looks like left-over code, probably safe to delete.

tests/test_project.py
```diff
@@ -505,22 +505,22 @@ def op4(job):
         op3_ = project.operations['op3']
         op4_ = project.operations['op4']
         for job in project:
-            assert not op3_.eligible(job)
-            assert not op4_.eligible(job)
+            assert not op3_._eligible([job])
+            assert not op4_._eligible([job])
```
csadorf, line 509: This is not a pure refactoring PR, because this changes public API. I'd have preferred to have that separated out.

flow/project.py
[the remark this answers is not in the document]
kidrahahjo: @csadorf  I'm just trying not to change any documentation here.
I think we should do all the documentation change in #336 as we'll have to change a lot of documentation in that PR.

flow/project.py
kidrahahjo: Thanks for this suggestion. I'll use f-strings whenever possible.

flow/project.py
kidrahahjo: That's a nice point. 
Thanks for this!

flow/project.py
kidrahahjo: Since we're not yet introducing the actual aggregation, I don't think there is a string reason to implement this for this PR. The reason behind my comment is that storing is only helpful for status-check and I have implemented this is #335.
This PR only changes the internal workflow logic.

flow/project.py
kidrahahjo: Addressed [here](https://github.com/glotzerlab/signac-flow/pull/324#discussion_r467892102)

flow/project.py
kidrahahjo: This comment here is not outdated.
We deprecated `JobOperation` in #325 and we copied the property method `job` from `JobOperation` into `_JobOperation`.
We don't have the attribute `_job` in `JobOperation` anymore. Instead we now have `_jobs` hence to maintain backwards compatibility, this method was implemented. 
Please Note: We don't need the property method `job` for `_JobOperation` (as we don't expose `_JobOperation` to users) hence I moved that method to `JobOperation`.

flow/project.py
kidrahahjo: While converting everything internally as a list of jobs, we have to make sure that we maintain backwards compatibility.
This means that all the methods which are public in the API and takes a single job as an argument should continue with the same behaviour.

flow/project.py
kidrahahjo: This method is internal hence we pass in a list of jobs here.
As `_FlowCondition` is also internal, @b-butler suggested me to pass in a list of jobs while calling `_FlowCondition` and then unpack the value while passing the jobs to the function.

flow/project.py
kidrahahjo: Addressed [here](https://github.com/glotzerlab/signac-flow/pull/324#discussion_r467902245)

flow/project.py
kidrahahjo: Addressed [here](https://github.com/glotzerlab/signac-flow/pull/324#discussion_r467902245)

flow/project.py
csadorf: Ok, that's fine. Can you add an in-code comment so that it is easier to keep track of this?

flow/project.py
csadorf: 👍

flow/project.py
csadorf: 👍

flow/project.py
csadorf: 👍

flow/project.py
kidrahahjo: I don't think that it's necessary for this particular id to be unique.
This is used for the "readable" part of the ID. We will store every aggregate associated with jobs in #335 hence if users want, they can look up aggregates associated with that particular id in the `.aggregates/` directory

tests/test_project.py
kidrahahjo: This is a nice point. 
I can make a PR today which deprecates `eligible` and `complete` methods of `FlowGroup` and `BaseFlowOperation`

flow/project.py
kidrahahjo: I'll do that. Thank you

kidrahahjo:
As @csadorf and @bdice suggest changing the data structure from `list` to `tuple`. I will start working on this today and make a commit which adapts tuples instead of lists due to efficiency and immutable behaviour of tuples.

flow/project.py
kidrahahjo: I don't have any problem with going with duck-typing. I'll make the change.
Thanks for suggesting this.

flow/project.py
csadorf: I was not aware of #335, however in that case this variable name is misleading. An id is by definition describing the *identity*, so it **must** be unique. If this string is describing something else, then that should be reflected in the semantics here. However, I see no reason why we are not showing the actual id here?

tests/test_project.py
csadorf: If that's not too hard to do, then I think that would be good and we can simply rebase this branch here.

flow/project.py
kidrahahjo: This is not a really hard decision to make, but can you please suggest me whether I should, for now, not check the length of jobs as I'm sure that it'll always be one?

csadorf:
> As @csadorf and @bdice suggest changing the data structure from `list` to `tuple`. I will start working on this today and make a commit which adapts tuples instead of lists due to efficiency and immutable behaviour of tuples.

No, I did not suggest to change the data structure, please read my comments more carefully. I suggested to have a brief discussion of pros and cons concerning the data structure. I *assume* that a tuple would be preferable here, but changing the data structure without a proper analysis of requirements and implications is premature.

flow/project.py
kidrahahjo: I can remove the substring part of the job in jobs(`[:8]`) and instead use the full job id, but if we decide to include all the job ids then this will get way too long for large aggregates.

flow/project.py
csadorf: Add `assert len(self._jobs) == 1` here and then remove it once you need to drop that assertion in a different change set.

flow/project.py
csadorf: Please reread my comment. If it is not necessary for this to be an id, that's fine, but then please do not call id *id*.

flow/project.py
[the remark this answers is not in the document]
b-butler: If we want to use tuples for immutability which makes sense here we would want the structure of aggregation to be a generator or list of aggregate tuples. I don't have a strong preference for the assert. Ideally we wouldn't need an assert or explicit conversion.

flow/project.py
```diff
@@ -291,13 +304,14 @@ def __init__(self, id, name, job, cmd, directives=None):
         self.directives._keys_set_by_user = keys_set_by_user
 
     def __str__(self):
-        return "{}({})".format(self.name, self.job)
+        assert len(self._jobs) == 1
+        return f"{self.name}({str(self._jobs[0])})"
```
b-butler, line 308: This is an internal class we don't need the `len == 1` assumption. We could use this in `JobOperation` though.

flow/project.py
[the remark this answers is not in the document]
b-butler: @kidrahahjo I would make a note to do a grep through project.py for `job` for changing in #306. Otherwise, these will be easy to miss.

flow/project.py
```diff
@@ -532,7 +648,17 @@ def __init__(self, cmd, pre=None, post=None):
...
         return "{type}(cmd='{cmd}')".format(type=type(self).__name__, cmd=self._cmd)
 
-    def __call__(self, job=None):
+    def __call__(self, *jobs, **kwargs):
+        job = kwargs.pop('job', None)
+        if kwargs:
+            raise ValueError(f"Invalid key-word arguments: {', '.join(kwargs)}")
+
+        if job is not None:
+            warnings.warn("The job argument is deprecated as of 0.11 and will be removed "
```
b-butler, line 659:
```suggestion
            warnings.warn("The job keyword argument is deprecated as of 0.11 and will be removed "
```

flow/project.py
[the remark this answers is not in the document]
b-butler: I concur. We should return an integer. An aggregate operation is a singular operation that has a singular resource requirement (partitioning the resources internally is the user's responsibility).

flow/project.py
b-butler: @kidrahahjo Bradley is referring to meta data is more than just the list of job ids. @bdice I do agree this is not a problem in the script working (long lines are not going to freeze up the scheduler). However, it would get inconvenient for users to look at their scripts. I am okay with waiting on this (I planned on doing that for this PR anyways), but I think raising an issue is important so we don't lose track of this.

flow/project.py
```diff
@@ -1563,9 +1737,9 @@ def _get_operations_status(self, job, cached_status):
         starting_dict = functools.partial(dict, scheduler_status=JobStatus.unknown)
         status_dict = defaultdict(starting_dict)
         for group in self._groups.values():
-            completed = group.complete(job)
-            eligible = False if completed else group.eligible(job)
-            scheduler_status = cached_status.get(group._generate_id(job),
+            completed = group._complete([job])
+            eligible = False if completed else group._eligible([job])
```
b-butler, line 1743: This could raise errors if single jobs are not accepted for a group correct? I know that this will not be a problem in this PR, but in the future it could be. I imagine this function is to be removed or unused in the status check PR? If so, than this is fine for now; otherwise, we need to change the function to work with aggregates of jobs.

flow/project.py
```diff
@@ -2226,11 +2400,12 @@ class _PickleError(Exception):
...
     @staticmethod
     def _dumps_op(op):
-        return (op.id, op.name, op.job._id, op.cmd, op.directives)
+        return (op.id, op.name, [job.get_id() for job in op._jobs], op.cmd, op.directives)
 
     def _loads_op(self, blob):
-        id, name, job_id, cmd, directives = blob
-        return _JobOperation(id, name, self.open_job(id=job_id), cmd, directives)
+        id, name, job_ids, cmd, directives = blob
+        jobs = [self.open_job(id=job_id) for job_id in job_ids]
```
b-butler, line 2409: Make a tuple. I would use a generator and the tuple function.

flow/project.py
```diff
@@ -2283,11 +2458,11 @@ def _execute_operation(self, operation, timeout=None, pretend=False):
                 "Executing operation '{}' with current interpreter "
                 "process ({}).".format(operation, os.getpid()))
             try:
-                self._operations[operation.name](operation.job)
+                self._operations[operation.name](*operation._jobs)
             except Exception as e:
                 raise UserOperationError(
                     'An exception was raised during operation {operation.name} '
-                    'for job {operation.job}.'.format(operation=operation)) from e
+                    'for job {operation._jobs[0]}.'.format(operation=operation)) from e
```
b-butler, line 2468: You don't need to in this PR in my opinion, but this is another example of us needing to change `job` to `jobs` in a future PR.

flow/project.py
```diff
@@ -2520,28 +2694,37 @@ def _get_submission_operations(self, jobs, default_directives, names=None,
...
-        for op in self._next_operations(* jobs, ignore_conditions=ignore_conditions):
+        for op in self._next_operations(jobs, ignore_conditions):
             if operation_names is None or any(re.fullmatch(n, op.name) for n in operation_names):
                 yield op
 
     def _verify_group_compatibility(self, groups):
         """Verifies that all selected groups can be submitted together."""
         return all(a.isdisjoint(b) for a in groups for b in groups if a != b)
 
+    def _verify_aggregate_project(self, aggregate):
```
b-butler, line 2724: Given the internal nature of `_JobOperation` now this is likely not necessary once `JobOperation` is deprecated completely.

I would make a developer comment regarding that for the future.

flow/project.py
```diff
@@ -3227,25 +3410,25 @@ def completed_operations(self, job):
...
-    def _next_operations(self, *jobs, ignore_conditions=IgnoreConditions.NONE):
+    def _next_operations(self, jobs, ignore_conditions=IgnoreConditions.NONE):
         """Determine the next eligible operations for jobs.
 
         :param jobs:
             The signac job handles.
-        :type job:
-            :class:`~signac.contrib.job.Job`
+        :type jobs:
+            list of :class:`~signac.contrib.job.Job`
```
b-butler, line 3434: If we change to tuples we need to change this docstring.

flow/project.py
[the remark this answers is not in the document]
csadorf: You're right.

kidrahahjo:
Since the team agrees upon the use of tuples internally rather than jobs, I'll now go ahead and implement this functionality.

flow/project.py
kidrahahjo: After we include actual aggregation, this will have to change to something like.
`f"{self.name}-{len(self._jobs)}: ({str(self._jobs[0])}-{str(self._jobs[-1])})"` for every aggregate (even for length 1).

Please see this [comment](https://github.com/glotzerlab/signac-flow/pull/324#discussion_r467830066) by @csadorf

flow/project.py
kidrahahjo: @b-butler  I'm not sure I completely follow what you suggest here.
Can you please help me understand?

flow/project.py
kidrahahjo: This is not the case because we're going to iterate over `group.aggregate(jobs)` in #336 which will always raise an error if and only if the users have failed to specify the function parameters for the conditions correctly.

flow/project.py
kidrahahjo: I have inserted a command `assert len(operation._jobs) == 1` above and throughout the codebase.
This was suggested by @csadorf so we know that what things we have to change when we include actual aggregation.

flow/project.py
b-butler: I am aware. I more mean, we can do that now given this is an internal class now. We just need this behavior for `JobOperation`

flow/project.py
b-butler: Okay, as long as this is changed there.

flow/project.py
b-butler: That only works with testing as well. We won't raise the assert until we actually test aggregation, but that does help. This wasn't really an issue just an example.

flow/project.py
b-butler: Actually, I am wrong on the point above. However, I would look for where users supply jobs in the API, (e.g. `FlowProject.submit` and such) and place the verification there. I don't think we would ever be in an internal state where the jobs we gather internally don't belong to the same project.

flow/project.py
kidrahahjo: @b-butler If we do this, I'll have to change the template-reference-data twice(once here and another in #334 )
This will result in a lot of additional work.
I think this should be addressed in #334

flow/project.py
```diff
@@ -263,10 +273,13 @@ class _JobOperation(object):
         :class:`dict`
     """
 
-    def __init__(self, id, name, job, cmd, directives=None):
+    def __init__(self, id, name, jobs, cmd, directives=None):
         self._id = id
         self.name = name
-        self.job = job
+
+        assert isinstance(jobs, tuple)
```
kidrahahjo, line 280: This is for code security. If the team want me to remove it, I'll do that.

flow/project.py
```diff
@@ -2668,7 +2672,7 @@ def _generate_operations(self, cmd, jobs, requires=None):
             if requires and set(requires).difference(self.labels(job)):
                 continue
             cmd_ = cmd.format(job=job)
-            yield _JobOperation(name=cmd_.replace(' ', '-'), cmd=cmd_, job=job)
+            yield JobOperation(name=cmd_.replace(' ', '-'), cmd=cmd_, job=job)
```
b-butler, line 2675: Why are we using `JobOperation` rather than `_JobOperation`?

kidrahahjo: I think I accidentally did this

kidrahahjo:
@b-butler I have added the job check for these methods `print_status`, `run`, `submit`. Other than these, I don't think there's a need to add this check. Please suggest.

flow/project.py
csadorf: Since this is an internal class, I believe it is perfectly fine to expect a specific container type here.

flow/project.py
csadorf: The assert should be in place wherever we actually make that assumption. Once that assumption is removed (e.g. in #334), the asserts will automatically alert us to spots in the code that require a revised implementation.

flow/project.py
csadorf: Does this even work? The `__repr__` return value should be able to be used in `eval()` like this: `new_foo == eval(repr(foo))`.

flow/project.py
csadorf: You marked this conversation as resolved, yet this is still not an f-string?

kidrahahjo: @csadorf I apologize. I wrote a comment but instead of posting, I must have clicked resolved. I'll take care of this from the next time.
The comment was: 
Throughout the code-base, we don't use f-strings. Instead, we format the strings in this manner `'{name}-{operation}'.format(name=name, operation=op_name)`.
I understand that using f-strings will be more efficient here but I think we should be consistent.
If you want, I'll use an f-string in this PR or I will make another PR which will internally convert all the string formatting to f-strings formatting to maintain consistency.
Please suggest

flow/project.py
kidrahahjo: This should possibly get resolved by using `map(repr, self._jobs)`

kidrahahjo:
@csadorf I addressed the comments which were not resolved in the `diff`. Please have a look

flow/project.py
csadorf: I think it is totally fine to prefer the use of f-strings moving forward even if we use `"...".format()` in other sections of the code.

kidrahahjo: Will make the changes now.

flow/project.py
```diff
@@ -326,20 +337,99 @@ def cmd(self):
...
+        # keys set by the user and check whether they have been evaluated by the template
+        # script engine later.
+        keys_set_by_user = set(directives)
+
+        # We use a special dictionary that allows us to track all keys that have been
+        # evaluated by the template engine and compare them to those explicitly set
+        # by the user. See also comment above.
+        self.directives = TrackGetItemDict(
+            {key: value for key, value in directives.items()})
+        self.directives._keys_set_by_user = keys_set_by_user
```
b-butler, line 419:
```suggestion
        jobs = (job,)
        super().__init__(id, name, job, cmd, directives)
```

flow/project.py
```diff
@@ -912,9 +1016,11 @@ def isdisjoint(self, group):
         """
         return set(self).isdisjoint(set(group))
 
-    def _generate_id(self, job, operation_name=None, index=0):
+    def _generate_id(self, jobs, operation_name=None, index=0):
         "Return an id, which identifies this group with respect to this job."
-        project = job._project
+        project = jobs[0]._project
+
+        project._verify_aggregate_project(jobs)
```
b-butler, line 1023: This check is unnecessary as it is internal.

kidrahahjo: @b-butler 
Since aggregator function is user-facing.
I think this check is necessary as user can yield jobs from a different project inside that method.

flow/project.py
kidrahahjo: After using a deprecated decorator, this class behaves like a method hence we have to manually define an init method

b-butler: I had forgotten thanks for the reminder.

flow/project.py
b-butler: Is `_generate_id` the right place to check this though then? If we are worried about aggregation should we not check closer to the aggregation? Even if it requires the function in a couple more places, it will be more clear to the developer as to why the check exists in the first place.

b-butler:
As for pros/cons on lists versus tuples. I don't think there are any cons with the tuples approach. Once a particular aggregate is generated, the idea is it is used immediately to create a `_JobOperation` which is not designed to have its jobs change. A tuple explicitly states this assumption. Any dynamic nature to aggregates is to be found in their generation.

csadorf:
@bdice Do you want to have a final look?

flow/project.py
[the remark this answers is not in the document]
bdice: Clarified via Slack, marking as resolved. I was misunderstanding this code. Thanks @csadorf @b-butler.

Issue #311, opened 2020-06-30
Deprecate the script command


vyasr:
The `FlowProject.script` method is mostly redundant to `submit(pretend=True)`. The primary difference is that submit also includes the scheduler directives. @csadorf has previously argued that supporting this mode of getting a runnable script is important so that users could easily generate a submittable script, e.g. `python project.py script > run.sh`. I think that's a valid use, but it's one we could support by adding an extra argument `no_header` or similar to submit, and all we would have to do is ignore the environment (submitting on the `StandardEnvironment` is equivalent to `script` I believe).

csadorf:
I don't think this is a really heavy used function, but I also don't quite see the point of deprecating it. If it is that easy to just use the `StandardEnvironment` in combination with `pretend=True`, why not leave this very thin wrapper function in place?

vyasr:
Actually I think what would be the most helpful is to unify `script` with `_generate_submit_script`. We can leave `script` as part of the API and just extend its arguments to account for both use-cases. With this change we should be able to unify the upstream logic as well with respect to `_main_script` and `_main_submit(pretend=True)`. So only an internal refactoring, no need to change user APIs (except adding a few extra optional arguments to `script`).

vyasr:
In light of #313, we are reverting to the original plan of deprecating the script command since the `JobOperation` class is being removed. However, as discussed above, re-enabling the same behavior using something like an underlying call to `_generate_submit_script` with a `StandardEnvironment` remains an option in the future version of the API.

vyasr:
@csadorf @bdice do we still want to bother supporting the command line script interface now that we're removing the Python functionality? If we want to, I outlined the path forward above (have it call `_generate_submit_script` with the necessary environment mocking.

csadorf:
> @csadorf @bdice do we still want to bother supporting the command line script interface now that we're removing the Python functionality? If we want to, I outlined the path forward above (have it call `_generate_submit_script` with the necessary environment mocking.

I always thought it was kind of nice to check the scripts using this command, on the other hand, one would want to use `run` anyways in a non-cluster/scheduler environment, so there is not much of a point. I also think that `run --pretend` can server the purpose of checking commands. So probably fine to drop this.

bdice:
I always use `--pretend` and prefer that approach over the `script` command. I support deprecation.

csadorf:
Looks like it is decided to deprecate.

PR #325, opened 2020-07-18
Deprecate JobOperation and the methods which uses its instances 

### LINKED ISSUE #313: Making JobOperation Private

kidrahahjo:
## Description

This PR deprecates the class `JobOperation` and the methods which use the `JobOperation` instances as their parameters.
User is warned via a deprecation warning.

Made SubmissionJobOperation private without any deprecation because this class is not exposed anywhere to the user.

## Motivation and Context

Resolves #313 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -1259,7 +1259,7 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
         """Setup the jinja2 template environment.
 
         The templating system is used to generate templated scripts for the script()
-        and submit_operations() / submit() function and the corresponding command line
+        and _submit_operations() / submit() function and the corresponding command line
         subcommands.
         """
         if self._config.get('flow') and self._config['flow'].get('environment_modules'):
@@ -2208,6 +2208,7 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
                             raise RuntimeError("Unable to parallelize execution due to a pickling "
                                                "error: {}.".format(error))
 
+    @deprecated(deprecated_in="0.11", removed_in="0.13", current_version=__version__)
     def run_operations(self, operations=None, pretend=False, np=None, timeout=None, progress=False):
         """Execute the next operations as specified by the project's workflow.
 
@@ -2632,9 +2633,9 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
             self._show_template_help_and_exit(template_environment, context)
         return template.render(** context)
 
-    def submit_operations(self, operations, _id=None, env=None, parallel=False, flags=None,
-                          force=False, template='script.sh', pretend=False,
-                          show_template_help=False, **kwargs):
+    def _submit_operations(self, operations, _id=None, env=None, parallel=False, flags=None,
+                           force=False, template='script.sh', pretend=False,
+                           show_template_help=False, **kwargs):
         r"""Submit a sequence of operations to the scheduler.
 
         :param operations:
@@ -2729,6 +2730,58 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
             else:
                 return env.submit(_id=_id, script=script, flags=flags, **kwargs)
 
+    @deprecated(deprecated_in="0.11", removed_in="0.13", current_version=__version__)
+    def submit_operations(self, operations, _id=None, env=None, parallel=False, flags=None,
+                          force=False, template='script.sh', pretend=False,
+                          show_template_help=False, **kwargs):
+        r"""Submit a sequence of operations to the scheduler.
+
+        :param operations:
+            The operations to submit.
+        :type operations:
+            A sequence of instances of :py:class:`.JobOperation`
+        :param _id:
+            The _id to be used for this submission.
+        :type _id:
+            str
+        :param parallel:
+            Execute all bundled operations in parallel.
+        :type parallel:
+            bool
+        :param flags:
+            Additional options to be forwarded to the scheduler.
+        :type flags:
+            list
+        :param force:
+            Ignore all warnings or checks during submission, just submit.
+        :type force:
+            bool
+        :param template:
+            The name of the template file to be used to generate the submission script.
+        :type template:
+            str
+        :param pretend:
+            Do not actually submit, but only print the submission script to screen. Useful
+            for testing the submission workflow.
+        :type pretend:
+            bool
+        :param show_template_help:
+            Show information about available template variables and filters and exit.
+        :type show_template_help:
+            bool
+        :param \*\*kwargs:
+            Additional keyword arguments to be forwarded to the scheduler.
+        :return:
+            Returns the submission status after successful submission or None.
+        """
+        warnings.warn("The submit_operations method is deprecated as of 0.11 and "
+                      "will be removed in 0.13.",
+                      DeprecationWarning)
+
+        return self._submit_operations(operations, _id, env, parallel, flags,
+                                       force, template, pretend,
+                                       show_template_help, **kwargs)
+
     def submit(self, bundle_size=1, jobs=None, names=None, num=None, parallel=False,
                force=False, walltime=None, env=None, ignore_conditions=IgnoreConditions.NONE,
                ignore_conditions_on_execution=IgnoreConditions.NONE, **kwargs):
@@ -2807,7 +2860,7 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
 
         # Bundle them up and submit.
         for bundle in _make_bundles(operations, bundle_size):
-            status = self.submit_operations(operations=bundle, env=env, parallel=parallel,
+            status = self._submit_operations(operations=bundle, env=env, parallel=parallel,
                                             force=force, walltime=walltime, **kwargs)
             if status is not None:  # operations were submitted, store status
                 for operation in bundle:
--- a/flow/project.py
+++ b/flow/project.py
@@ -245,7 +245,9 @@ def make_bundles(operations, size=None):
     return _make_bundles(operations, size)
 
 
-class JobOperation(object):
+@deprecated(
+    deprecated_in="0.11", removed_in="0.13", current_version=__version__)
+def JobOperation(id, name, job, cmd, directive=None):
     """This class represents the information needed to execute one group for one job.
 
     The execution or submission of a :py:class:`FlowGroup` uses a passed in command
@@ -259,11 +261,54 @@ class JobOperation(object):
         submission process and should not be instantiated by users themselves.
 
     :param id:
-        The id of this JobOperation instance. The id should be unique.
+        The id of this _JobOperation instance. The id should be unique.
     :type id:
         str
     :param name:
-        The name of the JobOperation.
+        The name of the _JobOperation.
+    :type name:
+        str
+    :param job:
+        The job instance associated with this operation.
+    :type job:
+        :py:class:`signac.Job`.
+    :param cmd:
+        The command that executes this operation. Can be a function that when
+        evaluated returns a string.
+    :type cmd:
+        callable or str
+    :param directives:
+        A dictionary of additional parameters that provide instructions on how
+        to execute this operation, e.g., specifically required resources.
+    :type directives:
+        :class:`dict`
+    """
+    warnings.warn("The JobOperation class is deprecated as of 0.11 and "
+                  "will be removed in 0.13.",
+                  DeprecationWarning)
+
+    return _JobOperation(id, name, job, cmd, directive=None)
+
+
+class _JobOperation(object):
+    """This class represents the information needed to execute one group for one job.
+
+    The execution or submission of a :py:class:`FlowGroup` uses a passed in command
+    which can either be a string or function with no arguments that returns a shell
+    executable command.  This won't be used if it is determined that the group can be
+    executed without forking.
+
+    .. note::
+
+        This class is used by the :class:`~.FlowGroup` class for the execution and
+        submission process and should not be instantiated by users themselves.
+
+    :param id:
+        The id of this _JobOperation instance. The id should be unique.
+    :type id:
+        str
+    :param name:
+        The name of the _JobOperation.
     :type name:
         str
     :param job:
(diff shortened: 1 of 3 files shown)


flow/project.py
```diff
@@ -259,11 +261,54 @@ class JobOperation(object):
         submission process and should not be instantiated by users themselves.
 
     :param id:
-        The id of this JobOperation instance. The id should be unique.
+        The id of this _JobOperation instance. The id should be unique.
```
kidrahahjo, line 264:
```suggestion
        The id of this JobOperation instance. The id should be unique.
```

flow/project.py
```diff
@@ -259,11 +261,54 @@ class JobOperation(object):
...
     :type id:
         str
     :param name:
-        The name of the JobOperation.
+        The name of the _JobOperation.
```
kidrahahjo, line 268:
```suggestion
        The name of the JobOperation.
```

flow/project.py
```diff
@@ -2541,7 +2618,7 @@ def _potentially_buffered(self):
         else:
             yield
 
-    def script(self, operations, parallel=False, template='script.sh', show_template_help=False):
+    def _script(self, operations, parallel=False, template='script.sh', show_template_help=False):
```
b-butler, line 2621: We don't need the underscore here, we don't plan on having even an internal function here.

flow/project.py
```diff
@@ -245,7 +245,9 @@ def make_bundles(operations, size=None):
     return _make_bundles(operations, size)
 
 
-class JobOperation(object):
+@deprecated(
+    deprecated_in="0.11", removed_in="0.13", current_version=__version__)
+def JobOperation(id, name, job, cmd, directive=None):
```
b-butler, line 250:
```suggestion
class JobOperation(_JobOperation):
```

This will allow the class name to remain the same while moving all the functionality to a private class.

b-butler: Never mind, we use the `deprecation` rather than `deprecated` Python package. The `deprecated` package offers a wrapper that works with classes as well rather than just methods and functions.

flow/project.py
```diff
@@ -3169,12 +3325,36 @@ def next_operations(self, *jobs, ignore_conditions=IgnoreConditions.NONE):
...
+            The signac job handles.
+        :type job:
+            :class:`~signac.contrib.job.Job`
+        :param ignore_conditions:
+            Specify if pre and/or post conditions check is to be ignored for eligibility check.
+            The default is :py:class:`IgnoreConditions.NONE`.
+        :type ignore_conditions:
+            :py:class:`~.IgnoreConditions`
+        :yield:
+            All instances of :class:`~.JobOperation` jobs are eligible for.
```
b-butler, line 3348:
```suggestion
            All instances of :class:`~._JobOperation` jobs are eligible for.
```

flow/project.py
```diff
@@ -2208,6 +2251,40 @@ def run_operations(self, operations=None, pretend=False, np=None, timeout=None,
...
+    @deprecated(deprecated_in="0.11", removed_in="0.13", current_version=__version__)
+    def run_operations(self, operations=None, pretend=False, np=None, timeout=None, progress=False):
+        """Execute the next operations as specified by the project's workflow.
+
+        See also: :meth:`~.run`
+
+        :param operations:
+            The operations to execute (optional).
+        :type operations:
+            Sequence of instances of :class:`.JobOperation`
```
b-butler, line 2263:
```suggestion
            Sequence of instances of :class:`._JobOperation`
```

flow/project.py
```diff
@@ -2573,6 +2650,33 @@ def script(self, operations, parallel=False, template='script.sh', show_template
...
         return template.render(** context)
 
+    @deprecated(deprecated_in="0.11", removed_in="0.13", current_version=__version__)
+    def script(self, operations, parallel=False, template='script.sh', show_template_help=False):
+        """Generate a run script to execute given operations.
+
+        :param operations:
+            The operations to execute.
+        :type operations:
+            Sequence of instances of :class:`.JobOperation`
```
b-butler, line 2660:
```suggestion
            Sequence of instances of :class:`._JobOperation`
```

flow/project.py
kidrahahjo: @b-butler can you please clarify? Currently `script` uses `JobOperation` instance and I think that we should deprecate `script`.

b-butler: We should, but we don't need an internal function since we will just remove both when the user facing function is removed.

flow/project.py
vyasr: This suggestion should work fine. Did you try it and find that it didn't? For instance:

```
>>> from deprecation import deprecated
>>> @deprecated()
... class Foo():
...     pass
...
>>> Foo()
<stdin>:1: DeprecatedWarning: Foo is deprecated
<__main__.Foo object at 0x7fdd6fcf9fa0>
```

vyasr:
> I have a few comments, here. We also, have the problem that we are now exposing the internal class `_JobOperation` though the `JobOperation` function. @vyasr do we have a standard way to deprecate classes? It would be nicer if we should still have `JobOperation` to be a class. I know the `deprecated` package https://github.com/tantale/deprecated can deprecate classes.

The deprecation package should work for a class, did you test it and find problems?

b-butler:
>The deprecation package should work for a class, did you test it and find problems?

Nope just going off of the documentation of the two. If it just works though that is helpful.

kidrahahjo:
@b-butler @vyasr 
When we use the decorator `@deprecated`, we make the class behave like a method because the decorator wraps it around a method.

I don't know if it's true but using the [deprecated](https://pypi.org/project/deprecated/) package instead of [deprecation](https://pypi.org/project/deprecation/). But at the same time, the latter provides us with more features.

flow/project.py
kidrahahjo: This internal function is used for the command `python project.py script`.
Removing the internal function means that are we planning to remove the command line argument, `script`?

kidrahahjo:
Also, the deprecated decorator is meant to throw a `DeprecatedWarning` but I have to explicitly use `warnings.warn` as it's not throwing any warning for pytest to catch.
Can someone please help me understand this behaviour?

vyasr:
@kidrahahjo I don't think that's quite right; the decorator generates a callable that will print a deprecation warning but otherwise transparently  and then return the result of the decorated callable. When you apply it, the generated callable is a method, but the return value of the method is the return value of the underlying callable. When applied to a class, which is itself callable (effectively the call operator for a class is the constructor, with the result being an initialized instance), the result is an instance of the class. Under the hood, I assume it's just using `functools.wraps` to mirror the docstring, name, etc. The result should be that applying this decorator has no effect on the constructed class.

Unless there's something that breaks as a result of using deprecation, I don't think there's a strong enough reason to change packages.

flow/project.py
vyasr: See the discussion on #311. The appropriate solution is to have the script CLI call the underlying `submit` API (temporarily mocking the environment as `StandardEnvironment`). In the interest of keeping PRs small, though I'd be fine with adding this underscore function in this PR and then making a follow-up PR where you refactor the CLI script call to not rely on this method anymore. Does that make sense? @b-butler let me know if I missed something here, but if not, then I think no further changes are needed in this PR.

vyasr:
@kidrahahjo regarding your question about why deprecation warnings aren't shown, Python [suppresses `DeprecationWarning` by default](https://docs.python.org/3/library/warnings.html#warning-categories).

flow/project.py
b-butler: That is fine for now then since we have an issue.

flow/project.py
```diff
@@ -245,6 +245,122 @@ def make_bundles(operations, size=None):
...
+        "Retrieve the operation's last known status."
+        try:
+            return JobStatus(self.job._project.document['_status'][self.id])
+        except KeyError:
+            return JobStatus.unknown
+
+
+@deprecated(
+    deprecated_in="0.11", removed_in="0.13", current_version=__version__)
 class JobOperation(object):
```
b-butler, line 364: I would just make `JobOperation` inherit from `_JobOperation` then it should work without the need to keep all the code in two places.

kidrahahjo: @b-butler I tried that but this code gives an error that super method requires a class not a function.
`super(JobOperation, self).__init__(*args)`

flow/project.py
```diff
@@ -2208,6 +2321,36 @@ def run_operations(self, operations=None, pretend=False, np=None, timeout=None,
...
+    @deprecated(deprecated_in="0.11", removed_in="0.13", current_version=__version__)
+    def run_operations(self, operations=None, pretend=False, np=None, timeout=None, progress=False):
+        """Execute the next operations as specified by the project's workflow.
+
+        See also: :meth:`~.run`
+
+        :param operations:
+            The operations to execute (optional).
+        :type operations:
+            Sequence of instances of :class:`._JobOperation`
```
b-butler, line 2333:
```suggestion
            Sequence of instances of :class:`.JobOperation`
```
Public facing methods even with deprecation should reference the public class even if that is also deprecated.

flow/project.py
```diff
@@ -2696,6 +2862,54 @@ def _msg(group):
...
+    @deprecated(deprecated_in="0.11", removed_in="0.13", current_version=__version__)
+    def submit_operations(self, operations, _id=None, env=None, parallel=False, flags=None,
+                          force=False, template='script.sh', pretend=False,
+                          show_template_help=False, **kwargs):
+        r"""Submit a sequence of operations to the scheduler.
+
+        :param operations:
+            The operations to submit.
+        :type operations:
+            A sequence of instances of :py:class:`._JobOperation`
```
b-butler, line 2874:
```suggestion
            A sequence of instances of :py:class:`.JobOperation`
```

flow/project.py
```diff
@@ -3169,12 +3383,33 @@ def next_operations(self, *jobs, ignore_conditions=IgnoreConditions.NONE):
...
+            The signac job handles.
+        :type job:
+            :class:`~signac.contrib.job.Job`
+        :param ignore_conditions:
+            Specify if pre and/or post conditions check is to be ignored for eligibility check.
+            The default is :py:class:`IgnoreConditions.NONE`.
+        :type ignore_conditions:
+            :py:class:`~.IgnoreConditions`
+        :yield:
+            All instances of :class:`~._JobOperation` jobs are eligible for.
```
b-butler, line 3406: Here we do return `_JobOperation` objects. I am not sure what the policy here is, @vyasr. We could transform from `_JobOperation` to `JobOperation` but that seems like too much work.

tests/test_project.py
```diff
@@ -446,30 +447,34 @@ def a(job):
...
             del job.doc['omp_num_threads']
 
         # test setting both nranks and omp_num_threads
         for i, job in enumerate(project):
             job.doc.omp_num_threads = i+1
             job.doc.nranks = i % 3 + 1
             expected_np = (i + 1) * (i % 3 + 1)
-            for next_op in project.next_operations(job):
-                assert next_op.directives['np'] == expected_np
+            with pytest.deprecated_call():
```
b-butler, line 475: Just use the `_next_operations` function no need to wrap in `pytest.deprecated_call`.

tests/test_project.py
```diff
@@ -914,11 +921,11 @@ def test_submit_operations(self):
...
         operations = []
         for job in project:
-            operations.extend(project.next_operations(job))
+            with pytest.deprecated_call():
+                operations.extend(project.next_operations(job))
         assert len(list(MockScheduler.jobs())) == 0
         cluster_job_id = project._store_bundled(operations)
-        with redirect_stderr(StringIO()):
-            project.submit_operations(_id=cluster_job_id, operations=operations)
+        project.submit_operations(_id=cluster_job_id, operations=operations)
```
b-butler, line 928: Isn't this deprecated now?

flow/project.py
vyasr: I think as long as JobOperation exists and is deprecated but not removed, that transformation is unfortunately probably necessary.

b-butler: Okay, that makes sense. @kidrahahjo It might be easiest to use the existing `_JobOperation` object's data to create corresponding `JobOperation` objects.

flow/project.py
b-butler: It worked locally for me. Don't use super just do
````python
@deprecated(...)
class JobOperation(_JobOperation):
    """Docstring here."""
    pass
````

File: flow/project.py [hunk 14]
```diff
@@ -361,7 +361,7 @@ def get_status(self):
 
 @deprecated(
     deprecated_in="0.11", removed_in="0.13", current_version=__version__)
-class JobOperation(object):
+class JobOperation(_JobOperation):
```
b-butler, line 364: We still have the `__init__` method which is also unnecessary.

flow/project.py
```diff
@@ -3169,12 +3309,33 @@ def next_operations(self, *jobs, ignore_conditions=IgnoreConditions.NONE):
...
             :py:class:`~.IgnoreConditions`
         :yield:
-            All instances of :class:`~.JobOperation` jobs are eligible for.
+            All instances of :class:`~._JobOperation` jobs are eligible for.
         """
         for job in jobs:
             for op in self._job_operations(job, ignore_conditions):
                 yield op
 
+    @deprecated(deprecated_in="0.10", removed_in="0.13", current_version=__version__)
```
b-butler, line 3318: It will be deprecated in `0.11`
```suggestion
    @deprecated(deprecated_in="0.11", removed_in="0.13", current_version=__version__)
```

flow/project.py
```diff
@@ -3169,12 +3309,33 @@ def next_operations(self, *jobs, ignore_conditions=IgnoreConditions.NONE):
...
+            The default is :py:class:`IgnoreConditions.NONE`.
+        :type ignore_conditions:
+            :py:class:`~.IgnoreConditions`
+        :yield:
+            All instances of :class:`~.JobOperation` jobs are eligible for.
+        """
+        warnings.warn("The JobOperation class is deprecated as of 0.11 and "
+                      "will be removed in 0.13.",
+                      DeprecationWarning)
+        return self._next_operations(*jobs, ignore_conditions=ignore_conditions)
```
b-butler, line 3337: We need to this to also transform `_JobOperation` objects into `JobOperation` objects. This should be doable as the necessary information is stored in the `_JobOperation` objects.

kidrahahjo: This can be resolved by not returning `self._next_operations(arugements)` but rather we could copy the code and in `next_operations` we should yield JobOperation(op.name, op.id, op.job, ...)`

b-butler: That works too. I am fine with either solution.

PR #326, opened 2020-07-18
Make FlowOperation Callable

### LINKED ISSUE #308: Make FlowOperation callable
Currently the FlowOperations work differently than FlowCmdOperations. While the latter store their command internally and can be directly evaluated, the FlowOperations require lookup through the `_operation_functions` attribute of the FlowProject. We should change this and make FlowOperation callable and have it evaluate the operation function. Removing this indirection will centralize the logic for the execution directly in BaseFlowOperation and its subclasses, making it easier to modify for future changes since we won't have to check internally whether a given operation is a FlowOperation or a FlowCmdOperation.

kidrahahjo:
## Description

Removes the internal logic:
```
if isinstance(op, FlowCmdOperation):
    func = op._cmd
else:
    func = self._operation_functions[op_name]
```

Replace the above with:
```
if isinstance(op, FlowCmdOperation):
    func = op._cmd
else:
    func = op._op_func
```
## Motivation and Context

Resolves #308 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -559,20 +559,23 @@ class FlowCmdOperation(BaseFlowOperation):
 class FlowOperation(BaseFlowOperation):
     """FlowOperation holds a Python function that does not return a shell executable string.
 
-    All operations without the ``@cmd`` directive use this class. No command can be specified, and
-    the function is not stored internally (but in an external list) stored by the
-    :class:`FlowProject`.
+    All operations without the ``@cmd`` directive use this class. Those operations should
+    be a function of :py:class:`~signac.contrib.job.Job`.
 
     .. note::
         This class should not be instantiated directly.
     """
 
-    def __init__(self, name, pre=None, post=None):
+    def __init__(self, op_func, pre=None, post=None):
         super(FlowOperation, self).__init__(pre=pre, post=post)
-        self.name = name
+        self._op_func = op_func
 
     def __str__(self):
-        return "{type}(name='{name}')".format(type=type(self).__name__, name=self.name)
+        return "{type}(op_func='{op_func}')" \
+               "".format(type=type(self).__name__, op_func=self._op_func)
+
+    def __call__(self, job):
+        return self._op_func(job)
 
 
 class FlowGroupEntry(object):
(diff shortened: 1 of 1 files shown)


kidrahahjo:
@vyasr this doesn't remove the instance check. Since we have two different variable names `_cmd` and `_op_func`, we can either use a try-except block or do the instance check. I think the instance check is more meaningful here. Please suggest.

flow/project.py
```diff
@@ -3487,7 +3488,7 @@ def _main_exec(self, args):
             jobs = self
         try:
             try:
-                operation_function = self._operation_functions[args.operation]
+                operation_function = self._operations[args.operation]
             except KeyError:
                 operation = self._operations[args.operation]
```
b-butler, line 3493: There is no need for the `try` `except` block any more.

kidrahahjo: @b-butler I removed the block.

kidrahahjo:
@b-butler @atravitz @lyrivera 
I am not sure why this check is there, should I remove it?
https://github.com/glotzerlab/signac-flow/pull/326/files#diff-896014ecae7b1581f5b837d124cdb4c9R3502

flow/project.py
```diff
@@ -2254,8 +2256,8 @@ def _execute_operation(self, operation, timeout=None, pretend=False):
             operation.directives.get('fork', False)
             # Separate process needed to cancel with timeout:
             or timeout is not None
-            # The operation function is not registered with the class:
-            or operation.name not in self._operation_functions
+            # The operation function is of an instace of FlowCmdOperation:
```
atravitz, line 2259:
```suggestion
            # The operation function is of an instance of FlowCmdOperation:
```

vyasr:
> @vyasr this doesn't remove the instance check. Since we have two different variable names `_cmd` and `_op_func`, we can either use a try-except block or do the instance check. I think the instance check is more meaningful here. Please suggest.

That's fine. This at least removes the indirection required to get operation functions from the project. Let's keep these types of PRs small, then you'll be able to get the merged faster :)

PR #333, opened 2020-08-03
Add pre-commit hook

kidrahahjo:
## Description

Enabling the support of [pre-commit](https://pre-commit.com/) for pre-commit hook.
Developers will now have to perform these steps in order to enable the pre-commit hook.
```
pip install -r requirements-precommit.txt
pre-commit install
```
This will add the hook to `git`.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


mikemhenry:
See https://github.com/glotzerlab/signac-docs/pull/92 for how to setup pre-commit

requirements-dev.txt
```diff
@@ -1,3 +1,4 @@
 click>=7.0
 ruamel.yaml>=0.15.89
 flake8>=3.7.7
```
jennyfothergill, line 3: Why do we pin two different flake8 versions in requirements-precommit and requirements-dev?

kidrahahjo: This is not meant to be in this file.
I'll remove this.

kidrahahjo:
@jennyfothergill I agree but the developers use `flake8` only before committing. Hence, I think we should keep all the `pre-checks` dependencies in a separate requirements file.
I proposed this for signac [here](https://github.com/glotzerlab/signac/pull/358#issuecomment-662323328) and it was implemented in this manner by @bdice for signac.

bdice:
@jennyfothergill Good question! The reason we split it is so that our CI doesn't have to install all the dependencies just to run the pre-checks. It speeds things up.

PR #334, opened 2020-08-04
Change the submission ID for aggregate logic

kidrahahjo:
## Description

Changes submission id for aggregation support

## Motivation and Context

This should get merged after #324

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:


kidrahahjo:
I merged `kidrahahjo:enable-submission-aggregate-logic` to `glotzerlab:feature/enable-submission-aggregate-logic`.
This contains a lot of extra `diff` from #324 hence I will comment on what should get reviewed in this PR.

flow/project.py
```diff
@@ -881,25 +1051,32 @@ def _generate_id(self, job, operation_name=None, index=0):
...
-                    op_string=op_string[:12],
-                    index=index)[:max_len]
+        readable_name = '{project}{sep}{jobs}{sep}{aggregate_size}{sep}{group}' \
+                        '{sep}{index:04d}{sep}'.format(
+                            sep=separator,
+                            project=str(project)[:12],
+                            jobs=aggregate_id,
+                            aggregate_size=len(jobs),
+                            group=self.name,
+                            index=index)[:max_len]
```
kidrahahjo, line 1067: Only this should get reviewed in this pull request.

kidrahahjo: I will change the template reference data after this will get reviewed

kidrahahjo:
@b-butler @csadorf 
Please verify the way we print a `_JobOperation` instance and the submission id.
After getting an approval for the changes I made I will go ahead and change the template reference data accordingly

flow/project.py
```diff
@@ -1048,12 +1049,8 @@ def _generate_id(self, jobs, operation_name=None, index=0):
             concat_jobs_str = str(jobs[0])[0:8]
 
         separator = getattr(project._environment, 'JOB_ID_SEPARATOR', '/')
-        readable_name = '{project}{sep}{jobs}{sep}{op_string}{sep}{index:04d}{sep}'.format(
-                    sep=separator,
-                    project=str(project)[:12],
-                    jobs=concat_jobs_str,
-                    op_string=op_string[:12],
-                    index=index)[:max_len]
+        readable_name = f'{str(project)[:12]}{separator}{concat_jobs_str}{separator}' \
```
b-butler, line 1052: We print the length first in `_JobOperation.__str__`. I think that order makes sense here too.

flow/project.py
```diff
@@ -1048,12 +1049,8 @@ def _generate_id(self, jobs, operation_name=None, index=0):
...
+                        f'{len(jobs)}{separator}{self.name}{separator}{index:04d}{separator}'
```
b-butler, line 1053: This needs to be restricted to a length of `max_len`.

b-butler:
@kidrahahjo the logic for id generation is fine for our purposes.

b-butler:
Also, it seems that the other PRs are being merged into each other, but is there a base for all these PR's currently (one that they will all eventually merge into before master)?

kidrahahjo:
@b-butler I like the idea of merging this into pseudo master (feature/enable-aggregation). 
We could also merge #335 into #334 and later merge #336 into this PR. At last, merging this into master. But I think this will lead to confusion.

flow/project.py
```diff
@@ -301,8 +301,9 @@ def __init__(self, id, name, jobs, cmd, directives=None):
         self.directives._keys_set_by_user = keys_set_by_user
 
     def __str__(self):
-        assert len(self._jobs) == 1
-        return f"{self.name}({str(self._jobs[0])})"
+        concat_jobs_str = str(self._jobs[0]) if len(self._jobs) == 1 else \
+            f'({str(self._jobs[0])}-{str(self._jobs[-1])})'
+        return f"{self.name}-{len(self._jobs)}: ({concat_jobs_str})"
```
csadorf, line 306:
```suggestion
        assert len(self.jobs) > 0
        max_len = 3
        shown = self.jobs[:max_len-2] + ['...'] + self.jobs[-1:] if len(self.jobs) > max_len else self.jobs
        return f"{self.name}[#{len(self.jobs)}]({', '.join([str(element)[:8] for element in shown])})"
```
I do not like this representation, because it is very unusual within the Python ecosystem. I would very much prefer to follow standard conventions here and use a representation that is inspired by the representation of large numpy arrays.
```
my-op[#1](26021048)
my-op[#2](26021048, 42b7b4f2)
my-op[#3](26021048, 42b7b4f2, 44550aef)
my-op[#4](26021048, ..., 4b893796)
```
We could even consider to omit the length indication (`#X`), since a user could always use `print(len(op))` to obtain that information. I am not sure how helpful that is in practice.

---
Code to generate above output:
```python
from dataclasses import dataclass
from typing import List

import signac

@dataclass
class AgOp:
    name: str
    jobs: List[signac.contrib.job.Job]

    def __str__(self):
        assert len(self.jobs) > 0
        max_len = 3
        shown = self.jobs[:max_len-2] + ['...'] + self.jobs[-1:] if len(self.jobs) > max_len else self.jobs
        return f"{self.name}[#{len(self.jobs)}]({', '.join([str(element)[:8] for element in shown])})"

jobs = list(signac.get_project())
for i in range(min(4, len(jobs))):
    print(AgOp(name='my-op', jobs=jobs[:i+1]))
```

flow/project.py
```diff
@@ -1048,12 +1049,9 @@ def _generate_id(self, jobs, operation_name=None, index=0):
...
         separator = getattr(project._environment, 'JOB_ID_SEPARATOR', '/')
-        readable_name = '{project}{sep}{jobs}{sep}{op_string}{sep}{index:04d}{sep}'.format(
-                    sep=separator,
-                    project=str(project)[:12],
-                    jobs=concat_jobs_str,
-                    op_string=op_string[:12],
-                    index=index)[:max_len]
+        readable_name = f'{str(project)[:12]}{separator}{self.name}{separator}' \
+                        f'{len(jobs)}{separator}{concat_jobs_str}{separator}{index:04d}' \
+                        f'{separator}'[:max_len]
```
csadorf, line 1054: I've stared at this for a while and I believe this should be fine. It would have been helpful to add a sample of this to the review comments.

kidrahahjo: @csadorf yes, I should have posted a sample so that it becomes easy for a review. I apologise.

flow/project.py
kidrahahjo: @csadorf we could drop the length part but since `_JobOperation` is private, I'm not sure how a user would know the length of it's instance

csadorf: I don't follow the logic. If this class is private why would any user care about `__str__` then?

flow/project.py
kidrahahjo: Also, if you're talking about the jobs passed in the operation function then yes I agree that users could do that but if we provide a little information from our side then I guess it'll save an extra print statement from the users. Please suggest.

flow/project.py
kidrahahjo: Also, if you're talking about the jobs passed in the operation function then yes I agree that users could do that but if we provide a little information from our side then I guess it'll save an extra print statement from the users. Please suggest.

Regarding the `__str__` method, this output is printed out in the scripts

csadorf: Ok, so then let's keep the length indicator. Do you like my proposed representation otherwise?

flow/project.py
csadorf: Sorry, I should have phrased that differently: It's always helpful to provide some samples etc., but absolutely no need to apologize. 👍

flow/project.py
kidrahahjo: Also I think we should print the full job ids in this case and not limit it to 8 characters. I don't think the output will be too long for 3 jobs.
It's not necessarily unique. @bdice once suggested me to use `min_len_unique_id` method. I could use that method if you agree.

csadorf: I think using `min_len_unique_id` would be ideal.

vyasr:
> I only have minor comments as this is a small PR. This is a breaking change in that the templates will change since we are changing the id generation (not in a way that would break unique ids). If anyone notices, I would be surprised. I would note, however, that if someone upgraded signac on master while jobs were submitted they would be unable to query them or prevent re submission at that time.
> 
> I would like to see this and future PR's merged to a psuedo-master (we could call it next or main-aggregation or something else). This would let us merge aggregation in all at once without having to review large PR's. All reviews would still be the same size until we are happy with the other branch's completeness and merge into master.
> 
> I can see an argument for this one to be merged into master though given its size, but I think the idea is worth considering.

@b-butler we are using a long-lived feature branch for Vishav's GSoC work on data structures. I had proposed using that structure for both GSoC projects, but I guess I never actually created that for @kidrahahjo's signac-flow project. I would also support that here, although up to now most of the changes have been things that IMO we intentionally wanted to push into master to avoid too many merge conflicts. I would keep that consideration in mind when deciding where to put changes (if you do decide to go the feature branch route).

PR #342, opened 2020-08-17
Modify node counting from resource sets for serial execution.

vyasr:
## Description

Bundling multiple operations into a script on Summit now uses the maximum number of nodes required for any operation rather than the sum when not parallelizing.

## Motivation and Context

The current behavior overestimates node usage because it assumes all operations must run in parallel.

I have tested this logic manually and it works well. I have not attempted to update the template tests because it would require adding sufficient operations within a single bundle to exceed a node on Summit, which would require changing the FlowProject used for testing and therefore alter all the template tests.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:


vyasr:
Explicitly requesting @bdice for review since he is the primary author and the other primary user that I am aware of.

bdice (approved):
@vyasr This looks right to me. I trust your testing is correct.

mikemhenry (approved):
LGTM!

PR #335, opened 2020-08-04
Add default aggregate support to flow

kidrahahjo:
## Description

This pull request adds the support of aggregator class in `project.py`

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change<sup>1</sup>

<sup>1</sup>The change breaks (or has the potential to break) existing functionality.

## Checklist:

- [x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac-flow/blob/master/ContributorAgreement.md).
- [x] My name is on the [list of contributors](https://github.com/glotzerlab/signac-flow/blob/master/contributors.yaml).
- [x] My code follows the [code style guideline](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md#code-style) of this project.
- [x] The changes introduced by this pull request are covered by existing or newly introduced tests.

If necessary:
- [ ] I have updated the API documentation as part of the package doc-strings.
- [ ] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
- [ ] I have updated the [changelog](https://github.com/glotzerlab/signac-flow/blob/master/changelog.txt).

PR #348, opened 2020-08-26
Add aggregator classes in flow

kidrahahjo:
## Description

Add aggregator classes to flow.
This pull request introduces a concept of storing the aggregates in an iterable class which will be then stored throughout the `FlowProject`
## Motivation and Context

To be merged before #335 

## Types of Changes

- [x] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/aggregates.py
+++ b/flow/aggregates.py
@@ -310,10 +310,6 @@ class _AggregatesStore:
     def __init__(self, aggregator, project):
         self._aggregator = aggregator
 
-        # We need to create a _MakeAggregate instance associated with this object
-        # in order to generate aggregates whenever necessary.
-        self._make_aggregates = _MakeAggregates.initialize_with_aggregator(aggregator)
-
         # We need to register the aggregates for this instance using the
         # project provided.
         self._aggregates = list()
@@ -358,12 +354,20 @@ class _AggregatesStore:
         """If the instance of this class is called then we will
         generate aggregates and store them in ``self._aggregates``.
         """
-        # Since we should always clear the aggregates before registering
-        # aggregates because we append the aggregates directly.
-        self._aggregates.clear()
-        self._aggregate_ids.clear()
-        aggregated_jobs = self._make_aggregates(project)
+        aggregated_jobs = self._generate_aggregates(project)
         self._create_nested_aggregate_list(aggregated_jobs, project)
+
+    def _generate_aggregates(self, project):
+        jobs = project
+        if self._aggregator._select is not None:
+            jobs = filter(self._aggregator._select, jobs)
+        if self._aggregator._sort_by is None:
+            jobs = list(jobs)
+        else:
+            jobs = sorted(jobs,
+                          key=lambda job: job.sp[self._aggregator._sort_by],
+                          reverse=self._aggregator._reverse_order)
+        yield from self._aggregator._aggregator_function(jobs)
 
     def _create_nested_aggregate_list(self, aggregated_jobs, project):
         """signac-flow internally assumes every aggregate to be a tuple of jobs.
@@ -445,59 +449,6 @@ class _DefaultAggregateStore:
         order to return an aggregates of one.
         """
         self._project = project
-
-
-class _MakeAggregates:
-    """This class acts as a funnctor which handles the creation of
-    aggregates using all the jobs in a project.
-
-    :param aggregator_function:
-        The ``aggregator_function`` attribute associated with the
-        :py:class:`aggregator`
-    :type aggregator_function:
-        callable
-    :param sort_by:
-        The ``sort_by`` attribute associated with the :py:class:`aggregator`
-    :type sort_by:
-        str or NoneType
-    :param reverse_order:
-        The ``reverse_order`` attribute associated with the :py:class:`aggregator`
-    :type reverse_order:
-        bool
-    :param select:
-        The ``select`` attribute associated with the :py:class:`aggregator`
-    :type select:
-        callable or NoneType
-    """
-    def __init__(self, aggregator_function, sort_by, reverse_order, select):
-        self._aggregator_function = aggregator_function
-        self._sort_by = sort_by
-        self._reverse_order = bool(reverse_order)
-        self._select = select
-
-    def __call__(self, project):
-        jobs = project
-        if self._select is not None:
-            jobs = filter(self._select, jobs)
-        if self._sort_by is None:
-            jobs = list(jobs)
-        else:
-            jobs = sorted(jobs,
-                          key=lambda job: job.sp[self._sort_by],
-                          reverse=self._reverse_order)
-        yield from self._aggregator_function(jobs)
-
-    @classmethod
-    def initialize_with_aggregator(cls, aggregator):
-        """Initialize this class using an aggregator object.
-
-        :param aggregator:
-            aggregator object associated with this class.
-        :type aggregator:
-            :py:class:`aggregator`
-        """
-        return cls(aggregator._aggregator_function, aggregator._sort_by,
-                   aggregator._reverse_order, aggregator._select)
 
 
 def get_aggregate_id(jobs):
--- a/flow/aggregates.py
+++ b/flow/aggregates.py
@@ -90,7 +90,7 @@ class aggregator:
         For instance, if 10 jobs are present in a project and they are aggregated in
         groups of 3, then the generated aggregates will have lengths 3, 3, 3, and 1.
 
-        The below code-block provides an example on how jobs can be aggregated in
+        The code block below provides an example on how jobs can be aggregated in
         groups of 2.
 
         .. code-block:: python
--- a/flow/aggregates.py
+++ b/flow/aggregates.py
@@ -391,7 +391,7 @@ class _DefaultAggregateStore:
         try:
             return (self._project.open_job(id=id),)
         except KeyError:
-            raise LookupError(f"Did not find job with id {id}.")
+            raise LookupError(f"Did not find aggregate with id {id}.")
 
     def __contains__(self, aggregate):
         """Return whether the job is present in the project associated with this
--- a/flow/aggregates.py
+++ b/flow/aggregates.py
@@ -2,7 +2,7 @@
 # All rights reserved.
 # This software is licensed under the BSD 3-Clause License.
 from collections.abc import Iterable
-from hashlib import sha1
+from hashlib import md5
 from itertools import groupby
 from itertools import zip_longest
 
@@ -20,10 +20,9 @@ class aggregator:
             print(len(jobs))
 
     :param aggregator_function:
-        Information describing how to aggregate jobs. Is a callable that
-        takes in a list of jobs and can return or yield subsets of jobs as
-        an iterable. The default behavior is creating a single aggregate
-        of all jobs.
+        A callable that performs aggregation of jobs. It takes in a list of
+        jobs and can return or yield subsets of jobs as an iterable. The
+        default behavior is creating a single aggregate of all jobs.
     :type aggregator_function:
         callable
     :param sort_by:
@@ -38,13 +37,15 @@ class aggregator:
         bool
     :param select:
         Condition for filtering individual jobs. This is passed as the
-        callable argument to `filter`.
-        The default behavior is no filtering.
+        callable argument to `filter`. The default behavior is no filtering.
     :type select:
         callable or NoneType
     :param tag:
         Additional information about the aggregator object which is a factor of
-        difference between two aggregator objects.
+        difference between two aggregator objects. signac-flow uses byte code
+        to compare callables like `aggregator_function` and `select`. Since a
+        byte code cannot truly differentiate between callables hence the tag,
+        if provided, becomes a factor of differentiation.
     :type tag:
         str or NoneType
     """
@@ -83,11 +84,11 @@ class aggregator:
 
         By default aggregate of a single job is created.
 
-        If the number of jobs present in the project can't be divided by the number
-        provided by the user then the last aggregate will have the maximum possible
-        jobs in it.
-        For instace, if 10 jobs are present in a project and they are aggregated in groups
-        of 3 then the length of aggregates would be 3, 3, 3, 1.
+        If the number of jobs present in the project is not divisible by the number
+        provided by the user, the last aggregate will be smaller and contain all
+        remaining jobs.
+        For instance, if 10 jobs are present in a project and they are aggregated in
+        groups of 3, then the generated aggregates will have lengths 3, 3, 3, and 1.
 
         The below code-block provides an example on how jobs can be aggregated in
         groups of 2.
@@ -129,6 +130,8 @@ class aggregator:
         except TypeError:
             raise TypeError('The num parameter should be an integer')
 
+        # The idea of this method was originally implemented in #52 which was copied from:
+        # https://docs.python.org/3/library/itertools.html#itertools.zip_longest
         def aggregator_function(jobs):
             args = [iter(jobs)] * num
             return zip_longest(*args)
@@ -145,7 +148,7 @@ class aggregator:
     def groupby(cls, key, default=None, sort_by=None, reverse_order=False, select=None):
         """Aggregates jobs according to matching state point key values.
 
-        The below code-block provides an example on how to aggregate jobs having
+        The below code block provides an example of how to aggregate jobs having a
         common state point parameter 'sp' whose value, when not found, is replaced by a
         default value of -1.
 
@@ -157,8 +160,10 @@ class aggregator:
                 print(len(jobs))
 
         :param key:
-            Parameter which specifies how to group jobs based on a single
-            or a list of state point parameters or a custom callable.
+            The method by which jobs are grouped. It may be a state point
+            or a sequence of state points to group by specific state point
+            keys. It may also be an arbitrary callable of :class:`signac.Job`
+            when greater flexibility is needed.
         :type key:
             str, Iterable, or callable
         :param default:
@@ -221,15 +226,15 @@ class aggregator:
     def __eq__(self, other):
         if type(self) != type(other):
             return False
-        elif(
-            self._sort_by != other._sort_by or
-            self._reverse_order != other._reverse_order or
-            self._tag != other._tag
+        elif (
+             self._sort_by != other._sort_by or
+             self._reverse_order != other._reverse_order or
+             self._tag != other._tag
         ):
             return False
-        elif(
-            self._select == other._select and
-            self._aggregator_function == other._aggregator_function
+        elif (
+             self._select == other._select and
+             self._aggregator_function == other._aggregator_function
         ):
             return True
 
@@ -251,7 +256,7 @@ class aggregator:
         ]
         blob = ','.join(str(attr) for attr in blob_l)
 
-        return int(sha1(blob.encode('utf-8')).hexdigest(), 16)
+        return id(_hash(blob))
 
     def _get_unique_function_id(self, func):
         """Generate unique id for the function passed. The id returned is used to generate
@@ -264,10 +269,13 @@ class aggregator:
             return hash(func)
 
     def _create_AggregatesStore(self, project):
-        """Create the instance of _AggregatesStore or _DefaultAggregateStore classes
-        to make sure that we don't explicitly store the aggregates in the FlowProject.
+        """Create the actual collections of jobs to be sent to aggregate operations.
 
-        When iterated over these classes, an aggregate is yielded.
+        The :class:`aggregate` class is just a decorator that provides a signal for
+        operation functions that should be treated as aggregate operations and
+        information on how to perform the aggregation. This function generates
+        the classes that actually hold sequences of jobs to which aggregate
+        operations will be applied.
 
         :param project:
             A signac project used to fetch jobs for creating aggregates.
@@ -291,7 +299,7 @@ class aggregator:
 
 class _AggregatesStore:
     """This class holds the information of all the aggregates associated with
-    a :py:class:`aggregator`.
+    a :class:`aggregator`.
 
     This is a callable class which, when called, generates all the aggregates.
     When iterated through it's instance, all the aggregates are yielded.
@@ -345,7 +353,7 @@ class _AggregatesStore:
 
     def __hash__(self):
         blob = str(hash(self._aggregator))
-        return int(sha1(blob.encode('utf-8')).hexdigest(), 16)
+        return id(_hash(blob))
 
     def _register_aggregates(self, project):
         """If the instance of this class is called then we will
@@ -438,7 +446,7 @@ class _DefaultAggregateStore:
 
     def __hash__(self):
         blob = self._project.get_id()
-        return int(sha1(blob.encode('utf-8')).hexdigest(), 16)
+        return id(_hash(blob))
 
     def _register_aggregates(self, project):
         """We have to store self._project when this method is invoked
@@ -460,4 +468,8 @@ def get_aggregate_id(jobs):
         return jobs[0].get_id()  # Return job id as it's already unique
 
     blob = ','.join((job.get_id() for job in jobs))
-    return f'agg-{sha1(blob.encode()).hexdigest()}'
+    return f'agg-{_hash(blob)}'
+
+
+def _hash(blob):
+    return md5(blob.encode()).hexdigest()
--- a/flow/aggregates.py
+++ b/flow/aggregates.py
@@ -9,14 +9,13 @@ from itertools import zip_longest
 
 
 class Aggregate:
-    """Decorator for operation functions that are to be aggregated.
+    r"""Decorator for operation functions that are to be aggregated.
     By default, if the aggregator parameter is not passed,
     an aggregate of all jobs will be created.
 
     .. code-block:: python
 
-        example_aggregate = Aggregate()
-        @example_aggregate
+        @Aggregate()
         @FlowProject.operation
         def foo(*jobs):
             print(len(jobs))
@@ -44,27 +43,33 @@ class Aggregate:
         The default behavior is no filtering.
     :type select:
         callable or NoneType
+    :param \*\*kwargs:
+        Additional information related to aggregator function.
     """
 
-    def __init__(self, aggregator=None, sort_by=None, reverse_order=False, select=None):
+    def __init__(self, aggregator=None, sort_by=None, reverse_order=False, select=None,
+                 **kwargs):
         if aggregator is None:
             def aggregator(jobs):
                 return [jobs]
 
         if not callable(aggregator):
             raise TypeError(f"Expected callable for aggregator, got {type(aggregator)}")
-
-        if sort_by is not None and not isinstance(sort_by, str):
+        elif sort_by is not None and not isinstance(sort_by, str):
             raise TypeError(f"Expected string sort_by parameter, got {type(sort_by)}")
-
-        if select is not None and not callable(select):
+        elif select is not None and not callable(select):
             raise TypeError(f"Expected callable for select, got {type(select)}")
 
+        # For "non-aggregate" functions we set the Aggregate object equals to
+        # Aggregate.groupsof(1). If any other Aggregate object is associated with an
+        # operation function then mark that as a "aggregate operation" else it's a
+        # "non-aggregate" operation.
         if sort_by is None and select is None and not reverse_order:
             self._is_aggregate = getattr(aggregator, '_num', 0) != 1
         else:
             self._is_aggregate = True
 
+        self._kwargs = kwargs
         self._aggregator = aggregator
         self._sort_by = sort_by
         self._reverse_order = reverse_order
@@ -72,8 +77,50 @@ class Aggregate:
 
     @classmethod
     def groupsof(cls, num=1, sort_by=None, reverse_order=False, select=None):
-        # copied from: https://docs.python.org/3/library/itertools.html#itertools.zip_longest
+        """signac jobs can be aggregated in groups of a number provided by an user.
+
+        By default aggregate of a single job is created.
+
+        If the number of jobs present in the project can't be divided by the number
+        provided by the user then the last aggregate will have the maximum possible
+        jobs in it.
+        For instace, if 10 jobs are present in a project and they are aggregated in groups
+        of 3 then the length of aggregates would be 3, 3, 3, 1.
+
+        The below code-block provides an example on how jobs can be aggregated in
+        groups of 2.
+
+        .. code-block:: python
+
+            @Aggregate.groupsof(num=2)
+            @FlowProject.operation
+            def foo(*jobs):
+                print(len(jobs))
+
+        :param num:
+            Maximum possible length of an aggregate.
+        :type num:
+            int
+        :param sort_by:
+            Before aggregating, sort the jobs by a given statepoint parameter.
+            The default behavior is no sorting.
+        :type sort_by:
+            str or NoneType
+        :param reverse_order:
+            States if the jobs are to be sorted in reverse order.
+            The default value is False.
+        :type reverse_order:
+            bool
+        :param select:
+            Condition for filtering individual jobs. This is passed as the
+            callable argument to `filter`.
+            The default behavior is no filtering.
+        :type select:
+            callable or NoneType
+        """
         try:
+            if num != int(num):
+                raise ValueError('The num parameter should be an integer')
             num = int(num)
             if num <= 0:
                 raise ValueError('The num parameter should have a value greater than 0')
@@ -86,10 +133,50 @@ class Aggregate:
 
         setattr(aggregator, '_num', num)
 
-        return cls(aggregator, sort_by, reverse_order, select)
+        return cls(aggregator, sort_by, reverse_order, select, num=num)
 
     @classmethod
     def groupby(cls, key, default=None, sort_by=None, reverse_order=False, select=None):
+        """signac jobs can be aggregated by a valid state point parameter.
+
+        Users can aggregate jobs by passing a single or a list of state point parameters
+        or a custom callable.
+        If a user provides a list of state point parameters then the default value, if passed,
+        expects a list of default values having length equal to the length of key passed.
+        By default, an error is raised if the key passed is invalid.
+
+        .. code-block:: python
+
+            @Aggregate.groupby('sp', -1)
+            @FlowProject.operation
+            def foo(*jobs):
+                print(len(jobs))
+
+        :param key:
+            Parameter which specifies how the jobs should be aggregated.
+        :type key:
+            str, Iterable, or callable
+        :param key:
+            Default values used for grouping if invalid key is passed
+        :type key:
+            str, Iterable, or callable
+        :param sort_by:
+            Before aggregating, sort the jobs by a given statepoint parameter.
+            The default behavior is no sorting.
+        :type sort_by:
+            str or NoneType
+        :param reverse_order:
+            States if the jobs are to be sorted in reverse order.
+            The default value is False.
+        :type reverse_order:
+            bool
+        :param select:
+            Condition for filtering individual jobs. This is passed as the
+            callable argument to `filter`.
+            The default behavior is no filtering.
+        :type select:
+            callable or NoneType
+        """
         if isinstance(key, str):
             if default is None:
                 def keyfunction(job):
@@ -97,10 +184,8 @@ class Aggregate:
             else:
                 def keyfunction(job):
                     return job.sp.get(key, default)
-
         elif isinstance(key, Iterable):
             keys = list(key)
-
             if default is None:
                 def keyfunction(job):
                     return [job.sp[key] for key in keys]
@@ -116,10 +201,8 @@ class Aggregate:
                 def keyfunction(job):
                     return [job.sp.get(key, default_value)
                             for key, default_value in zip(keys, default)]
-
         elif callable(key):
             keyfunction = key
-
         else:
             raise TypeError("Invalid key argument. Expected either str, Iterable "
                             f"or a callable, got {type(key)}")
@@ -128,27 +211,66 @@ class Aggregate:
             for key, group in groupby(sorted(jobs, key=keyfunction), key=keyfunction):
                 yield group
 
-        return cls(aggregator, sort_by, reverse_order, select)
+        return cls(aggregator, sort_by, reverse_order, select, key=key, default=default)
+
+    def __eq__(self, other):
+        if type(self) != type(other):
+            return False
+        elif(
+            self._sort_by != other._sort_by or
+            self._reverse_order != other._reverse_order or
+            self._kwargs != other._kwargs
+        ):
+            return False
+        elif(
+            self._select == other._select and
+            self._aggregator == other._aggregator
+        ):
+            return True
+
+        # Get unique id for _select attribute
+        self_select = self._get_unique_function_id(self._select)
+        other_select = self._get_unique_function_id(other._select)
+        # Get unique id for _aggregator attribute
+        self_aggregator = self._get_unique_function_id(self._aggregator)
+        other_aggregator = self._get_unique_function_id(other._aggregator)
+
+        if(
+            self_select == other_select and
+            self_aggregator == other_aggregator
+        ):
+            return True
+
+        return False
 
     def __hash__(self):
-        blob_l = map(hash, [self._aggregator, self._sort_by,
-                            self._reverse_order, self._select])
+        blob_l = list(map(hash, [self._sort_by, self._reverse_order]))
+        # Get unique id for _aggregator and _select attributes
+        blob_l += list(map(self._get_unique_function_id, [self._aggregator, self._select]))
         blob = ','.join(str(attr) for attr in blob_l)
 
         return int(sha1(blob.encode('utf-8')).hexdigest(), 16)
 
-    def _create_StoreAggregates(self):
-        # We create the instance of _StoreAggregates or _StoreDefaultAggregates classes
-        # to make sure that we don't explicitly store the aggregates in the FlowProject.
-        # When iterated over these classes, an aggregate is yielded.
+    def _get_unique_function_id(self, func):
+        try:
+            return func.__code__.co_code
+        except Exception:  # Got partial function
+            return str(hash(func))
+
+    def _create_AggregatesStore(self, jobs):
+        """Create the instance of _AggregatesStore or _DefaultAggregateStore classes
+        to make sure that we don't explicitly store the aggregates in the FlowProject.
+
+        When iterated over these classes, an aggregate is yielded.
+        """
         if not self._is_aggregate:
-            return _StoreDefaultAggregate()
+            return _DefaultAggregateStore(jobs)
         else:
-            return _StoreAggregates(self)
+            return _AggregatesStore(self, jobs)
 
     def __call__(self, func=None):
         if callable(func):
-            setattr(func, '_flow_aggregate', self._create_StoreAggregates())
+            setattr(func, '_flow_aggregate', self)
             return func
         else:
             raise TypeError('Invalid argument passed while calling '
@@ -156,7 +278,7 @@ class Aggregate:
                             f'got {type(func)}.')
 
 
-class _StoreAggregates:
+class _AggregatesStore:
     """This class holds the information of all the aggregates associated with
     a :py:class:`Aggregate`.
 
@@ -167,17 +289,61 @@ class _StoreAggregates:
         Aggregate object associated with this class.
     :type aggregate:
         :py:class:`Aggregate`
+    :param jobs:
+        The signac job handles
+    :type jobs:
+        Iterator
     """
-    def __init__(self, aggregate):
+    def __init__(self, aggregate, jobs):
         self._aggregate = aggregate
+
+        # We need to register the aggregates for this instance using the
+        # jobs provided.
         self._aggregates = list()
+        self._aggregate_ids = dict()
+        self._register_aggregates(jobs)
 
     def __iter__(self):
         yield from self._aggregates
 
-    def __call__(self, jobs):
-        # If the instance of this class is called then we will
-        # generate aggregates and store them in self._aggregates
+    def __getitem__(self, id):
+        "Return an aggregate "
+        try:
+            return self._aggregate_ids[id]
+        except KeyError:
+            raise LookupError(f'Unable to find the aggregate having id {id} in '
+                              'the FlowProject')
+
+    def __contains__(self, aggregate):
+        """Return whether an aggregate is stored in the this
+        instance of :py:class:`_AggregateStore`
+
+        :param aggregate:
+            Aggregate of jobs
+        :type jobs:
+            tuple of :py:class:`signac.contrib.job.Job`
+        """
+        if aggregate in self._aggregates:
+            return True
+        else:
+            return False
+
+    def __len__(self):
+        return len(self._aggregates)
+
+    def __eq__(self, other):
+        return type(self) == type(other) and \
+               self._aggregates == other._aggregates and \
+               self._aggregate == other._aggregate
+
+    def __hash__(self):
+        blob = str(hash(self._aggregate))
+        return int(sha1(blob.encode('utf-8')).hexdigest(), 16)
+
+    def _register_aggregates(self, jobs):
+        """If the instance of this class is called then we will
+        generate aggregates and store them in self._aggregates
+        """
         if isinstance(jobs, list):  # Got a tuple of jobs
             project = jobs[0]._project
         else:  # Got a project
@@ -190,14 +356,13 @@ class _StoreAggregates:
             aggregated_jobs = list(sorted(aggregated_jobs,
                                           key=lambda job: job.sp[self._aggregate._sort_by],
                                           reverse=bool(self._aggregate._reverse_order)))
-
         aggregated_jobs = self._aggregate._aggregator([job for job in aggregated_jobs])
         self._create_nested_aggregate_list(aggregated_jobs, project)
 
     def _create_nested_aggregate_list(self, aggregated_jobs, project):
-        # This method converts the returned subset of jobs as an Iterable
-        # from an aggregator function to a subset of jobs as tuple.
-        aggregated_jobs = list(aggregated_jobs)
+        """This method converts the returned subset of jobs as an Iterable
+        from an aggregator function to a subset of jobs as tuple.
+        """
         nested_aggregates = []
 
         for aggregate in aggregated_jobs:
@@ -207,48 +372,80 @@ class _StoreAggregates:
                     if job is None:
                         continue
                     elif job not in project:
-                        raise LookupError(f'The signac job {str(job)} not found in {project}')
+                        raise LookupError(f'The signac job {job.get_id()} not found in {project}')
                     filter_aggregate.append(job)
                 filter_aggregate = tuple(filter_aggregate)
+                # Store aggregate in this instance
                 nested_aggregates.append(filter_aggregate)
+                # Store aggregate by their ids in order to search through id
+                self._aggregate_ids[get_aggregate_id(filter_aggregate)] = filter_aggregate
             except TypeError:  # aggregate is not iterable
                 raise ValueError("Invalid aggregator function provided by "
                                  "the user.")
 
         self._aggregates = nested_aggregates
 
-    def __hash__(self):
-        blob = str(hash(self._aggregate))
-        return int(sha1(blob.encode('utf-8')).hexdigest(), 16)
 
-
-class _StoreDefaultAggregate:
+class _DefaultAggregateStore:
     """This class holds the information of the project associated with
     an operation function aggregated by the default aggregates i.e.
     :py:class:`Aggregate.groupsof(1)`.
 
     When iterated through it's instance, it yields a tuple of a single job from
     the Project.
+
+    :param jobs:
+        The signac job handles
+    :type jobs:
+        Iterator
     """
-    def __init__(self):
+    def __init__(self, jobs):
         self._project = None
+        self._register_project(jobs)
 
     def __iter__(self):
         for job in self._project:
             yield (job,)
 
-    def __call__(self, jobs):
-        # We have to store self._project is a instance of this class
-        # is called. This is because we will then iterate over that
-        # project in order to return an aggregates of one.
-        if isinstance(jobs, list):  # Got a tuple of jobs
-            self._project = jobs[0]._project
-        else:  # Got a project
-            self._project = jobs
+    def __getitem__(self, id):
+        "Return a tuple of a single job via job id"
+        try:
+            return (self._project.open_job(id=id),)
+        except KeyError:
+            raise LookupError(f"Did not find job with id {id}.")
+
+    def __contains__(self, job):
+        """Return whether the job is present in the project associated with this
+        instance of :py:class:`_DefaultAggregateStore`
+        """
+        if job in self._project:
+            return True
+        else:
+            return False
+
+    def __len__(self):
+        return len(self._project)
+
+    def __eq__(self, other):
+        return type(self) == type(other) and \
+               self._project == other._project
 
     def __hash__(self):
         blob = str(self._project)
         return int(sha1(blob.encode('utf-8')).hexdigest(), 16)
+
+    def _generate_aggregates(self, jobs):
+        """We have to store self._project when this method is invoked
+        This is because we will then iterate over that project in
+        order to return an aggregates of one.
+        """
+        self._register_project(jobs)
+
+    def _register_project(self, jobs):
+        if isinstance(jobs, list):  # Got a list of jobs
+            self._project = jobs[0]._project
+        else:  # Got a project
+            self._project = jobs
 
 
 def get_aggregate_id(jobs):
@@ -260,7 +457,7 @@ def get_aggregate_id(jobs):
         tuple
     """
     if len(jobs) == 1:
-        return str(jobs[0])  # Return job id as it's already unique
+        return jobs[0].get_id()  # Return job id as it's already unique
 
     blob = ','.join((job.get_id() for job in jobs))
     return f'agg-{md5(blob.encode()).hexdigest()}'
--- a/flow/aggregates.py
+++ b/flow/aggregates.py
@@ -430,7 +430,7 @@ class _DefaultAggregateStore:
                self._project == other._project
 
     def __hash__(self):
-        blob = self._project.id
+        blob = self._project.get_id()
         return int(sha1(blob.encode('utf-8')).hexdigest(), 16)
 
     def _register_aggregates(self, project):
--- a/flow/aggregates.py
+++ b/flow/aggregates.py
@@ -78,8 +78,7 @@ class aggregator:
 
     @classmethod
     def groupsof(cls, num=1, sort_by=None, reverse_order=False, select=None):
-        """Decorator for operation function that aggregates signac jobs in groups of
-        a number provided by the user.
+        """Aggregates jobs of a set group size.
 
         By default aggregate of a single job is created.
 
@@ -139,8 +138,7 @@ class aggregator:
 
     @classmethod
     def groupby(cls, key, default=None, sort_by=None, reverse_order=False, select=None):
-        """Decorator for operation function that aggregates signac jobs by grouping them
-        via a key.
+        """Aggregates jobs according to matching state point key values.
 
         The below code-block provides an example on how to aggregate jobs having
         common state point parameter 'sp' whose value, when not found, is replaced by a
@@ -252,9 +250,9 @@ class aggregator:
         return int(sha1(blob.encode('utf-8')).hexdigest(), 16)
 
     def _get_unique_function_id(self, func):
-        """Generate unique id for the function passed. This id is used to generate hash
-        and compare the ``self._aggregator_function`` and ``self._select`` attributes
-        of instances of this class.
+        """Generate unique id for the function passed. The id returned is used to generate
+        hash and compare arbitrary types which are callable like ``self._aggregator_function``
+        and ``self._select`` attributes.
         """
         try:
             return func.__code__.co_code
@@ -294,23 +292,29 @@ class _AggregatesStore:
     This is a callable class which, when called, generates all the aggregates.
     When iterated through it's instance, all the aggregates are yielded.
 
-    :param aggregate:
+    :param aggregator:
         aggregator object associated with this class.
-    :type aggregate:
+    :type aggregator:
         :py:class:`aggregator`
     :param project:
         A signac project used to fetch jobs for creating aggregates.
     :type project:
         :py:class:`flow.FlowProject` or :py:class:`signac.contrib.project.Project`
     """
-    def __init__(self, aggregate, project):
-        self._aggregate = aggregate
+    def __init__(self, aggregator, project):
+        self._aggregator = aggregator
+
+        # We need to create a _MakeAggregate instance associated with this object
+        # in order to generate aggregates whenever necessary.
+        self._make_aggregates = _MakeAggregates(
+            aggregator._aggregator_function, aggregator._sort_by,
+            aggregator._reverse_order, aggregator._select, project)
 
         # We need to register the aggregates for this instance using the
         # project provided.
         self._aggregates = list()
         self._aggregate_ids = dict()
-        self._register_aggregates(project)
+        self._register_aggregates()
 
     def __iter__(self):
         yield from self._aggregates
@@ -340,20 +344,17 @@ class _AggregatesStore:
     def __eq__(self, other):
         return type(self) == type(other) and \
                self._aggregates == other._aggregates and \
-               self._aggregate == other._aggregate
+               self._aggregator == other._aggregator
 
     def __hash__(self):
         blob = str(hash(self._aggregate))
         return int(sha1(blob.encode('utf-8')).hexdigest(), 16)
 
-    def _register_aggregates(self, project):
+    def _register_aggregates(self):
         """If the instance of this class is called then we will
         generate aggregates and store them in ``self._aggregates``.
         """
-        make_aggregates = _MakeAggregates(self._aggregate, project)
-        aggregates, aggregate_ids = make_aggregates()
-        self._aggregates = aggregates
-        self._aggregate_ids = aggregate_ids
+        self._aggregates, self._aggregate_ids = self._make_aggregates()
 
 
 class _DefaultAggregateStore:
@@ -412,29 +413,45 @@ class _MakeAggregates:
     """This class acts as a funnctor which handles the creation of
     aggregates using all the jobs in a project.
 
-    :param aggregate:
-        aggregator object associated with this class.
-    :type aggregate:
+    :param aggregator_function:
+        The ``aggregator_function`` attribute associated with the
         :py:class:`aggregator`
+    :type aggregator_function:
+        callable
+    :param sort_by:
+        The ``sort_by`` attribute associated with the :py:class:`aggregator`
+    :type sort_by:
+        str or NoneType
+    :param reverse_order:
+        The ``reverse_order`` attribute associated with the :py:class:`aggregator`
+    :type reverse_order:
+        bool
+    :param select:
+        The ``select`` attribute associated with the :py:class:`aggregator`
+    :type select:
+        callable or NoneType
     :param project:
         A signac project used to fetch jobs for creating aggregates.
     :type project:
         :py:class:`flow.FlowProject` or :py:class:`signac.contrib.project.Project`
     """
-    def __init__(self, aggregate, project):
-        self._aggregate = aggregate
+    def __init__(self, aggregator_function, sort_by, reverse_order, select, project):
+        self._aggregator_function = aggregator_function
+        self._sort_by = sort_by
+        self._reverse_order = bool(reverse_order)
+        self._select = select
         self._project = project
 
     def __call__(self):
         jobs = list(self._project)
 
-        if self._aggregate._select is not None:
-            jobs = filter(self._aggregate._select, jobs)
-        if self._aggregate._sort_by is not None:
+        if self._select is not None:
+            jobs = list(filter(self._select, jobs))
+        if self._sort_by is not None:
             jobs = sorted(jobs,
-                          key=lambda job: job.sp[self._aggregate._sort_by],
-                          reverse=self._aggregate._reverse_order)
-        aggregated_jobs = self._aggregate._aggregator_function(jobs)
+                          key=lambda job: job.sp[self._sort_by],
+                          reverse=self._reverse_order)
+        aggregated_jobs = self._aggregator_function(jobs)
         return self._create_nested_aggregate_list(aggregated_jobs)
 
     def _create_nested_aggregate_list(self, aggregated_jobs):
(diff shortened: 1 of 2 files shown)


kidrahahjo:
@b-butler I think making a functor would make the code more readable (I guess) but I think this is also fine.
Please suggest.

flow/aggregates.py
```diff
@@ -0,0 +1,266 @@
...
+        def aggregator(jobs):
+            for key, group in groupby(sorted(jobs, key=keyfunction), key=keyfunction):
+                yield group
+
+        return cls(aggregator, sort_by, reverse_order, select)
+
+    def __hash__(self):
+        blob_l = map(hash, [self._aggregator, self._sort_by,
+                            self._reverse_order, self._select])
+        blob = ','.join(str(attr) for attr in blob_l)
```
b-butler, line 136: To use an object as a key in a dictionary it has to support `__hash__` and `__eq__`

tests/test_aggregates.py
```diff
@@ -0,0 +1,264 @@
...
+        with pytest.raises(TypeError):
+            aggregate_instance()
+
+    def test_call_with_function(self):
+        aggregate_instance = Aggregate()
+
+        def test_function(x):
+            return x
+
+        assert not getattr(test_function, '_flow_aggregate', False)
```
b-butler, line 82: I would prefer using `hasattr` here.

tests/test_aggregates.py
```diff
@@ -0,0 +1,264 @@
...
+        test_function = aggregate_instance(test_function)
+        assert getattr(test_function, '_flow_aggregate', False)
```
b-butler, line 84: Likewise

tests/test_aggregates.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def test_with_decorator_with_pre_initialization(self):
+        aggregate_instance = Aggregate()
+
+        @aggregate_instance
+        def test_function(x):
+            return x
+
+        assert getattr(test_function, '_flow_aggregate', False)
```
b-butler, line 93: hasattr

b-butler, line 93: Although different in terms of syntax used; this and the previous test are identical.

tests/test_aggregates.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def test_with_decorator_without_pre_initialization(self):
+        @Aggregate()
+        def test_function(x):
+            return x
+
+        assert getattr(test_function, '_flow_aggregate', False)
```
b-butler, line 100: This is also unnecessary.

tests/test_aggregates.py
```diff
@@ -0,0 +1,264 @@
...
+        assert [aggregate for aggregate in aggregate_job_manual] == \
+               list(aggregate_instance)
+
+    def test_valid_aggregator_partial(self, setUp, project):
+        aggregate_instance = Aggregate(lambda jobs: [jobs])
+        aggregate_instance = aggregate_instance._create_StoreAggregates()
+        aggregate_instance(project)  # Generate Aggregates
+
+        assert [tuple([job for job in project])] == \
+               list(aggregate_instance)
```
b-butler, line 148:
```suggestion
        assert [tuple(project)] == list(aggregate_instance)
```

tests/test_aggregates.py
```diff
@@ -0,0 +1,264 @@
...
+        aggregate_instance = aggregate_instance._create_StoreAggregates()
+        aggregate_instance(project)  # Generate Aggregates
+        assert [tuple(helper_sort(project))] == list(aggregate_instance)
+
+    def test_valid_reversed_sort(self, setUp, project):
+        helper_sort = partial(sorted, key=lambda job: job.sp.i, reverse=True)
+        aggregate_instance = Aggregate(sort_by='i', reverse_order=True)
+        aggregate_instance = aggregate_instance._create_StoreAggregates()
+        aggregate_instance(project)  # Generate Aggregates
+        assert [tuple(helper_sort(project))] == list(aggregate_instance)
```
b-butler, line 162: It may be worth creating a helper function that does the comparison between a function's and an `_StoreAggregate`'s aggregates given a `FlowProject`.

tests/test_aggregates.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def test_groups_of_valid_num(self, setUp, project):
```
b-butler, line 164: We should also test that the size of each aggregate is correct.

tests/test_aggregates.py
```diff
@@ -0,0 +1,264 @@
...
+            assert len(list(aggregate_instance)) == length_of_aggregate
+
+    def test_groupby_with_valid_string_key(self, setUp, project):
+        aggregate_instance = Aggregate.groupby('even')
+        aggregate_instance = aggregate_instance._create_StoreAggregates()
+        aggregates = 0
+        aggregate_instance(project)  # Generate Aggregates
+        for agg in aggregate_instance:
+            aggregates += 1
+        assert aggregates == 2
```
b-butler, line 186: I would assert that all of the values for `job.sp.even` are identical.

tests/test_aggregates.py
```diff
@@ -0,0 +1,264 @@
...
+            aggregate_instance(project)
+
+    def test_groupby_with_default_key_for_string(self, setUp, project):
+        aggregate_instance = Aggregate.groupby('half', default=-1)
+        aggregate_instance = aggregate_instance._create_StoreAggregates()
+        aggregate_instance(project)  # Generate Aggregates
+        aggregates = 0
+        for agg in aggregate_instance:
+            aggregates += 1
+        assert aggregates == 6
```
b-butler, line 203: I would likewise test that all the values of `job.sp.half` are identical.

b-butler:
It is fine to have the functionality in `_AggregateStore` my preference is to have a functor class that does this, but I won't be adamant about it.

flow/aggregates.py
```diff
@@ -0,0 +1,255 @@
...
+        example_aggregate = Aggregate()
+        @example_aggregate
+        @FlowProject.operation
+        def foo(*jobs):
+            print(len(jobs))
+
+    :param aggregator:
+        Information describing how to aggregate jobs. Is a callable that
+        takes in a list of jobs and can return or yield subsets of jobs as
+        an iterable. The default behavior is creating a single aggregate
```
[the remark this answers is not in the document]
kidrahahjo: I think we should be flexible in terms of users point of view.
To convert those into tuples, we're already converting those iterables into tuples in `_create_nested_aggregate_list`

flow/aggregates.py
```diff
@@ -0,0 +1,255 @@
...
+        def aggregator(jobs):
+            for key, group in groupby(sorted(jobs, key=keyfunction), key=keyfunction):
+                yield group
+
+        return cls(aggregator, sort_by, reverse_order, select)
+
+    def __eq__(self, other):
+        return self._aggregator.__code__.co_code == other._aggregator.__code__.co_code and \
+               self._select.__code__.co_code == other._select.__code__.co_code and \
+               self._sort_by == other._sort_by and self._reverse_order == other._reverse_order
```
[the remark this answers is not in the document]
kidrahahjo: I thought we'd require an `__eq__` method to put these object as keys in a dictionary but I learnt that we require the objects to be hashable and it turns out we can uniquely hash these objects (partial functions, lambda functions, or normal functions).
I'll use`__hash__` here

flow/aggregates.py
```diff
@@ -0,0 +1,255 @@
...
+from collections.abc import Iterable
+from hashlib import md5
+from itertools import groupby
+from itertools import zip_longest
+
+from flow import FlowProject
+from signac.contrib.project import Project
+
+
+class Aggregate:
```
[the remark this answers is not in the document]
kidrahahjo: This is technically breaking the API. Should I post this on slack for further discussion?

flow/aggregates.py
```diff
@@ -0,0 +1,255 @@
...
+        # to make sure that we don't explicitly store the aggregates in the FlowProject.
+        # When iterated over these classes, an aggregate is yielded.
+        if not self._is_aggregate:
+            return _StoreDefaultAggregate()
+        else:
+            return _StoreAggregates(self)
+
+    def __call__(self, func=None):
+        if callable(func):
+            setattr(func, '_flow_aggregate', self._create_StoreAggregates())
```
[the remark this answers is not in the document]
kidrahahjo: That is why we'll be using a `__hash__` method to uniquely identify a `_StoreAggregate` instance

flow/aggregates.py
```diff
@@ -0,0 +1,255 @@
...
+                            'the aggregate instance. Expected a callable, '
+                            f'got {type(func)}.')
+
+
+class _StoreAggregates:
+    """This class holds the information of all the aggregates associated with
+    a :py:class:`Aggregate`.
+
+    This is a callable class which, when called, generates all the aggregates.
+    When iterated through it's instance, all the aggregates are yielded.
```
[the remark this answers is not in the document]
kidrahahjo: I'll use a different method `_generate_aggregates` instead

flow/aggregates.py
```diff
@@ -0,0 +1,255 @@
...
+
+    :param aggregate:
+        Aggregate object associated with this class.
+    :type aggregate:
+        :py:class:`Aggregate`
+    """
+    def __init__(self, aggregate):
+        self._aggregate = aggregate
+        self._aggregates = list()
```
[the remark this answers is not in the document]
kidrahahjo: Hmm, I see what you mean here.
I also agree, will make the necessary changes

flow/aggregates.py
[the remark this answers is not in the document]
b-butler: It is breaking an API that doesn't exist yet, so you can ask, but it isn't a breaking change.

flow/aggregates.py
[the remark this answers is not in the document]
b-butler: I guess I am fine then.

flow/aggregates.py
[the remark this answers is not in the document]
b-butler: I would caution against just using hash for both `__eq__` and `__hash__`. If `hash(obj) == hash(other)` the equality method is used to differentiate between objects. Here if we use it for both we end up with no way to discern the objects.

flow/aggregates.py
```diff
@@ -0,0 +1,266 @@
...
+                return [jobs]
+
+        if not callable(aggregator):
+            raise TypeError(f"Expected callable for aggregator, got {type(aggregator)}")
+
+        if sort_by is not None and not isinstance(sort_by, str):
+            raise TypeError(f"Expected string sort_by parameter, got {type(sort_by)}")
+
+        if select is not None and not callable(select):
+            raise TypeError(f"Expected callable for select, got {type(select)}")
```
vyasr, line 61: All of this error checking feels rather un-Pythonic to me. However, I can see a good reason for this here since none of these attributes will actually get used until downstream in the aggregation pipeline, at which point the errors would be much harder for users to interpret. @bdice you're usually the style police for EAFP (I'm of the opinion that both LBYL and EAFP have their place in Python, so I'm less opposed to this on principle), what do you think?

flow/aggregates.py
```diff
@@ -0,0 +1,266 @@
...
+                    return [job.sp.get(key, default_value)
+                            for key, default_value in zip(keys, default)]
+
+        elif callable(key):
+            keyfunction = key
+
+        else:
+            raise TypeError("Invalid key argument. Expected either str, Iterable "
+                            f"or a callable, got {type(key)}")
+
```
vyasr, line 126: I don't know if I've ever said this before, but there's actually too much white space and empty lines in this block (and in various other parts of the code) and it makes it less readable. In particular, there's a newline before every `elif` or `else` clause. Can you tighten some of that up?

flow/aggregates.py
vyasr: Just FYI @kidrahahjo both of these methods have default implementations for any class (e.g. `class A: pass` will have default versions of both methods defined), but I assume @b-butler wants you to define a meaningful equality check here.

flow/aggregates.py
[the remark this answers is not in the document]
vyasr: How will that prevent making a new one? It seems like you will create a new one anyway in `_create_StoreAggregates`.

flow/aggregates.py
```diff
@@ -0,0 +1,255 @@
...
+        if callable(func):
+            setattr(func, '_flow_aggregate', self._create_StoreAggregates())
+            return func
+        else:
+            raise TypeError('Invalid argument passed while calling '
+                            'the aggregate instance. Expected a callable, '
+                            f'got {type(func)}.')
+
+
+class _StoreAggregates:
```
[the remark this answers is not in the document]
vyasr: Can we just combine the classes? I'm not seeing much benefit to the separation. Especially if you change the `_StoreAggregates` class to compute the aggregates on construction, then both this and the `*Default` class can just store `self._aggregates` (which would just be the project for the default case) and `yield from` that in `__iter__`. Hashing should also work transparently.

flow/aggregates.py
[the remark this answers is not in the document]
vyasr: This change supports us moving away from the two class model.

flow/aggregates.py
```diff
@@ -0,0 +1,266 @@
...
+def get_aggregate_id(jobs):
+    """Generate hashed id for an aggregate of jobs
+
+    :param jobs:
+        The signac job handles
+    :type jobs:
+        tuple
+    """
+    if len(jobs) == 1:
+        return str(jobs[0])  # Return job id as it's already unique
```
vyasr, line 263: I would prefer to return `jobs[0].id`, which is strictly correct and doesn't depend on us not changing the string representation of a `Job`.

flow/aggregates.py
```diff
@@ -0,0 +1,266 @@
...
+
+    @classmethod
+    def groupsof(cls, num=1, sort_by=None, reverse_order=False, select=None):
+        # copied from: https://docs.python.org/3/library/itertools.html#itertools.zip_longest
+        try:
+            num = int(num)
+            if num <= 0:
+                raise ValueError('The num parameter should have a value greater than 0')
+        except TypeError:
+            raise TypeError('The num parameter should be an integer')
```
vyasr, line 81: @bdice another instance of type checking up front that is probably justified due to the delayed execution of aggregator.

@kidrahahjo the sequence here is a little confusing though. The `TypeError` you're catching would be thrown by the first line in the `try` block, so there's no need for the rest to be there. I would move the `if num<=0` check out of the try block. Furthermore, rather than trying to convert to int like this, if we're going to type check I would do it rigorously. Currently, passing `num=3.5` will result in the code silently using `num=3`.
```suggestion
        if num != int(num):
            raise TypeError('The num parameter should be an integer')
        num = int(num)
        if num <= 0:
            raise ValueError('The num parameter should have a value greater than 0')
```

flow/aggregates.py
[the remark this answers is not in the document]
vyasr: Whatever you call the method, make sure it is called in the constructor (and can be called again later if necessary).

flow/aggregates.py
```diff
@@ -0,0 +1,266 @@
+# Copyright (c) 2020 The Regents of the University of Michigan
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+from collections.abc import Iterable
+from hashlib import md5
+from hashlib import sha1
+from itertools import groupby
+from itertools import zip_longest
```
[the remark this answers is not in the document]
kidrahahjo: There's no specific style, but in `project.py` we import them separately.

flow/aggregates.py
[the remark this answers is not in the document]
b-butler: I agree that the first constructor argument would need to change.

flow/aggregates.py
b-butler: That is correct.

flow/aggregates.py
```diff
@@ -0,0 +1,266 @@
+# Copyright (c) 2020 The Regents of the University of Michigan
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+from collections.abc import Iterable
+from hashlib import md5
+from hashlib import sha1
```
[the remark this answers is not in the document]
kidrahahjo: `md5` is used to generate the id of an aggregate. No specific preference was there, it was just a suggestion that we could use `md5` for getting aggregate id.

flow/aggregates.py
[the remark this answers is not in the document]
b-butler: We could if this implementation had the complete desired behavior. However, once we add `__getitem__` and `__contains__` both of which would look different for the classes and the fact that for the default case the regeneration of aggregates is a no-op. I think having separate classes is warranted.

flow/aggregates.py
kidrahahjo: Will do

flow/aggregates.py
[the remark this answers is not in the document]
b-butler: Are you referring to having `Aggregate` be the same class as `_AggregateStore`? or the distinction between default aggregation and non-default aggregation.

flow/aggregates.py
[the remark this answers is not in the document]
kidrahahjo: It can be `aggregator_function`?

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+
+        .. code-block:: python
+
+            @Aggregate.groupsof(num=2)
+            @FlowProject.operation
+            def foo(*jobs):
+                print(len(jobs))
+
+        :param num:
+            Maximum possible length of an aggregate.
```
b-butler, line 101: Since only the last aggregate can be a different size. I would say something like the default size of aggregates excluding the final aggregate.

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+            args = [iter(jobs)] * num
+            return zip_longest(*args)
+
+        setattr(aggregator, '_num', num)
+
+        return cls(aggregator, sort_by, reverse_order, select, num=num)
+
+    @classmethod
+    def groupby(cls, key, default=None, sort_by=None, reverse_order=False, select=None):
+        """signac jobs can be aggregated by a valid state point parameter.
```
b-butler, line 140: See comment about `groupsof` above.

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+                print(len(jobs))
+
+        :param key:
+            Parameter which specifies how the jobs should be aggregated.
+        :type key:
+            str, Iterable, or callable
+        :param key:
+            Default values used for grouping if invalid key is passed
+        :type key:
+            str, Iterable, or callable
```
b-butler, line 162: the key parameter needs more explanation. Also, you have the key parameter listed twice.

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+        # Get unique id for _aggregator and _select attributes
+        blob_l += list(map(self._get_unique_function_id, [self._aggregator, self._select]))
+        blob = ','.join(str(attr) for attr in blob_l)
+
+        return int(sha1(blob.encode('utf-8')).hexdigest(), 16)
+
+    def _get_unique_function_id(self, func):
+        try:
+            return func.__code__.co_code
+        except Exception:  # Got partial function
```
b-butler, line 257: This is not necessarily true. A variety of things could have been passed here. I would just mention that we were passed something other than a function.

flow/aggregates.py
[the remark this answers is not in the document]
kidrahahjo: The point of `_DefaultAggregateStore` was that we wouldn't want to compute those aggregates as aggregates of 1. We could just yield single jobs.
We could merge those two classes but the current model is a simple way to differentiate between those two.

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+
+        :param aggregate:
+            Aggregate of jobs
+        :type jobs:
+            tuple of :py:class:`signac.contrib.job.Job`
+        """
+        if aggregate in self._aggregates:
+            return True
+        else:
+            return False
```
b-butler, line 329: I would prefer if we used the id to check for inclusion. It would be constant time rather than linear then.

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+        else:  # Got a project
+            project = jobs
+
+        aggregated_jobs = list(jobs)
+        if self._aggregate._select is not None:
+            aggregated_jobs = list(filter(self._aggregate._select, aggregated_jobs))
+        if self._aggregate._sort_by is not None:
+            aggregated_jobs = list(sorted(aggregated_jobs,
+                                          key=lambda job: job.sp[self._aggregate._sort_by],
+                                          reverse=bool(self._aggregate._reverse_order)))
```
b-butler, line 358: There is no need for either of the explicit list calls.

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+    def _register_aggregates(self, jobs):
+        """If the instance of this class is called then we will
+        generate aggregates and store them in self._aggregates
+        """
+        if isinstance(jobs, list):  # Got a tuple of jobs
+            project = jobs[0]._project
+        else:  # Got a project
+            project = jobs
+
+        aggregated_jobs = list(jobs)
```
b-butler, line 352: We don't need this list conversion.

flow/aggregates.py
b-butler, line 358: Do worry about the explicit `bool` conversion.

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+        if self._aggregate._select is not None:
+            aggregated_jobs = list(filter(self._aggregate._select, aggregated_jobs))
+        if self._aggregate._sort_by is not None:
+            aggregated_jobs = list(sorted(aggregated_jobs,
+                                          key=lambda job: job.sp[self._aggregate._sort_by],
+                                          reverse=bool(self._aggregate._reverse_order)))
+        aggregated_jobs = self._aggregate._aggregator([job for job in aggregated_jobs])
```
b-butler, line 359: Here you are converting to a list again. We don't need to convert to list multiple times. `sorted` always returns a list, and `filter` does not, but if we pass it into `sorted` it will return a list.

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+
+        for aggregate in aggregated_jobs:
+            try:
+                filter_aggregate = []
+                for job in aggregate:
+                    if job is None:
+                        continue
+                    elif job not in project:
+                        raise LookupError(f'The signac job {job.get_id()} not found in {project}')
+                    filter_aggregate.append(job)
```
b-butler, line 376: I think it might look cleaner to have a small function `validate_and_filter_jobs` that returns a Boolean or raises an error and just use `nested_aggregate = tuple(filter(validate_and_filter_jobs, aggregate))`

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+
+    def __len__(self):
+        return len(self._project)
+
+    def __eq__(self, other):
+        return type(self) == type(other) and \
+               self._project == other._project
+
+    def __hash__(self):
+        blob = str(self._project)
```
b-butler, line 434: Use project id.

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+    def _generate_aggregates(self, jobs):
+        """We have to store self._project when this method is invoked
+        This is because we will then iterate over that project in
+        order to return an aggregates of one.
+        """
+        self._register_project(jobs)
+
+    def _register_project(self, jobs):
+        if isinstance(jobs, list):  # Got a list of jobs
+            self._project = jobs[0]._project
```
b-butler, line 446: This is internal, should we ever be passed a list of jobs rather than a `FlowProject`? Considering we control this.

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+        else:  # Got a project
+            self._project = jobs
```
b-butler, line 448: No need for two functions here.

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+            raise LookupError(f"Did not find job with id {id}.")
+
+    def __contains__(self, job):
+        """Return whether the job is present in the project associated with this
+        instance of :py:class:`_DefaultAggregateStore`
+        """
+        if job in self._project:
+            return True
+        else:
+            return False
```
b-butler, line 424:
```suggestion
return job in self._project
```

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+    an operation function aggregated by the default aggregates i.e.
+    :py:class:`Aggregate.groupsof(1)`.
+
+    When iterated through it's instance, it yields a tuple of a single job from
+    the Project.
+
+    :param jobs:
+        The signac job handles
+    :type jobs:
+        Iterator
```
b-butler, line 400: Why would we ever be passed anything other than the project?

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+        blob_l += list(map(self._get_unique_function_id, [self._aggregator, self._select]))
+        blob = ','.join(str(attr) for attr in blob_l)
+
+        return int(sha1(blob.encode('utf-8')).hexdigest(), 16)
+
+    def _get_unique_function_id(self, func):
+        try:
+            return func.__code__.co_code
+        except Exception:  # Got partial function
+            return str(hash(func))
```
b-butler, line 258: Don't convert to string here. No need.

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+
+    def __hash__(self):
+        blob_l = list(map(hash, [self._sort_by, self._reverse_order]))
+        # Get unique id for _aggregator and _select attributes
+        blob_l += list(map(self._get_unique_function_id, [self._aggregator, self._select]))
+        blob = ','.join(str(attr) for attr in blob_l)
+
+        return int(sha1(blob.encode('utf-8')).hexdigest(), 16)
+
+    def _get_unique_function_id(self, func):
```
b-butler, line 254: I would explain what this function is doing as it is not immediately apparent. I understand, but someone fresh coming to the code might not.

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+        if(
+            self_select == other_select and
+            self_aggregator == other_aggregator
+        ):
+            return True
+
+        return False
+
+    def __hash__(self):
+        blob_l = list(map(hash, [self._sort_by, self._reverse_order]))
```
b-butler, line 247: Using map for these two attributes seems a bit much. Just writing it explicitly would be fine.

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+        self_aggregator = self._get_unique_function_id(self._aggregator)
+        other_aggregator = self._get_unique_function_id(other._aggregator)
+
+        if(
+            self_select == other_select and
+            self_aggregator == other_aggregator
+        ):
+            return True
+
+        return False
```
b-butler, line 244:
```suggestion
            return self_select == other_select and self_aggregator == other_aggregator
```

flow/aggregates.py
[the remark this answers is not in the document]
kidrahahjo: We use md5 for id calculation of Jobs.
I think that was the reason we were using md5 for calculating aggregate id.
@b-butler @vyasr Please suggest what should be done here.

flow/aggregates.py
kidrahahjo: I'm not sure what you exactly mean by this.
I will shift this conversion to `__init__` method of `Aggregate` but what else should I be worried about?

flow/aggregates.py
kidrahahjo: Oh, I'm sorry about that. Thanks for letting me know.

flow/aggregates.py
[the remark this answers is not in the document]
kidrahahjo: I changed the functionality. Please have a look at it once you get time

flow/aggregates.py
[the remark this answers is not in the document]
kidrahahjo: @b-butler The two classes, though does the same thing, are technically not at all similar.
Even if we create `_AggregateStoreBase` we'd then have to override nearly all the methods. (Because even `_register_aggregates` for both the classes registers aggregates differently.

flow/aggregates.py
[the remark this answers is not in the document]
b-butler: Is there a reason to use sha1 anywhere?

flow/aggregates.py
b-butler: This is indeed better, but the main purpose it to provide a way to compare arbitrary types that are callable, and that isn't mentioned here.

flow/aggregates.py
```diff
@@ -0,0 +1,481 @@
...
+            return zip_longest(*args)
+
+        setattr(aggregator_function, '_num', num)
+
+        return cls(aggregator_function, sort_by, reverse_order, select, num=num)
+
+    @classmethod
+    def groupby(cls, key, default=None, sort_by=None, reverse_order=False, select=None):
+        """Decorator for operation function that aggregates signac jobs by grouping them
+        via a key.
```
b-butler, line 143:
```suggestion
        """Aggregates jobs according to matching state point key values.
```

flow/aggregates.py
```diff
@@ -0,0 +1,481 @@
...
+        self._kwargs = kwargs
+        self._aggregator_function = aggregator_function
+        self._sort_by = sort_by
+        self._reverse_order = bool(reverse_order)
+        self._select = select
+
+    @classmethod
+    def groupsof(cls, num=1, sort_by=None, reverse_order=False, select=None):
+        """Decorator for operation function that aggregates signac jobs in groups of
+        a number provided by the user.
```
b-butler, line 82:
```suggestion
        """Aggregates jobs of a set group size.
```

flow/aggregates.py
```diff
@@ -0,0 +1,481 @@
...
+        return nested_aggregates, aggregate_ids
+
+    def _validate_and_filter_job(self, job):
+        "Validate whether a job is eligible to be a part of an aggregate or not."
+        if job is None:
+            return False
+        elif job in self._project:
+            return True
+        else:
+            raise LookupError(f'The signac job {job.get_id()} not found in {self._project}')
```
b-butler, line 466: If we have this class which I generally am in favor of. It should not compose an `Aggregate` object but rather take the values from the object and use them itself.

b-butler, line 466: This can be a static method

flow/aggregates.py
```diff
@@ -0,0 +1,481 @@
...
+        nested_aggregates = []
+        aggregate_ids = dict()
+
+        for aggregate in aggregated_jobs:
+            try:
+                filter_aggregate = tuple(filter(self._validate_and_filter_job, aggregate))
+                # Store aggregate in this instance
+                nested_aggregates.append(filter_aggregate)
+                # Store aggregate by their ids in order to search through id
+                aggregate_ids[get_aggregate_id(filter_aggregate)] = filter_aggregate
```
b-butler, line 453: These don't need to be in the try block.

flow/aggregates.py
```diff
@@ -0,0 +1,481 @@
...
+    def __call__(self):
+        jobs = list(self._project)
+
+        if self._aggregate._select is not None:
+            jobs = filter(self._aggregate._select, jobs)
+        if self._aggregate._sort_by is not None:
+            jobs = sorted(jobs,
+                          key=lambda job: job.sp[self._aggregate._sort_by],
+                          reverse=self._aggregate._reverse_order)
+        aggregated_jobs = self._aggregate._aggregator_function(jobs)
```
b-butler, line 437: I may have been confusing here. If we just use `filter` without `sorted` then `jobs` will not be a list when passed to `self._aggregate._aggregator_function`. It will be a `filter` object, so if `sort_by` is `None` then we do need to explicitly cast to a list, if that is an API guarantee to the user for the aggregation function rather than promising to return an iterable of jobs.

flow/aggregates.py
```diff
@@ -0,0 +1,481 @@
...
+                            f"or a callable, got {type(key)}")
+
+        def aggregator_function(jobs):
+            for key, group in groupby(sorted(jobs, key=keyfunction), key=keyfunction):
+                yield group
+
+        return cls(aggregator_function, sort_by, reverse_order, select,
+                   key=key, default=default)
+
+    def __eq__(self, other):
```
b-butler, line 219: If we make the change to `_MakeAggregates` I suggest, then we should move the `__eq__` and `__hash__` methods to it.

flow/aggregates.py
```diff
@@ -0,0 +1,481 @@
...
+        return int(sha1(blob.encode('utf-8')).hexdigest(), 16)
+
+    def _register_aggregates(self, project):
+        """If the instance of this class is called then we will
+        generate aggregates and store them in ``self._aggregates``.
+        """
+        make_aggregates = _MakeAggregates(self._aggregate, project)
+        aggregates, aggregate_ids = make_aggregates()
+        self._aggregates = aggregates
+        self._aggregate_ids = aggregate_ids
```
b-butler, line 356: If we are going to use `_MakeAggregates` then we would just want this class to have a `_MakeAggregates` object it can call when it needs to.

tests/test_aggregates.py
```diff
@@ -0,0 +1,253 @@
...
+            aggregator(helper_non_default_aggregator_function),
+            aggregator.groupsof(1), aggregator.groupsof(1),
+            aggregator.groupsof(2), aggregator.groupsof(3), aggregator.groupsof(4),
+            aggregator.groupby('half'), aggregator.groupby('half'), aggregator.groupby('even'),
+            aggregator.groupby('half', -1), aggregator.groupby('even', -1),
+            aggregator.groupby(['half', 'even'], default=[-1, -1])
+        ]
+
+        assert len(set(list_of_aggregates)) == 11
+
```
b-butler, line 127: Maybe a comment on how this tests hashing works properly would be helpful (i.e. mentioning that you have 11 unique aggregation definitions).

tests/test_aggregates.py
```diff
@@ -0,0 +1,253 @@
...
+        aggregate_job_manual = helper_aggregator_function(project)
+
+        assert [aggregate for aggregate in aggregate_job_manual] == \
+               list(aggregate_instance)
+
+    def test_valid_lambda_aggregator_function(self, setUp, project):
+        aggregate_instance = aggregator(lambda jobs: [jobs])
+        aggregate_instance = aggregate_instance._create_AggregatesStore(project)
+
+        assert [tuple(project)] == list(aggregate_instance)
```
b-butler, line 150: If this is to just test with lambda's versus standard functions, we don't need both tests. If this is to test two different functions then it can be in the same test.

tests/test_aggregates.py
```diff
@@ -0,0 +1,253 @@
...
+        assert [tuple(helper_sort(project))] == list(aggregate_instance)
+
+    def test_groups_of_valid_num(self, setUp, project):
+        valid_values = [1, 2, 3, 6, 10]
+        expected_aggregates = [10, 5, 4, 2, 1]
+
+        for i, valid_value in enumerate(valid_values):
+            aggregate_instance = aggregator.groupsof(valid_value)
+            aggregate_instance = aggregate_instance._create_AggregatesStore(project)
+            assert len(aggregate_instance) == expected_aggregates[i]
```
b-butler, line 172: This does not test that each aggregate of `aggegrator.groupsof(2)` for instance has two jobs (except the last perhaps).

tests/test_aggregates.py
```diff
@@ -0,0 +1,253 @@
...
+
+    def test_groupby_with_valid_string_key(self, setUp, project):
+        aggregate_instance = aggregator.groupby('even')
+        aggregate_instance = aggregate_instance._create_AggregatesStore(project)
+        for aggregate in aggregate_instance:
+            even = aggregate[0].sp.even
+            for job in aggregate:
+                assert even == job.sp.even
```
b-butler, line 180:
```suggestion
            assert all(even == job.sp.even for job in aggregate)
```

tests/test_aggregates.py
```diff
@@ -0,0 +1,253 @@
...
+            # doing so due to the invalid key
+            aggregate_instance._create_AggregatesStore(project)
+
+    def test_groupby_with_default_key_for_string(self, setUp, project):
+        aggregate_instance = aggregator.groupby('half', default=-1)
+        aggregate_instance = aggregate_instance._create_AggregatesStore(project)
+        for aggregate in aggregate_instance:
+            half = aggregate[0].sp.get('half', -1)
+            for job in aggregate:
+                assert half == job.sp.get('half', -1)
```
b-butler, line 196:
```suggestion
            assert all(assert half == job.sp.get('half', -1) for job in aggregate)
                
```

tests/test_aggregates.py
```diff
@@ -0,0 +1,253 @@
...
+        assert len(aggregate_instance) == 6
+
+    def test_groupby_with_Iterable_key(self, setUp, project):
+        aggregate_instance = aggregator.groupby(['i', 'even'])
+        aggregate_instance = aggregate_instance._create_AggregatesStore(project)
+        # No aggregation takes place
```
b-butler, line 202: Thanks for the comment! Could you add how that means we don't need to check whether all the aggregate members are equivalent. Of course this is true for aggregates of size one, but just to be clear for future developers.

tests/test_aggregates.py
```diff
@@ -0,0 +1,253 @@
...
+            aggregate_instance._create_AggregatesStore(project)
+
+    def test_groupby_with_valid_default_key_for_Iterable(self, setUp, project):
+        aggregate_instance = aggregator.groupby(['half', 'even'], default=[-1, -1])
+        aggregate_instance = aggregate_instance._create_AggregatesStore(project)
+        for aggregate in aggregate_instance:
+            half = aggregate[0].sp.get('half', -1)
+            even = aggregate[0].sp.get('even', -1)
+            for job in aggregate:
+                assert half == job.sp.get('half', -1) and even == job.sp.get('even', -1)
```
b-butler, line 219: Use all here too.

tests/test_aggregates.py
```diff
@@ -0,0 +1,253 @@
...
+    def test_groupby_with_callable_key(self, setUp, project):
+        def keyfunction(job):
+            return job.sp['even']
+
+        aggregate_instance = aggregator.groupby(keyfunction)
+        aggregate_instance = aggregate_instance._create_AggregatesStore(project)
+        for aggregate in aggregate_instance:
+            even = aggregate[0].sp.even
+            for job in aggregate:
+                assert even == job.sp.even
```
b-butler, line 231: Use all here.

flow/aggregates.py
b-butler: I am saying do not worry about it at all just assume it is something Boolean like.

flow/aggregates.py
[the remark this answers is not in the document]
kidrahahjo: Nope, no specific reason.
It was used while hashing `_JobOperation`, bundle id in `project.py`

flow/aggregates.py
kidrahahjo: the key can also be an arbitrary callable hence, according to me, users may get confused with this doc

flow/aggregates.py
kidrahahjo: I'm not sure I see the use case of moving them to `_MakeAggregates` can you please explain?

flow/aggregates.py
kidrahahjo: So, I should probably execute `jobs = list(jobs)` after we filter jobs

flow/aggregates.py
kidrahahjo: We'll be eventually storing the instances of `_AggregateStore` and not `_MakeAggregates`. Hence, I think it is not useful to move `__eq__` and `__hash__` method to `_MakeAggregate`

flow/aggregates.py
kidrahahjo: The `_validate_and_filter_job` uses a instance attribute `_project`.
If we convert this to a static method then we'll have to execute something like this
```
filter_aggregate = tuple(filter(
    lambda job: _validate_and_filter_job(job, self._project), aggregate))
```
I'm not sure if this is a right approach or not. Please suggest

flow/aggregates.py
b-butler: Nevermind, we do need this at first to be able to discern if an user creates multiple of the same `aggregator` objects... We will likely need a way to hash and equate `_AggregateStores` as well though.

flow/aggregates.py
b-butler: I don't think we need to internally store the project in this class. We can just expect it to be passed when called. Then we do not need to use `self._project` in fact such an attribute would not even exist. Then `_validate_and_filter-job` could be a closure created within the `_create_nested_aggregates` method. Sorry I missed that we were storing the project before.

flow/aggregates.py
```diff
@@ -0,0 +1,498 @@
...
+        A signac project used to fetch jobs for creating aggregates.
+    :type project:
+        :py:class:`flow.FlowProject` or :py:class:`signac.contrib.project.Project`
+    """
+    def __init__(self, aggregator_function, sort_by, reverse_order, select, project):
+        self._aggregator_function = aggregator_function
+        self._sort_by = sort_by
+        self._reverse_order = bool(reverse_order)
+        self._select = select
+        self._project = project
```
b-butler, line 443: I don't know why we are storing the project in this class. I know I missed this last review. I believe passing in the project to `_MakeAggregates` is the proper approach. That provides the intuitive behavior of `_MakeAggregates` as a function that takes in a project and returns aggregates.

flow/aggregates.py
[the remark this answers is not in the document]
b-butler: Okay, then I would be consistent here for this file as it is a bit cleaner.

flow/aggregates.py
```diff
@@ -0,0 +1,498 @@
...
+
+    def __call__(self):
```
b-butler, line 445: This should accept a `FlowProject`.

flow/aggregates.py
```diff
@@ -0,0 +1,498 @@
...
+        jobs = list(self._project)
+
+        if self._select is not None:
+            jobs = list(filter(self._select, jobs))
+        if self._sort_by is not None:
+            jobs = sorted(jobs,
+                          key=lambda job: job.sp[self._sort_by],
+                          reverse=self._reverse_order)
+        aggregated_jobs = self._aggregator_function(jobs)
+        return self._create_nested_aggregate_list(aggregated_jobs)
```
b-butler, line 455: This should still be done by the `_AggregateStore` class. I believe the `_MakeAggregate` class should only be responsible for correctly handing the aggregation, not the data structures used by `_AggregateStore`.

flow/aggregates.py
```diff
@@ -0,0 +1,498 @@
...
+
+    def _create_nested_aggregate_list(self, aggregated_jobs):
```
b-butler, line 457: This should be moved back to `_AggregateStore`.

flow/aggregates.py
b-butler, line 455: If we make the change I suggest. I would like to see this yield aggregates. That is just a preference though.

flow/aggregates.py
kidrahahjo: @b-butler 
In this implementation, we do have an `__eq__` and `__hash__` methods for both `_DefaultAggregateStore` and `_AggregatesStore`

flow/aggregates.py
```diff
@@ -303,12 +302,12 @@ class _AggregatesStore:
     """
     def __init__(self, aggregator, project):
         self._aggregator = aggregator
-
+        self._project = project
```
b-butler, line 305: We do not need to store the project here.

flow/aggregates.py
```diff
@@ -303,12 +302,12 @@ class _AggregatesStore:
...
         # We need to create a _MakeAggregate instance associated with this object
         # in order to generate aggregates whenever necessary.
         self._make_aggregates = _MakeAggregates(
             aggregator._aggregator_function, aggregator._sort_by,
-            aggregator._reverse_order, aggregator._select, project)
+            aggregator._reverse_order, aggregator._select)
```
b-butler, line 310: We could use a classmethod for `_MakeAggregates` that takes the `aggregator` and handles this setting internally.

flow/aggregates.py
```diff
@@ -354,7 +353,39 @@ def _register_aggregates(self):
         """If the instance of this class is called then we will
         generate aggregates and store them in ``self._aggregates``.
         """
-        self._aggregates, self._aggregate_ids = self._make_aggregates()
+        # Since we should always clear the aggregates before registering
+        # aggregates because we append the aggregates directly.
+        self._aggregates.clear()
+        self._aggregate_ids.clear()
+        aggregated_jobs = self._make_aggregates(self._project)
```
b-butler, line 360: We should pass in the project to this function and the `_create_nested_aggregate_list` function, or create the `_validate_and_filter_job` function in this function and pass that to `_create_nested_aggregate_list`. Either is fine with me. It is up to your preference.

flow/aggregates.py
```diff
@@ -430,57 +461,22 @@ class _MakeAggregates:
...
         self._sort_by = sort_by
         self._reverse_order = bool(reverse_order)
         self._select = select
-        self._project = project
-
-    def __call__(self):
-        jobs = list(self._project)
 
+    def __call__(self, project):
+        jobs = list(project)
```
b-butler, line 472: This list call is unnecessary. If we require that the aggregator function be passed a list, then we should only call list if `self._sort_by is None`.

````python
if self._sort_by is None:
    jobs = list(jobs)
else:
    jobs = sorted(...)

File: flow/aggregates.py [hunk 65]
```diff
@@ -469,15 +467,28 @@ def __init__(self, aggregator_function, sort_by, reverse_order, select):
...
-        if self._sort_by is not None:
-            jobs = sorted(jobs,
+        if self._sort_by is None:
+            jobs = list(project)
+        else:
+            jobs = sorted(project,
                           key=lambda job: job.sp[self._sort_by],
                           reverse=self._reverse_order)
+        if self._select is not None:
+            jobs = list(filter(self._select, jobs))
```
b-butler, line 477: Why is this after `sort_by` now? With this being before the `sort_by` conditional, we only convert to a list once. Now we would do it twice if `sort_by is not None`. If we put it before, we also don't need the `list` conversion there.

flow/aggregates.py
b-butler: I am fine because this makes errors easier for the user to manage.

flow/aggregates.py
```diff
@@ -0,0 +1,505 @@
...
+    :type reverse_order:
+        bool
+    :param select:
+        Condition for filtering individual jobs. This is passed as the
+        callable argument to `filter`.
+        The default behavior is no filtering.
+    :type select:
+        callable or NoneType
+    :param \*\*kwargs:
+        Additional information related to aggregator_function function.
```
b-butler, line 46: Why are we accepting `kwargs` when they are not used anywhere meaningfully?

flow/aggregates.py
```diff
@@ -0,0 +1,505 @@
...
+        """
+        return get_aggregate_id(aggregate) in self._aggregate_ids
+
+    def __len__(self):
+        return len(self._aggregates)
+
+    def __eq__(self, other):
+        return type(self) == type(other) and \
+               self._aggregates == other._aggregates and \
+               self._aggregator == other._aggregator
```
b-butler, line 344: We just need to check that the aggregate ids all all the same.

flow/aggregates.py
```diff
@@ -0,0 +1,505 @@
...
+        aggregator object associated with this class.
+    :type aggregator:
+        :py:class:`aggregator`
+    :param project:
+        A signac project used to fetch jobs for creating aggregates.
+    :type project:
+        :py:class:`flow.FlowProject` or :py:class:`signac.contrib.project.Project`
+    """
+    def __init__(self, aggregator, project):
+        self._aggregator = aggregator
```
b-butler, line 304: We don't need to store the aggregator.

flow/aggregates.py
```diff
@@ -0,0 +1,505 @@
...
+    def __len__(self):
+        return len(self._aggregates)
+
+    def __eq__(self, other):
+        return type(self) == type(other) and \
+               self._aggregates == other._aggregates and \
+               self._aggregator == other._aggregator
+
+    def __hash__(self):
+        blob = str(hash(self._aggregate))
```
b-butler, line 347:
```suggestion
        blob = str(hash(self._aggregates))
```

tests/test_aggregates.py
```diff
@@ -0,0 +1,270 @@
...
+
+        # The above list contains 11 distinct aggregator objects and some duplicates.
+        # When this list is converted to set, then these objects hashed first and
+        # then compared. Since sets don't carry duplicate values, we test whether the
+        # length of the set obtained from the list is equal to 11 or not.
+        assert len(set(list_of_aggregates)) == 11
+
+
+# Test the _AggregatesStore and _DefaultAggregateStore classes
+class TestAggregateStoring(AggregateProjectSetup):
```
b-butler, line 135: Can we test the `__hash__` and `__eq__` of this class given the typo I found.

flow/aggregates.py
kidrahahjo: These are used to check equality between two aggregator objects. I was having difficulty in comparing aggregators hence I thought that these additional parameters would be helpful to test the equality

tests/test_aggregates.py
kidrahahjo: Yes, I think I forgot to write those. Sorry about that. Will do it.

flow/aggregates.py
kidrahahjo: We compare the `aggregator` hence I think that we should store the aggregator

flow/aggregates.py
kidrahahjo: Do you mean something like this
`self._aggregate_ids == other._aggregate_ids` ?

flow/aggregates.py
kidrahahjo: This should be `self._aggregator`

flow/aggregates.py
b-butler: If this is just for testing equality, please remove it. This feature is not used for any purpose. We should deal directly with the problem of testing equality. What part was giving you difficulty?

flow/aggregates.py
b-butler: Yep.

flow/aggregates.py
b-butler: We just need to compare aggregation_ids or hash to determine uniqueness for standard aggregate store object the default is easier with just comparing and using the project directly.

flow/aggregates.py
```diff
@@ -0,0 +1,509 @@
...
+        except KeyError:
+            raise LookupError(f"Did not find job with id {id}.")
+
+    def __contains__(self, aggregate):
+        """Return whether the job is present in the project associated with this
+        instance of :py:class:`_DefaultAggregateStore`.
+        """
+        # signac-flow internally assumes every aggregate to be a tuple.
+        # Hence this method will also get a tuple as an input.
+        assert len(aggregate) == 1
```
b-butler, line 422: We should not assert that the length of something passed to `__contains__` here is one. It may be convenient to check for inclusion of an aggregate against all `_AggregateStore` objects default or not. If the length is not 1 then we just return false.

flow/aggregates.py
```diff
@@ -0,0 +1,509 @@
...
+        """
+        return get_aggregate_id(aggregate) in self._aggregate_ids
+
+    def __len__(self):
+        return len(self._aggregates)
+
+    def __eq__(self, other):
+        return type(self) == type(other) and \
+               self._aggregate_ids.keys() == other._aggregate_ids.keys() and \
+               self._aggregator == other._aggregator
```
b-butler, line 344: The last part is unnecessary. If all the aggregate ids are the same then the store is the same for all intents and purposes, unless you think that multiple forms of aggregation may lead to the same exact aggregates.

flow/aggregates.py
```diff
@@ -0,0 +1,509 @@
...
+
+    def __hash__(self):
+        blob = str(hash(self._aggregator))
+        return int(sha1(blob.encode('utf-8')).hexdigest(), 16)
```
b-butler, line 348: I would prefer to just use the aggregate ids here too. Even with a million aggregates `int(sha1(','.join(a.keys())))` is fairly fast.

I think it is cleaner this way, and if we find that users like to make many huge aggregates, we can rethink that decision.

tests/test_aggregates.py
```diff
@@ -268,3 +270,38 @@ def _select(job):
...
+        list_of_storing = set(map(_create_storing, list_of_aggregates))
+        for stored_aggregate in list_of_storing:
+            for aggregate in stored_aggregate:
+                assert aggregate == stored_aggregate[get_aggregate_id(aggregate)]
+
+    def test_get_invalid_id(self, setUp, project):
+        aggregator_instance = aggregator()._create_AggregatesStore(project)
+        for job in project:
+            with pytest.raises(LookupError):
+                aggregator_instance[get_aggregate_id((job,))]
```
b-butler, line 301: Only testing one of these would be fine.

tests/test_aggregates.py
```diff
@@ -268,3 +270,38 @@ def _select(job):
...
+
+    def test_contains(self, setUp, project):
+        aggregator_instance = aggregator()._create_AggregatesStore(project)
+        assert tuple(project) in aggregator_instance
+        for job in project:
+            assert not (job,) in aggregator_instance
```
b-butler, line 307: Only testing one or two of these would be fine. We should also test that the default aggregate store handles getitem and contains correctly as well.

flow/aggregates.py
kidrahahjo: I believe we should be extra sure, there may be a case when we only have a single job in the project and two operations which groups them in groups of 2, groups of 3. Then we won't be differentiating between them. This is completely theoretical and I'll do as you suggest

flow/aggregates.py
kidrahahjo: As per discussion in slack, we're now only going to compare `_AggregatesStore` object by `aggregator` attributes.

flow/aggregates.py
```diff
@@ -69,7 +67,6 @@ def aggregator_function(jobs):
         else:
             self._is_aggregate = True
```
b-butler: We can just set this in the `groupsof` classmethod without needing to set an attribute to the function.

kidrahahjo: I'm not sure if I follow this. The `groupsof` method returns an instance of `aggregator`.
But since we're going to introduce the tag attribute, hence I think I'll simplify this logic.

flow/aggregates.py
```diff
@@ -58,19 +63,20 @@ def aggregator_function(jobs):
...
-
+        self._is_aggregate = getattr(aggregator_function, '_groupby_is_aggregate', True)
         self._aggregator_function = aggregator_function
         self._sort_by = sort_by
         self._reverse_order = bool(reverse_order)
         self._select = select
+        # The idea of providing users with an additional tag attribute comes from the fact
+        # that the byte code of an aggregator function cannot truly differentiate it with
+        # some other aggregator function. Hence while checking the equality, we also
+        # check the equality based on tags provided by an user.
```
b-butler, line 78:
```suggestion
        # check the equality based on tags provided by an user or us in a classmethod
```

flow/aggregates.py
```diff
@@ -128,9 +134,11 @@ def aggregator_function(jobs):
             args = [iter(jobs)] * num
             return zip_longest(*args)
 
-        setattr(aggregator_function, '_num', num)
+        # Special attribute for default aggregates in flow (Groups of 1)
+        if sort_by is None and select is None and not reverse_order:
+            setattr(aggregator_function, '_groupby_is_aggregate', num != 1)
 
-        return cls(aggregator_function, sort_by, reverse_order, select)
+        return cls(aggregator_function, sort_by, reverse_order, select, f'predefined-groupsof{num}')
```
b-butler: What I meant was that we can create the object and if it is the default aggregation, change the `_is_aggregate` flag after instantiation.

b-butler: Also I think `groupsof{num}` is sufficient.

flow/aggregates.py
```diff
@@ -207,26 +215,17 @@ def aggregator_function(jobs):
             for key, group in groupby(sorted(jobs, key=keyfunction), key=keyfunction):
                 yield group
 
-        setattr(aggregator_function, '_key', key)
-        setattr(aggregator_function, '_default', default)
+        tag = f'predefined-groupby-{key}-{default}'
```
b-butler, line 218: just `groupby{key}-{default}` is fine.

flow/aggregates.py
kidrahahjo: I pushed a code, please have a look. I hope that's what you meant.

b-butler:
@kidrahahjo I left one comment you can address. It isn't vital, though.

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+            self._aggregator_function == other._aggregator_function
+        ):
+            return True
+
+        # Get unique id for _select attribute
+        self_select = self._get_unique_function_id(self._select)
+        other_select = self._get_unique_function_id(other._select)
+        # Get unique id for _aggregator_function attribute
+        self_aggregator_function = self._get_unique_function_id(self._aggregator_function)
+        other_aggregator_function = self._get_unique_function_id(other._aggregator_function)
```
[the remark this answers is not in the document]
kidrahahjo: I'm not pretty sure about this.
What if `_aggregator` attributes are equal but the `_select` attributes are not.

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+                raise ValueError('The num parameter should be an integer')
+            num = int(num)
+            if num <= 0:
+                raise ValueError('The num parameter should have a value greater than 0')
+        except TypeError:
+            raise TypeError('The num parameter should be an integer')
+
+        def aggregator_function(jobs):
+            args = [iter(jobs)] * num
+            return zip_longest(*args)
```
vyasr, line 134: This solution is pretty hacky... it's abusing the fact that you can store multiple references to the same iterator and then iterate over each reference in sequence to produce chunks of the original. It certainly works, though, and maybe it is the most efficient way to do this without making copies. I assume you found this recipe somewhere, could you add a link to document that, and maybe a comment here as well describing how it works?

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+        if num == 1 and sort_by == select is None and not reverse_order:
+            aggregator_obj._is_aggregate = False
+
+        return aggregator_obj
+
+    @classmethod
+    def groupby(cls, key, default=None, sort_by=None, reverse_order=False, select=None):
+        """Aggregates jobs according to matching state point key values.
+
+        The below code-block provides an example on how to aggregate jobs having
```
vyasr, line 148:
```suggestion
        The below code block provides an example of how to aggregate jobs having a
```

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+        .. code-block:: python
+
+            @aggregator.groupby(key='sp', default=-1)
+            @FlowProject.operation
+            def foo(*jobs):
+                print(len(jobs))
+
+        :param key:
+            Parameter which specifies how to group jobs based on a single
+            or a list of state point parameters or a custom callable.
```
vyasr, line 161:
```suggestion
            The method by which jobs are grouped. It may be a state point
            or a sequence of state points to group by specific state point
            keys. It may also be an arbitrary callable of :class:`signac.Job`
            when greater flexibility is needed.
```

flow/aggregates.py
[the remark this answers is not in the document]
vyasr: I think the intention here was to use `md5` everywhere rather than use `sha1` everywhere. We use `md5` for the `id` of jobs, so let's be consistent here. Additionally, `md5` is a faster hash function and it generates shorter hashes, which is good for keeping the directory names shorter for jobs and the strings shorter for aggregate stores. In general we're not concerned with the security implications; if that changes we'll need to change hashes everywhere in signac anyway (and we'd need to use something more secure than sha1).

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+
+    def __hash__(self):
+        blob_l = [
+            hash(self._sort_by), hash(self._reverse_order), hash(self._tag),
+            self._get_unique_function_id(self._aggregator_function),
+            self._get_unique_function_id(self._select)
+        ]
+        blob = ','.join(str(attr) for attr in blob_l)
+
+        return int(sha1(blob.encode('utf-8')).hexdigest(), 16)
```
vyasr, line 254: See comment on import: use md5 instead of sha1.

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+        try:
+            return func.__code__.co_code
+        except Exception:  # Got something other than a function
+            return hash(func)
+
+    def _create_AggregatesStore(self, project):
+        """Create the instance of _AggregatesStore or _DefaultAggregateStore classes
+        to make sure that we don't explicitly store the aggregates in the FlowProject.
+
+        When iterated over these classes, an aggregate is yielded.
```
vyasr, line 270:
```suggestion
        """Create the actual collections of jobs to be sent to aggregate operations.
        
        The :class:`aggregate` class is just a decorator that provides a signal for
        operation functions that should be treated as aggregate operations and
        information on how to perform the aggregation. This function generates
        the classes that actually hold sequences of jobs to which aggregate
        operations will be applied.
```

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+            return func
+        else:
+            raise TypeError('Invalid argument passed while calling '
+                            'the aggregate instance. Expected a callable, '
+                            f'got {type(func)}.')
+
+
+class _AggregatesStore:
+    """This class holds the information of all the aggregates associated with
+    a :py:class:`aggregator`.
```
vyasr, line 294:
```suggestion
    a :class:`aggregator`.
```

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+    def __iter__(self):
+        for job in self._project:
+            yield (job,)
+
+    def __getitem__(self, id):
+        "Return a tuple of a single job via job id."
+        try:
+            return (self._project.open_job(id=id),)
+        except KeyError:
+            raise LookupError(f"Did not find job with id {id}.")
```
vyasr, line 422: Should this error be about a job or an aggregate? This is a forward-looking question about how the aggregate's contains method will actually be invoked, @b-butler in future work making use of this class which error would be more appropriate? It seems like we'd want the error to state something about aggregates, but since this class will become the default mode for all aggregation maybe that's not the case since this error could get thrown for operations that act on individual jobs.

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+    :param jobs:
+        The signac job handles
+    :type jobs:
+        tuple
+    """
+    if len(jobs) == 1:
+        return jobs[0].get_id()  # Return job id as it's already unique
+
+    blob = ','.join((job.get_id() for job in jobs))
+    return f'agg-{sha1(blob.encode()).hexdigest()}'
```
vyasr, line 463: Since this pattern of string[->sequence of strings]->blob->hash is very common, can you define a simple internal function `_hash` that we can call everywhere? That will ensure consistency in our hashing and simplify any future changes (for instance, to which hash function we call).

tests/test_aggregates.py
```diff
@@ -0,0 +1,319 @@
...
+        self._tmp_dir = TemporaryDirectory(prefix='flow-aggregate_')
+        request.addfinalizer(self._tmp_dir.cleanup)
+        self.project = self.project_class.init_project(
+            name='AggregateTestProject',
+            root=self._tmp_dir.name)
+
+    def mock_project(self):
+        project = self.project_class.get_project(root=self._tmp_dir.name)
+        for i in range(10):
+            even = i % 2 == 0
```
vyasr, line 47:
```suggestion
            even = (i % 2) == 0
```

Relying on operator precedence when using anything other than pure arithmetic operators always makes me uneasy.

flow/aggregates.py
kidrahahjo: This was originally implemented by @csadorf in #52 and that had a link from where it was copied.
I'll add the link into this PR and also mention #52

flow/aggregates.py
kidrahahjo: That is true.
In #335 I have added support for these classes and I believe that it should be about an aggregate because everything will become an aggregate internally.
I'm curious to know what @b-butler thinks about this

flow/aggregates.py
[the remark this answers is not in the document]
kidrahahjo: A user could do something like this
```
@aggregator.groupsof(2)
@FlowProject.operation
def op_1(*jobs):
    pass

@aggregator.groupsof(2)
@FlowProject.operation
def op_2(*jobs):
    pass
```
We won't be able to differentiate between these two if we don't use `__code__.co_code` along with `tag` attribute

flow/aggregates.py
```diff
@@ -0,0 +1,475 @@
...
+            if num != int(num):
+                raise ValueError('The num parameter should be an integer')
+            num = int(num)
+            if num <= 0:
+                raise ValueError('The num parameter should have a value greater than 0')
+        except TypeError:
+            raise TypeError('The num parameter should be an integer')
+
+        # The idea of this method was originally implemented in #52 which was copied from:
+        # https://docs.python.org/3/library/itertools.html#itertools.zip_longest
```
vyasr, line 134: Putting #52 doesn't help very much, it requires somebody to know that it refers to a pull request (and in the event that we ever switched off of Github we could lose PRs entirely). I would include the link and mention the `grouper` recipe (that's the function at that link that implements this technique).

flow/aggregates.py
```diff
@@ -0,0 +1,475 @@
...
+            return False
+        elif (
+             self._sort_by != other._sort_by or
+             self._reverse_order != other._reverse_order or
+             self._tag != other._tag
+        ):
+            return False
+        elif (
+             self._select == other._select and
+             self._aggregator_function == other._aggregator_function
```
vyasr, line 237: I can't seem to access the original thread of discussion here, so I'll just make a new comment. The point that I was trying to make about this function (the whole `__eq__` method) being complicated was not about why we need the `tag`. I get that the tag might be necessary to distinguish between two. I was asking whether this `elif` clause is worth having at all. The remainder of this method basically tries to compare either the `co_code` or the hash of the `select`s and `aggregator_function`s to determine their equality, and either of those will be basically the same speed as the equality check in this `elif` (the `co_code` is just cached in the object on creation AFAIK). So I don't see any reason not to change this `elif` to an `else` clause, remove the equality checks that are currently here, and move the code with `self_select`, `other_select`, etc into this `else` clause.

tests/test_aggregates.py
vyasr: Have these tests been written?

flow/aggregates.py
```diff
@@ -0,0 +1,463 @@
...
+        # then we explicitly set the `_is_aggregate` attribute to False.
+        self._is_aggregate = True
+        self._aggregator_function = aggregator_function
+        self._sort_by = sort_by
+        self._reverse_order = bool(reverse_order)
+        self._select = select
+        # The idea of providing users with an additional tag attribute comes from the fact
+        # that the byte code of an aggregator function cannot truly differentiate it with
+        # some other aggregator function. Hence while checking the equality, we also
+        # check the equality based on tag provided by an user or us in a classmethod.
```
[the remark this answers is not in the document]
vyasr: Now that the information is in the docstring, you can remove this comment entirely.

flow/aggregates.py
```diff
@@ -0,0 +1,475 @@
...
+        Condition for filtering individual jobs. This is passed as the
+        callable argument to `filter`. The default behavior is no filtering.
+    :type select:
+        callable or NoneType
+    :param tag:
+        Additional information about the aggregator object which is a factor of
+        difference between two aggregator objects. signac-flow uses byte code
+        to compare callables like `aggregator_function` and `select`. Since a
+        byte code cannot truly differentiate between callables hence the tag,
+        if provided, becomes a factor of differentiation.
```
vyasr, line 48: Does setting tags affect correctness, or is it just a performance optimization? Now I'm actually a little confused what the goal of the tags are. Are there cases where a user would have to provide tags, otherwise we might think that two aggregates are equivalent and therefore reuse the same set of job aggregates for them even though they should actually be using different ones? That seems _highly_ problematic to me. If the worst case scenario is just that we regenerate too many aggregates, then providing tags as a way for users to speed that up is fine, but we should never be in a situation where we silently use the wrong aggregates.

It should be possible to avoid any false positives with equality, it's just not possible to avoid getting false negatives (which is where the optimization comes in). Can you give an example where the current test fails to distinguish functions without the use of tags? Does that issue get fixed if you also check `co_consts` in addition to `co_code`?

flow/aggregates.py
kidrahahjo: @vyasr, @b-butler asked me to check this first in order to reduce the complexity.
For example, if we're able to compare the `aggregator_function` and `select` directly then there's no need to compare the binary code. But if we fail, then we end up comparing the binary code.

tests/test_aggregates.py
kidrahahjo: @vyasr yes, I have written the tests.
You can have a look at them (They are the last 4 tests of this class)

flow/aggregates.py
vyasr: I understand, but I'm saying that I don't expect this to lead to any real performance improvement. In practice I expect users to use the provided `groupsof` and `groupby` options the vast majority of the time, and those will never pass this check because those decorators create new internal functions each time they are called. When users do bother to create custom aggregation functions, I expect them to use lambdas or something most of the time, in which case again this equality check will fail. While checking the type and then checking the sort/reverse/tag are probably useful, this branch adds complexity without any evidence that it will improve performance. That is the hallmark of premature optimization, so I'm recommending not doing this unless we see that the `__eq__` method is actually a performance bottleneck that could benefit from this extra check.

kidrahahjo: I understand what you say.
@b-butler It'd be really helpful to know your views on this. If you agree, then I'll go ahead and remove the block.

flow/aggregates.py
kidrahahjo: I did not try `co_consts`. I will try to make that work and let you know

vyasr: Well first, can you give an example where the tag is _necessary_, i.e. we would think two aggregators are identical when they really aren't? I have no idea if `co_consts` would fix the problem because I don't know what it is, I just suggested that because there is more information available than `co_code` if that is insufficient to distinguish.

For a feature like this we _cannot_ rely on the user to provide tags to prevent incorrectly identifying two distinct aggregates as the same. If we really can't distinguish them for some reason, I think we would have to eat the performance cost and just copy all aggregates and not reuse any of them. Optimizations cannot result in incorrect behavior, and we can't push that responsibility onto the user.

kidrahahjo: When we use `co_code`, we are not able to differentiate between `groupsof(2)` and `groupsof(3)`.
I believe this is because the function doesn't itself store any information about the number provided.
It just references to the value stored for num.
This is where `tag` comes handy. `co_consts` will not work here because it will just provide the constants declared in the function (which we don't do).

vyasr: Yes that's what I would expect. In this particular case you can get around the problem by instead looking at `x.__closure__`. However, I'm not sure if this would work in *all* cases. I think if you traversed all members of `x.__code__` and `x.__closure__` and checked them each for equality with the same members of another function you could guarantee that you don't get false positive, but I'm not 100% sure about that. For this particular use case, I think guaranteeing 0 false positives is far more important than preventing false negatives; the latter is just a performance hit, whereas the former leads to incorrect behavior. Reiterating what I said in my previous comment, it's much better to store some extra aggregates than to use the wrong aggregates for operations.

@b-butler what are your thoughts here?

flow/aggregates.py
b-butler: I agree with @vyasr here generally, hence the comment I made before.

flow/aggregates.py
b-butler: I was not aware of `__closure__`, but that would likely be sufficient then. I don't care about false negatives, but yes we need to be sure of no false positives. We should also be fairly sure that `groupsof` and `groupby` are not giving false negatives.

kidrahahjo:
In signac dev meeting, we concluded that for now, we should always differentiate between `aggregator` objects if it's not a default.
This means while equating the aggregator objects, we only compare the `_is_aggregate` attribute associated with the object.

flow/aggregates.py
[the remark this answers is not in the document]
bdice: I also prefer fully qualified names (`import itertools` and `itertools.groupby`) to avoid confusion about the similarly-named aggregate functions.

PR #351, opened 2020-09-02
Fix hashing of aggregate classes, rename reverse_order to sort_ascending

bdice:
## Description
Fixes a few minor issues in #348 that I caught after it was merged.

## Motivation and Context
I will add comments in the PR discussing the motivation for my changes.

## Types of Changes

- [x] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/aggregates.py
+++ b/flow/aggregates.py
@@ -422,8 +422,5 @@ def get_aggregate_id(jobs):
         return jobs[0].get_id()  # Return job id as it's already unique
 
     blob = ','.join((job.get_id() for job in jobs))
-    return f'agg-{_hash(blob)}'
-
-
-def _hash(blob):
-    return md5(blob.encode('utf-8')).hexdigest()
+    hash_ = md5(blob.encode('utf-8')).hexdigest()
+    return f'agg-{hash_}'
--- a/flow/aggregates.py
+++ b/flow/aggregates.py
@@ -261,7 +261,7 @@ class _AggregatesStore:
     a :class:`aggregator`.
 
     This is a callable class which, when called, generates all the aggregates.
-    When iterated through it's instance, all the aggregates are yielded.
+    Iterating over this class yields all aggregates.
 
     :param aggregator:
         aggregator object associated with this class.
@@ -361,11 +361,10 @@ class _AggregatesStore:
 
 class _DefaultAggregateStore:
     """This class holds the information of the project associated with
-    an operation function aggregated by the default aggregates i.e.
-    :py:class:`aggregator.groupsof(1)`.
+    an operation function using the default aggregator, i.e.
+    ``aggregator.groupsof(1)``.
 
-    When iterated through it's instance, it yields a tuple of a single job from
-    the Project.
+    Iterating over this class yields tuples each containing one job from the project.
 
     :param project:
         A signac project used to fetch jobs for creating aggregates.
@@ -412,7 +411,7 @@ class _DefaultAggregateStore:
 
 
 def get_aggregate_id(jobs):
-    """Generate hashed id for an aggregate of jobs
+    """Generate hashed id for an aggregate of jobs.
 
     :param jobs:
         The signac job handles
--- a/flow/aggregates.py
+++ b/flow/aggregates.py
@@ -219,6 +219,10 @@ class aggregator:
         """Generate unique id for the function passed. The id returned is used to generate
         hash and compare arbitrary types which are callable like ``self._aggregator_function``
         and ``self._select`` attributes.
+
+        Since ``select`` and ``aggregator_function`` are internal hence hash for the similar
+        functions can also differ. Hence we need to generate byte code in order to be able to
+        equate the functions properly.
         """
         try:
             return hash(func.__code__.co_code)
--- a/flow/aggregates.py
+++ b/flow/aggregates.py
@@ -30,10 +30,10 @@ class aggregator:
         The default behavior is no sorting.
     :type sort_by:
         str or NoneType
-    :param reverse_order:
-        States if the jobs are to be sorted in reverse order.
-        The default value is False.
-    :type reverse_order:
+    :param sort_ascending:
+        States if the jobs are to be sorted in ascending order.
+        The default value is True.
+    :type sort_ascending:
         bool
     :param select:
         Condition for filtering individual jobs. This is passed as the
@@ -42,7 +42,7 @@ class aggregator:
         callable or NoneType
     """
 
-    def __init__(self, aggregator_function=None, sort_by=None, reverse_order=False, select=None):
+    def __init__(self, aggregator_function=None, sort_by=None, sort_ascending=True, select=None):
         if aggregator_function is None:
             def aggregator_function(jobs):
                 return [jobs]
@@ -61,11 +61,11 @@ class aggregator:
         self._is_aggregate = True
         self._aggregator_function = aggregator_function
         self._sort_by = sort_by
-        self._reverse_order = bool(reverse_order)
+        self._sort_ascending = bool(sort_ascending)
         self._select = select
 
     @classmethod
-    def groupsof(cls, num=1, sort_by=None, reverse_order=False, select=None):
+    def groupsof(cls, num=1, sort_by=None, sort_ascending=True, select=None):
         """Aggregates jobs of a set group size.
 
         By default aggregate of a single job is created.
@@ -95,10 +95,10 @@ class aggregator:
             The default behavior is no sorting.
         :type sort_by:
             str or NoneType
-        :param reverse_order:
-            States if the jobs are to be sorted in reverse order.
-            The default value is False.
-        :type reverse_order:
+        :param sort_ascending:
+            States if the jobs are to be sorted in ascending order.
+            The default value is True.
+        :type sort_ascending:
             bool
         :param select:
             Condition for filtering individual jobs. This is passed as the
@@ -122,15 +122,15 @@ class aggregator:
             args = [iter(jobs)] * num
             return zip_longest(*args)
 
-        aggregator_obj = cls(aggregator_function, sort_by, reverse_order, select)
+        aggregator_obj = cls(aggregator_function, sort_by, sort_ascending, select)
 
-        if num == 1 and sort_by == select is None and not reverse_order:
+        if num == 1 and sort_by == select is None and sort_ascending:
             aggregator_obj._is_aggregate = False
 
         return aggregator_obj
 
     @classmethod
-    def groupby(cls, key, default=None, sort_by=None, reverse_order=False, select=None):
+    def groupby(cls, key, default=None, sort_by=None, sort_ascending=True, select=None):
         """Aggregates jobs according to matching state point key values.
 
         The below code block provides an example of how to aggregate jobs having a
@@ -160,10 +160,10 @@ class aggregator:
             The default behavior is no sorting.
         :type sort_by:
             str or NoneType
-        :param reverse_order:
-            States if the jobs are to be sorted in reverse order.
-            The default value is False.
-        :type reverse_order:
+        :param sort_ascending:
+            States if the jobs are to be sorted in ascending order.
+            The default value is True.
+        :type sort_ascending:
             bool
         :param select:
             Condition for filtering individual jobs. This is passed as the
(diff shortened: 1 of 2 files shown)


flow/aggregates.py
```diff
@@ -30,10 +30,10 @@ def foo(*jobs):
         The default behavior is no sorting.
     :type sort_by:
         str or NoneType
-    :param reverse_order:
-        States if the jobs are to be sorted in reverse order.
-        The default value is False.
-    :type reverse_order:
+    :param sort_ascending:
```
bdice, line 33: We should call this `sort_ascending` with a default value of `True` so it matches Pandas' `sort_values(by=..., ascending=True, ...)`. I find "ascending" unambiguous, while "reverse order" is less clear. https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.sort_values.html

flow/aggregates.py
```diff
@@ -206,30 +206,24 @@ def aggregator_function(jobs):
...
     def _get_unique_function_id(self, func):
         """Generate unique id for the function passed. The id returned is used to generate
         hash and compare arbitrary types which are callable like ``self._aggregator_function``
         and ``self._select`` attributes.
         """
         try:
-            return func.__code__.co_code
-        except Exception:  # Got something other than a function
+            return hash(func.__code__.co_code)
+        except AttributeError:  # Cannot access function's compiled bytecode
```
bdice, line 226: We should hash the bytecode before returning, since this is purely used as a unique identifier. We also shouldn't catch `Exception`! We know this should raise an `AttributeError` if it can't access the attribute.

flow/aggregates.py
```diff
@@ -206,30 +206,24 @@ def aggregator_function(jobs):
...
-            hash(self._sort_by), hash(self._reverse_order), hash(self._is_aggregate),
-            self._get_unique_function_id(self._aggregator_function),
-            self._get_unique_function_id(self._select)
-        ]
-        blob = ','.join(str(attr) for attr in blob_l)
-
-        return int(_hash(blob), 16)
+        return hash((self._sort_by, self._sort_ascending, self._is_aggregate,
+                     self._get_unique_function_id(self._aggregator_function),
+                     self._get_unique_function_id(self._select)))
```
bdice, line 217: We're severely over-complicating things here. Just take the hash of a tuple of hashable objects that uniquely define this instance. There's no need to do all the string operations and sha1 digests.
> The only required property is that objects which compare equal have the same hash value; it is advised to mix together the hash values of the components of the object that also play a part in comparison of objects by packing them into a tuple and hashing the tuple.
https://docs.python.org/3/reference/datamodel.html#object.__hash__

flow/aggregates.py
```diff
@@ -282,8 +276,8 @@ def __init__(self, aggregator, project):
 
         # We need to register the aggregates for this instance using the
         # project provided.
-        self._aggregates = list()
-        self._aggregate_ids = dict()
+        self._aggregates = []
+        self._aggregate_ids = {}
```
bdice, line 280: Initialize empty lists/dicts with object literals. It's faster and more Pythonic.

flow/aggregates.py
```diff
@@ -368,11 +360,10 @@ def _validate_and_filter_job(job):
 
 class _DefaultAggregateStore:
     """This class holds the information of the project associated with
-    an operation function aggregated by the default aggregates i.e.
-    :py:class:`aggregator.groupsof(1)`.
+    an operation function using the default aggregator, i.e.
+    ``aggregator.groupsof(1)``.
```
bdice, line 364: This probably won't be parsed correctly by Sphinx. `:py:class:` will create a cross-link to the corresponding class, but we just want a code snippet.

flow/aggregates.py
```diff
@@ -405,12 +396,10 @@ def __len__(self):
...
 
     def __eq__(self, other):
-        return type(self) == type(other) and \
-               self._project == other._project
+        return type(self) == type(other) and self._project == other._project
 
     def __hash__(self):
-        blob = self._project.get_id()
-        return int(_hash(blob), 16)
+        return hash(repr(self._project))
```
bdice, line 402: **This was not a safe hash function.** The project id is just a name, and isn't actually a unique identifier (two projects in different locations can have the same name, and thus the same id). For the Project's `__eq__` method, we use `repr(self) == repr(other)`, because the `repr` includes a unique identifier: the project's path on disk. I use the same approach here.

flow/aggregates.py
```diff
@@ -432,8 +421,5 @@ def get_aggregate_id(jobs):
         return jobs[0].get_id()  # Return job id as it's already unique
 
     blob = ','.join((job.get_id() for job in jobs))
-    return f'agg-{_hash(blob)}'
-
-
-def _hash(blob):
-    return md5(blob.encode('utf-8')).hexdigest()
+    hash_ = md5(blob.encode('utf-8')).hexdigest()
+    return f'agg-{hash_}'
```
bdice, line 425: The separate function was a little bit overkill -- it's only used once.

flow/aggregates.py
kidrahahjo: We sort in ascending order by default.
Hence we'll always have to use the `not` keyword.
While this will definitely work, my preference here would be using `sort_descending`

kidrahahjo (approved):
Thank you @bdice for reviewing #348 and fixing the issues through this PR.
I added a few commits which address your comments related to using fully qualified import statements.
Apart from this, it looks good to me. Thanks once, again.

flow/aggregates.py
bdice: I believe I only had to invert its value twice. Once in the call to `sorted` and also once in the check for “default values” where I replaced `not reverse_order` with `sort_ascending`. So it was an equal number of “nots” in each direction, which seems no more complex to me.

vyasr: I don't have a strong opinion either way, but this usage is distinct enough from anything you would do with pandas that I don't think there's any reason for confusion. I'm assuming the performance difference is negligible, but if that's not the case and the default sorting is the opposite of what pandas does I think choosing performance over exact API consistency with pandas is unnecessary here. I'm fine with whatever decision is made, though.

flow/aggregates.py
vyasr: See my comment above (where this function is used).

flow/aggregates.py
vyasr: Does the `_get_unique_function_id` even need to exist? @kidrahahjo why were you hashing just the bytecode of the function (which is what `_get_unique_function_id` returns for functions) rather than just hashing the function directly? I honestly don't remember thinking through whether it was necessary, I think I just assumed there was a reason you bothered to define that rather than just hashing the function directly and moved on. Any object with a `__code__` attribute should already be hashable though, right?

flow/aggregates.py
vyasr: My original intent intent in requesting a separate function was to also include the logic for blob construction (which was being done in many places in this file), but as you've pointed out you can always just hash a tuple of things rather than build up that object which is better :)

flow/aggregates.py
kidrahahjo: @vyasr Since `aggregator_function` is defined internally hence the hash would depend on it's location.
I believe this was the main reason for not creating a hash of a method directly.

flow/aggregates.py
kidrahahjo: @bdice @vyasr Then we can stick with how @bdice implemented.

flow/aggregates.py
bdice: This comment is just for others' reference. In my understanding, `_get_unique_function_id` is useful for the same reason we use `obj.__code__.co_code` for the "tags" on `_condition` objects in flow. It allows us to know that two function objects are the same if their bytecode is equivalent (even if they're not actually located in the same object). Hashing the bytecode (when possible) is a helpful way to uniquely identify the aggregator/select function.

Demonstration of how hashes of functions are not equal (even though the functions are equivalent), but hashing the bytecode shows their equivalence:
```python
>>> def foo():
...   print('asdf')
...
>>> def bar():
...   print('asdf')
...
>>>
>>> foo is bar
False
>>> hash(foo)
8793327702568
>>> hash(bar)
8793327702577
>>> foo.__code__.co_code
b't\x00d\x01\x83\x01\x01\x00d\x00S\x00'
>>> bar.__code__.co_code
b't\x00d\x01\x83\x01\x01\x00d\x00S\x00'
>>> assert hash(foo.__code__.co_code) == hash(bar.__code__.co_code)
```

flow/aggregates.py
b-butler: I don't feel this needs to conform to pandas, but I do think the naming is clearer with `sort_ascending` or `sort_descending`.

b-butler (approved):
Thanks for the changes @bdice. They look good.

flow/aggregates.py
vyasr: @kidrahahjo @bdice yes, that is true, but we're not using hashing to check equality or depending on being able to check equality at all any more since, as per our discussions, checking `co_code` is insufficient: for instance:

```
>>> def f(x):
...     return x**2
...
>>> def g(x):
...     return x**3
...
>>> f.__code__.co_code
b'|\x00d\x01\x13\x00S\x00'
>>> g.__code__.co_code
b'|\x00d\x01\x13\x00S\x00'
```

Although I pointed out some ways to get around the particular problem above (as well other even more subtle ways in which this comparison can fail), the point is that this id isn't really unique, right? So do we gain anything by defining a function whose name is not accurate and whose primary purpose is no longer valid? If `Aggregate.__eq__` only returns `True` if (roughly) `a is b`, then wouldn't it make sense for our hashes to also be distinct?

bdice: @vyasr @kidrahahjo Wow, you're right. I didn't realize that would give different results. The 2 and 3 are stored in `__code__.co_consts`, and otherwise the functions' code looks identical. We should probably change this behavior to just use the behavior of `a is b` and not try to out-smart it.

vyasr: Right. Actually the situation even gets worse than that, I discussed some cases where comparing `__code__.co_consts` is also not sufficient (and possible ways to get around those edge cases as well). However, overall I think there are enough such possibilities that it's safest to avoid the problem altogether and just require strict object equality, which is also what we changed `__eq__` to do. @kidrahahjo could you make the relevant change here (on a separate branch/PR at this point)? Basically just remove the usage of this helper function and directly hash the object unless this will cause any other problems. @b-butler feel free to comment if you think I'm missing something here, but I think this is directly in line with my suggestion for `__eq__`.

b-butler: @bdice @vyasr The `__hash__` method does not have to distinguish all elements that are not equivalent it just requires that `a == b -> hash(a) == hash(b)`. We can do this for now, but I think eventually having a hash method that can distinguish between commonly used aggregates would be useful. hhttps://docs.python.org/3/reference/datamodel.html#object.__hash__

I don't think there is any safety concern given since if `a is b` then `a == b` and `hash(a) == hash(b)` in our model. `__hash__` is not necessarily meant in Python to provide a unique id. It is more for hashable collections like `dict` and `set`. We can safely give objects the same hash when they are not equal as long as our `__eq__` gives no false positives which the `is` operator ensures.

vyasr: I agree, in general hashes don't need to be equivalent and we only depend on `__eq__`. I just figured that for now we should keep the two roughly in sync.

signac at present basically assumes no hash collisions... but that's another story.

bdice: To be clear, hash collisions in signac would be caught and raise errors. We have tests for this behavior. (Just didn’t want readers of the above comment to fear unexpected behavior.)

vyasr: Yes, hash collisions where two statepoints refer to the same job ID will raise an error. Nothing will break.

AFAIK hash collisions will raise the same `DestinationExistsError` as is raised when actually creating two identical jobs though. I'm not aware of any logic that explicitly checks for the _reason_ and indicates to the user that there's a hash collision, so the error will look the same as if the user did something like

```
j1 = project.open_job(dict(a=1)).init()
j2 = project.open_job(dict(a=1, b=2)).init()
del j2.sp.b  # This will raise a DestinationExistsError because now j2 and j1 have the same statepoint.
```

In the event that a user somehow managed to trigger a collision (which is obviously highly unlikely, and not something that I think we need to deal with), the resulting error message would prevent any corruption of the data space, it would just be very confusing. That's all I was referring to, nothing needs to be done there.

PR #354, opened 2020-09-25
Fix performance regression in job submission.

bdice:
## Description
Corrects a serious performance regression in job submission, introduced in 7e2722fe25bf5fcf3009a5b77e28e187a5eb8050.
Makes sure the `operations` generator is consumed *inside* the buffering context manager.

I also added a second buffered region during submission, so that status updates do not immediately write to the project document (only when finished, an error occurs, or the process is gracefully interrupted, all of which will trigger the context manager's exit and cause a flush to disk). Without buffering, this region is very expensive in large projects (which need to read/write the project document for every job that is submitted).

## Motivation and Context
Submission was very slow in a large workspace (100k jobs).

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,6 @@
-=======
-Changes
-=======
+=========
+Changelog
+=========
 
 The **signac-flow** package follows `semantic versioning <https://semver.org/>`_.
 The numbers in brackets denote the related GitHub issue and/or pull request.
@@ -26,6 +26,7 @@ Changed
 Fixed
 +++++
 - Serial execution on Summit correctly counts total node requirements (#342).
+- Fixed performance regression in job submission in large workspaces (#354).
 
 Removed
 +++++++
--- a/flow/project.py
+++ b/flow/project.py
@@ -3027,10 +3027,15 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
         # Gather all pending operations.
         with self._potentially_buffered():
             default_directives = self._get_default_directives()
-            # The generator must be used *inside* the buffering context manager for performance reasons
-            operation_generator = self._get_submission_operations(
-                jobs, default_directives, names, ignore_conditions, ignore_conditions_on_execution)
-            # islice takes the first "num" elements from the generator, or all items if num is None
+            # The generator must be used *inside* the buffering context manager
+            # for performance reasons
+            operation_generator = self._get_submission_operations(jobs,
+                                                                  default_directives,
+                                                                  names,
+                                                                  ignore_conditions,
+                                                                  ignore_conditions_on_execution)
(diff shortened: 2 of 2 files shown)


b-butler (approved):
@bdice thanks for the fix for the regression I introduced.

bdice:
@b-butler I added a second region of buffering, which should still be safe-ish and should provide significant performance benefits. However, I'm going to re-request your review since this change is "new" code and not just a fix. I will verify the performance separately and comment again on this PR to confirm that it's as fast as expected.

changelog.txt
```diff
@@ -1,6 +1,6 @@
-=======
-Changes
-=======
+=========
+Changelog
+=========
```
csadorf, line 3: I'll allow it. ;)

flow/project.py
```diff
@@ -3027,19 +3027,28 @@ def submit(self, bundle_size=1, jobs=None, names=None, num=None, parallel=False,
         # Gather all pending operations.
         with self._potentially_buffered():
             default_directives = self._get_default_directives()
-            operations = self._get_submission_operations(jobs, default_directives, names,
-                                                         ignore_conditions,
-                                                         ignore_conditions_on_execution)
-        if num is not None:
-            operations = list(islice(operations, num))
+            # The generator must be used *inside* the buffering context manager
+            # for performance reasons
```
csadorf, line 3031:
```suggestion
            # for performance reasons.
```

csadorf (approved):
Great, thx! 😃

changelog.txt
bdice: @csadorf This makes the name consistent with signac. 😉

flow/project.py
```diff
@@ -3027,19 +3027,28 @@ def submit(self, bundle_size=1, jobs=None, names=None, num=None, parallel=False,
...
-            operations = list(islice(operations, num))
+            # The generator must be used *inside* the buffering context manager
+            # for performance reasons.
+            operation_generator = self._get_submission_operations(jobs,
+                                                                  default_directives,
+                                                                  names,
+                                                                  ignore_conditions,
+                                                                  ignore_conditions_on_execution)
+            # islice takes the first "num" elements from the generator, or all
+            # items if num is None
```
bdice, line 3038:
```suggestion
            # items if num is None.
```

bdice:
I can confirm that this improves performance (I did not record any values, but the output scrolls by faster). There is still significant room for performance improvements in the code that is called before this part. For a submission of ~25,000 operations, signac-flow took nearly 45 minutes to do ... something (no idea what) ... before submission began. I was not able to profile this on the system where this submission took place, and will need to build a large workspace and local testing SLURM server to profile and test on.

PR #357, opened 2020-10-09
Release 0.11.0

bdice:
## Description
Release of signac-flow 0.11.0.


setup.cfg
```diff
@@ -28,6 +28,6 @@ exclude = mistune,doc/conf.py
 [bumpversion:file:.zenodo.json]
 
 [tool:pytest]
-filterwarnings =
+filterwarnings =
```
b-butler, line 31: Is this space needed?

b-butler (commented):
I left a comment about trailing whitespace, but otherwise looks good to me. I took the liberate of removing a deprecated code path.

vyasr:
What's up with the Zenodo changes?

mikemhenry (approved):
LGTM, I did some quick grepping for any missed version bumps and deprecation and didn't find anything.

kidrahahjo (approved):
We could edit the line 142 in changelog.txt to maintain consistency. ("Fixes" to "Fixed")
https://github.com/glotzerlab/signac-flow/pull/357/files#diff-e8800259c4985f4a73b0936128034c53R142

setup.cfg
bdice: It's added automatically by `bumpversion`. I added this file to the pre-commit "ignore whitespace" list so pre-commit passes regardless. This has flipped between "space" and "no space" a lot of times.

bdice:
> 
> 
> LGTM, I did some quick grepping for any missed version bumps and deprecation and didn't find anything.

@mikemhenry The files with versions to bump are listed in `setup.cfg` and handled by bumpversion, so as long as we don't add new files that include a version number, we're safe. 😄

bdice:
> 
> 
> What's up with the Zenodo changes?

This is copied from signac. I just wanted the two to match, and I think signac's style is better.

setup.cfg
b-butler: Ah well it is what it is.

PR #363, opened 2020-11-10
Fix inconsistant commandline arguments name.

b-butler:
## Description

Fixes inconsistent argument names in our command line API (in `exec`).
This relates to a positional argument, so I wouldn't consider this a breaking change.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -3720,8 +3720,8 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
         self.submit(jobs=jobs, names=names, **kwargs)
 
     def _main_exec(self, args):
-        if len(args.jobid):
-            jobs = [self.open_job(id=jid) for jid in args.jobid]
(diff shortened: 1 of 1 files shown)


bdice (approved):
I’m fine with this change. This change allows us to simplify #335.

kidrahahjo:
I think that we should also allow the `--j` flag to be more consistent.

b-butler:
@kidrahahjo in this case the desire CLI is `python3 project.py exec {op_name} {job_ids}`. Adding an optional argument that does the same as a previous positional one seems redundant. `exec` also requires at least one job so this would also not work well as a solely optional argument.

klywang (approved):
This looks good!

kidrahahjo:
Aah, yes. I forgot about that point. Thanks for this!

PR #364, opened 2020-11-16
Feature/enable aggregation

vyasr:
Add infrastructure for aggregation into flow. All the necessary classes are in place and the code has been modified where appropriate to handle aggregates instead of jobs. However, no actual aggregation occurs yet (everything is an aggregate of size 1). We plan to refactor and optimize the existing code paths before implementing that feature.


vyasr:
No review necessary here. I'll merge into next as soon as the tests on enable-aggregation pass (from merging in the #335).

PR #374, opened 2020-11-29
Linting dicts.

### LINKED ISSUE #323: Refactor OrderedDict to dict
Python's `OrderedDict` class is known to be slower than the plain `dict`. Since 3.6 in CPython, `dict` objects have been ordered as well. This was made an API guarantee in 3.7 of the Python standard. However, within `flow` I do not believe we give an API guarantee of `dict` orderings anywhere. We do need the ordering as tests are currently written, but since tests aren't an API promise with the recent drop of 3.5 support, we can use the fact that in 3.6 we have the implementation order `dict` objects without needing the 3.7 standard for testing purposes.

Other opinions on this are welcome.

bdice:
## Description
Refactored dict usage, removed OrderedDict, and simplified dict comprehensions. Resolves #323.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -853,7 +853,7 @@ class FlowGroup:
         # pretend submission output for templates.
         self.operations = {} if operations is None else dict(operations)
         if operation_directives is None:
-            self.operation_directives = dict()
+            self.operation_directives = {}
         else:
             self.operation_directives = operation_directives
 
--- a/flow/project.py
+++ b/flow/project.py
@@ -22,7 +22,7 @@ import threading
 import time
 import traceback
 import warnings
-from collections import Counter, OrderedDict, defaultdict
+from collections import Counter, defaultdict
 from copy import deepcopy
 from enum import IntFlag
 from hashlib import sha1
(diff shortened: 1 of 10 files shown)


flow/project.py
```diff
@@ -851,11 +849,9 @@ def op2(job):
...
-        # An OrderedDict is not necessary here, but is used to ensure
-        # consistent ordering of pretend submission output for templates.
-        self.operations = (
-            OrderedDict() if operations is None else OrderedDict(operations)
-        )
+        # Requires Python >=3.6: dict must be ordered to ensure consistent
+        # pretend submission output for templates.
+        self.operations = {} if operations is None else dict(operations)
         if operation_directives is None:
             self.operation_directives = dict()
```
b-butler, line 856: We should probably be consistent here and use `{}`.

b-butler (changes requested):
It may be worth grepping the repo for `dict()` and replacing with the literal `{}`. I caught one place where `dict()` is still used. Otherwise LGTM.

flow/project.py
bdice: Hmm. I grepped but must have missed this one.
```suggestion
            self.operation_directives = {}
```

bdice:
I forgot to mention that some of the dict/{} changes may be encompassed in other PRs. I did a lot of work and then split it into separate PRs, so they wouldn’t conflict or have interdependency. I’d like to merge this as-is and I will take a second look to make sure no other refactorings were missed, after the other PRs I made are merged.

b-butler (approved):
LGTM

PR #376, opened 2020-11-29
Use descriptive variable names.

bdice:
## Description
This PR updates much of the codebase to use longer/clearer variable names, such as `operation` instead of `op`. I aimed for ensuring consistency between different sections of the codebase, so that similarly-named variables tend to be of the same type.

## Types of Changes

- [x] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -228,12 +228,13 @@ def _make_bundles(operations, size=None):
     This function splits an iterable of operations into equally
     sized bundles and a possibly smaller final bundle.
     """
-    n = None if size == 0 else size
+    if size == 0:
+        size = None
     operations = iter(operations)
     while True:
-        b = list(islice(operations, n))
-        if b:
-            yield b
+        bundle = list(islice(operations, size))
+        if bundle:
+            yield bundle
         else:
             break
 
@@ -577,8 +578,8 @@ class BaseFlowOperation:
         if post is None:
             post = []
 
-        self._prereqs = [_FlowCondition(cond) for cond in pre]
-        self._postconds = [_FlowCondition(cond) for cond in post]
+        self._preconditions = [_FlowCondition(cond) for cond in pre]
+        self._postconditions = [_FlowCondition(cond) for cond in post]
 
     def __str__(self):
         return "{type}(cmd='{cmd}')".format(type=type(self).__name__, cmd=self._cmd)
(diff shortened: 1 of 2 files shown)


cbkerr (approved):
The changes look fine. Wondering if the following variables in directives.py should be expanded?
- line 256: `np`
- line 276 `v` and `o`
- also `_NP` and `_NP_DEFAULT`

bdice:
@cbkerr Thanks for the review!
1. I left `np` as-is because that's a specific argument name that we pass to schedulers/mpirun during submission. It's part of our API for directives and thus changing it would be difficult.
2. Oops. I thought I had changed that one, but I was conflicted about exactly what to call it. I made it `_no_aggregation(value, other)`, which is what I'd guess the original intention was for `v, o`.
3. Leaving `_NP` and `_NP_DEFAULT` for same reasons as (1).

flow/project.py
```diff
@@ -2835,17 +2861,17 @@ class _PickleError(Exception):
         pass
 
     @staticmethod
-    def _dumps_op(op):
+    def _operation_to_tuple(operation):
```
b-butler, line 2864:
```suggestion
    def _job_operation_to_tuple(operation):
```

flow/project.py
```diff
@@ -2835,17 +2861,17 @@ class _PickleError(Exception):
...
+            operation.id,
+            operation.name,
+            [job.get_id() for job in operation._jobs],
+            operation.cmd,
+            operation.directives,
         )
 
-    def _loads_op(self, blob):
-        id, name, job_ids, cmd, directives = blob
+    def _operation_from_tuple(self, data):
```
b-butler, line 2873:
```suggestion
    def _job_operation_from_tuple(self, data):
```

flow/project.py
```diff
@@ -2862,16 +2888,24 @@ def _run_operations_in_parallel(self, pool, pickle, operations, progress, timeou
...
-            s_tasks = [
-                (pickle.loads, s_root, self._dumps_op(op), s_ops, "run_operations")
-                for op in tqdm(operations, desc="Serialize tasks", file=sys.stderr)
+            serialized_root = pickle.dumps(self.root_directory())
+            serialized_operations = pickle.dumps(self._operations)
+            serialized_tasks = [
+                (
+                    pickle.loads,
+                    serialized_root,
+                    self._operation_to_tuple(operation),
```
b-butler, line 2897:
```suggestion
                    self._job_operation_to_tuple(operation),
```

flow/project.py
```diff
@@ -4888,9 +4931,9 @@ def _serializer(loads, root, *args):
     root = loads(root)
     project = FlowProject.get_project(root)
     if args[-1] == "run_operations":
-        operation = args[0]
+        operation_data = args[0]
         project._operations = loads(args[1])
-        project._execute_operation(project._loads_op(operation))
+        project._execute_operation(project._operation_from_tuple(operation_data))
```
b-butler, line 4936:
```suggestion
        project._execute_operation(project._job_operation_from_tuple(operation_data))
```

b-butler (approved):
Thanks for the PR! I think we could use job_operation more frequently given the number of operation classes. However, this can be left until later.

bdice:
@b-butler Good idea about “job operation.” I think future refactorings can improve that further.

PR #377, opened 2020-11-29
Remove dangling else/elif clauses after raise/return statements.

bdice:
## Description
This PR removes dangling else/elif clauses after raise/return statements.
By making this change, the indentation level is reduced, the code is simpler to read, and methods are more likely to hit a return statement rather than implicitly returning None (which reduces the likelihood of bugs).

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/directives.py
+++ b/flow/directives.py
@@ -99,8 +99,7 @@ class _Directive:
                 return self._validator(value(*jobs))
 
             return validate_callable
-        else:
-            return self._validator(value)
+        return self._validator(value)
 
 
 class _Directives(MutableMapping):
@@ -131,11 +130,10 @@ class _Directives(MutableMapping):
             raise TypeError(
                 f"Expected a _Directive object. Received {type(directive)}."
             )
-        elif directive._name in self._directive_definitions:
+        if directive._name in self._directive_definitions:
             raise ValueError(f"Cannot redefine directive name {directive._name}.")
-        else:
-            self._directive_definitions[directive._name] = directive
-            self._defined_directives[directive._name] = directive._default
+        self._directive_definitions[directive._name] = directive
+        self._defined_directives[directive._name] = directive._default
 
     def _set_defined_directive(self, key, value):
         try:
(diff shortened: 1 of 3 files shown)


flow/environment.py
```diff
@@ -88,7 +88,7 @@ def __init__(cls, name, bases, dct):
             cls.registry = OrderedDict()
         else:
             cls.registry[name] = cls
-        return super().__init__(name, bases, dct)
+        super().__init__(name, bases, dct)
```
jennyfothergill, line 91: what is the difference between calling the init and returning it here?

jennyfothergill (approved):
LGTM! I have one question, but it's more "how does this work" than I think anything is wrong.

flow/environment.py
bdice: The return is just a mistake. The return value of an `__init__` method isn’t used — returning the result of the parent constructor did nothing here, but it was triggering a warning from pylint. (I was using pylint to help me analyze the flow codebase and catch potential issues that I fixed in this series of PRs.)

jennyfothergill: Ah, OK! There are so many good tools... 🤩 I am just learning about many of them, and they are so useful!

flow/directives.py
```diff
@@ -99,8 +99,7 @@ def validate_callable(*jobs):
                 return self._validator(value(*jobs))
 
             return validate_callable
-        else:
```
b-butler, line 102: I don't have a problem with this, but having talked to other developers. Whether including the `else` block here is a code-smell or whether not including it is not well agreed upon. I tend to think the else block helps, but I don't feel strongly either way.

b-butler (commented):
The PR is fine. I like the changes where `raise` statements are concerned. The other changes are oaky; however, I am not sure that they are all improvements. I don't feel too strongly either way though. Are these changes suggestions from a standard linter/fixer?

flow/directives.py
bdice: @b-butler, pylint suggested all of these changes that remove dangling else clauses and the like. I thought about it for a while and concluded that it was a justifiable (though minor) change.

bdice: I’m not strongly opinionated on it, but I do like that it expresses the same logic at a lower indentation level, in fewer lines, and with less possibility of future code changes introducing a bug (e.g. leaving an “elif” at the end instead of an else) and returning None.

b-butler: If a tool like pylint suggested it, then I am good with the changes.

PR #378, opened 2020-11-29
Use isinstance with IgnoreConditions.

bdice:
## Description
This PR uses an `isinstance` check for IgnoreConditions instead of manually checking types. Using raw type checks is not advisable for several reasons. It's incompatible with subclasses, for example, and it raises warnings from pylint.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -597,10 +597,10 @@ class BaseFlowOperation:
         :type ignore_conditions:
             :py:class:`~.IgnoreConditions`
         """
-        if type(ignore_conditions) != IgnoreConditions:
+        if not isinstance(ignore_conditions, IgnoreConditions):
             raise ValueError(
                 "The ignore_conditions argument of FlowProject.run() "
-                "must be a member of class IgnoreConditions"
+                "must be a member of class IgnoreConditions."
             )
         # len(self._prereqs) check for speed optimization
         pre = (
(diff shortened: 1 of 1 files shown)


atravitz:
Should we also do this for the instance in `utils.py`? (I'm not sure how to link to other files here)

bdice:
@atravitz Good question. I'm guessing you mean this: https://github.com/glotzerlab/signac-flow/blob/088150ba035413851dc14178cbf9202deb77838d/flow/util/misc.py#L167-L174

That may be worth changing, but I think it's out of scope for this PR. I'm unsure if that helper function is necessary at all, and the only place it's called seems like it could use some refactoring.

There is also a private method for the same thing in signac.
https://github.com/glotzerlab/signac/blob/9db354f758145e957b8813c69a7d50425cc5c013/signac/contrib/utility.py#L405-L429

This hashing behavior was introduced here:
https://github.com/glotzerlab/signac/pull/218
https://github.com/glotzerlab/signac/issues/217

For now, I'm going to make the `to_hashable` method private on `next` without any deprecation cycle. It's only used in one place and is clearly not meant to be part of the public API. I will open an issue with a suggested refactoring for the rest of the surrounding code.

mikemhenry (approved):
LGTM

b-butler (approved):
LGTM

PR #379, opened 2020-11-29
Change _verify_group_compatibility to classmethod.

bdice:
## Description
This PR makes `_verify_group_compatibility` a class method, since it doesn't use `self` in the method body.

This would be a good candidate to refactor into a helper function outside the FlowProject class if reviewers think that is a better outcome.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -3176,7 +3176,7 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
             else:
                 continue
         operations = list(operations.values())
-        if not self._verify_group_compatibility(operations):
+        if not FlowProject._verify_group_compatibility(operations):
             raise ValueError(
                 "Cannot specify groups or operations that "
                 "will be included twice when using the"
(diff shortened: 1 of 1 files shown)


mikemhenry (approved):
LGTM! I think we can assess later what we want to pull out as helper/utility functions. This change will make it easier to re-factor and pullout later if we decide to do that. When we change a private API, do we want to document that in the change log? I can't remember if we have a "developer" change log or section to document those kinds of changes, but I wouldn't want this to clutter user release notes.

klywang (approved):
Looks good!

bdice:
> When we change a private API, do we want to document that in the change log? I can't remember if we have a "developer" change log or section to document those kinds of changes, but I wouldn't want this to clutter user release notes.

@mikemhenry I usually include internal changes if they may materially impact users. Optimizations, for example, don’t change APIs but are a meaningful improvement for users. For private method refactors like this, I don’t bother putting a changelog entry. On the other side of API change (deprecation), there have been some recent cases where I removed tiny public methods from utility modules and made them private because a deprecation cycle seemed to be not worth it. I generally trust our reviewers and contributors to make their own conclusions on these things.

PR #380, opened 2020-11-29
Remove unused attributes/methods.

bdice:
## Description
This PR removes some unused arguments/methods and performs a couple other small edits that didn't fit into other PRs.

- Remove unused `__init__` from pre/post conditions classes, because the implicit parent constructor is equivalent.
- Remove unused argument from `_add_operation_selection_arg_group`. At one point, `python project.py run --help` would show the list of operations that a user could supply to the `-o` flag. That was added in 0aa28b5b02acf733ac6604c2d68a73972ae116eb but I couldn't find exactly when it was removed. I would guess it was some time around the introduction of groups, since groups are also valid choices for `-o` and that behavior change might have been the cause for the removal of the feature.
- Use public `self.job` property instead of private data in deprecated class `JobOperation`'s `__repr__` method.
- Use `logger.warning` instead of the deprecated method `logger.warn`.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -1327,9 +1327,6 @@ class _FlowProjectClass(type):
 
             _parent_class = parent_class
 
-            def __init__(self, condition, tag=None):
-                super().__init__(condition, tag)
-
             def __call__(self, func):
                 operation_functions = [
                     operation[1]
@@ -1398,9 +1395,6 @@ class _FlowProjectClass(type):
             """
 
             _parent_class = parent_class
-
-            def __init__(self, condition, tag=None):
-                super().__init__(condition, tag)
 
             def __call__(self, func):
                 operation_functions = [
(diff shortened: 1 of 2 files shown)


flow/project.py
```diff
@@ -1327,9 +1327,6 @@ def hello(job):
 
             _parent_class = parent_class
 
-            def __init__(self, condition, tag=None):
```
bdice, line 1330: These methods are not necessary because they are doing exactly what the parent class does in its constructor. Thus it is fine to inherit the constructor instead of defining it again.

kidrahahjo: I agree by this comment

kidrahahjo (approved):
LGTM

PR #382, opened 2020-11-29
Fix group name check

bdice:
## Description
I noticed a potential issue in `add_operation`. It looks like an operation can be added via `add_operation` that can introduce a conflict with an identical group name. I'd like to get input from @b-butler on this PR before seeking general review.

If this is indeed an issue, I will add a corresponding test.

Also, I'm a bit confused by the usage of `name in self._groups` when there's also a `cls._GROUP_NAMES` attribute. I would appreciate any insight on this from @b-butler or others, since it seems like the two might conflict / overlap / duplicate purpose?

## Motivation and Context
Might fix/prevent a bug.

## Types of Changes
- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -4048,13 +4048,13 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
         :type pre:
             sequence of callables
         """
-        if name in self.operations:
-            raise KeyError("An operation with this identifier is already added.")
-        op = self.operations[name] = FlowCmdOperation(cmd=cmd, pre=pre, post=post)
-        if name in self._groups:
-            raise KeyError("A group with this identifier already exists.")
+        if name in self.operations or name in self._groups:
+            raise KeyError("An operation or group with this name already exists.")
+        operation = self.operations[name] = FlowCmdOperation(
+            cmd=cmd, pre=pre, post=post
+        )
         self._groups[name] = FlowGroup(
-            name, operations={name: op}, operation_directives=dict(name=kwargs)
+            name, operations={name: operation}, operation_directives=dict(name=kwargs)
         )
 
     def completed_operations(self, job):
--- a/flow/project.py
+++ b/flow/project.py
@@ -4128,7 +4128,7 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
     def add_operation(self, name, cmd, pre=None, post=None, **kwargs):
         """Add an operation to the workflow.
 
-        This method will add an instance of :py:class:`~.FlowOperation` to the
+        This method will add an instance of :py:class:`~.FlowCmdOperation` to the
         operations of this project.
 
         .. seealso::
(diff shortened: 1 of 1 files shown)


b-butler:
@bdice the reason for the `name in self._groups` rather than in `cls._GROUP_NAMES` is due to `cls._GROUP_NAMES` only coming from `_FlowProjectClass`. If a project inherits from multiple classes or groups are added after instantiation, we can't rely on `cls._GROUP_NAMES`. `cls._GROUP_NAMES` is to help prevent name duplication, honestly looking at it, a `dict` could be used just to combine `cls._GROUPS` and `cls._GROUP_NAMES`.

b-butler (approved):
LGTM

bdice:
@b-butler This is not a complete PR yet, in terms of solving the underlying issue(s). There's a lot of duplication and potential for errors with `FlowProject._groups`, `FlowProject._GROUPS`, and `FlowProject._GROUP_NAMES`. These are altered in different ways by `add_operation`, the `@FlowProject.operation` decorator, and `make_group`. It seems like we have very little test coverage for these methods, I don't see `add_operation` in any test files.

Below is a rough summary of what's going on.

## `add_operation`
- `self.operations[name] = FlowCmdOperation(...)`
- `self._groups[name] = FlowGroup(name, ...)`
- Doesn't update `cls._OPERATION_FUNCTIONS` or `cls._GROUPS` or `cls._GROUP_NAMES`
- Seems like this won't be inherited (by design?)

## `@operation`
- `cls._OPERATION_FUNCTIONS.append((name, func))`
- `cls._GROUPS.append(FlowGroupEntry(name=name, ...))`
- Doesn't update `cls._GROUP_NAMES` with the singleton group name
- Behavior does not match `add_operation` at all! It was my understanding at the time that the decorator syntax was introduced (in 0.6) that the decorator was designed to do the same thing as `add_operation`, but with a simpler user API. The lack of alignment here makes me think we're probably duplicating code and introducing a high likelihood for confusion/bugs.
- Seems like this _will_ be inherited since it does class-level manipulation.

## `make_group`
- `cls._GROUP_NAMES.add(name)`
- `cls._GROUPS.append(FlowGroupEntry(name=name, ...))`
- I can call `make_group` after defining an `@operation` with the same name and have multiple elements of `cls._GROUPS` with the same name. That seems problematic.

I propose we set up a time to chat through this, perhaps we can summarize the problem in our next dev meeting and then schedule a follow-up discussion to come to a good solution.

vyasr:
@bdice @b-butler There are multiple distinctions to be made here, in part due to historical considerations but also due to some fundamental design choices. I'll try to provide a brief summary, but I think that fixing the underlying issues is out of scope for this PR and may end up involving significant API-breaking changes.

The `add_operation` method was the original API for a `FlowProject`. It is an instance method that was primarily designed for a usage pattern in which a user subclassed a `FlowProject` and overrode its `__init__` method to add all the relevant operations by calling this method. It can also be used more dynamically, but since most usage of flow at that time was based on copy-pasting idioms I very rarely saw much deviation from that pattern.

The decorator pattern was introduced as a simpler way to add operations to a workflow. Making `FlowProject.operation` a classmethod has the benefits of being able to inherit and compose workflows via class inheritance. However, we also wanted to exploit the existing infrastructure for operations in `FlowProject`, meaning that per-class and per-instance operations needed to be unified. In order to support this usage pattern, the all caps variables like `_OPERATIONS` were introduced along with the `_FlowProjectClass` metaclass that ensures that each class has a separate set of operations added via decorator. The `_register_operations` method is then called at instantiation to collect the class-level operations from the MRO and populate the instance-level variable.

When groups were added, they were added to match this pattern, so there are both class-level and instance-level variables to store this. The `make_group` method is class-level and therefore adds to the class level, and `_register_groups` mirrors `_register_operations`. As a result there is now a lot of apparent duplication due to handling class-level and instance-level operations _and_ groups.

It may be possible to reduce some of the smaller redundancies in the code as part of these small PRs. Making larger improvements requires making larger changes concerning the design of the `FlowProject`. @b-butler and I have previously discussed questions on whether operations should be class-level or instance-level, both in the context of addressing some the specific complexities you've noticed and more broadly to enforce the right design paradigm for what type of object best encapsulates a workflow. I think this is a good discussion to have as part of our bigger refactoring efforts. Hopefully it's something that we could address incrementally, but it falls into the bucket of larger refactoring projects that I have not had the bandwidth to tackle.

vyasr:
I'm going to close this for now since, as I explained in the last comment, there are a number of larger refactorings that need to be done to address some of this duplication. We can revisit refactorings after #464 is merged.

bdice:
@vyasr We don't have an issue to track this discussion. We will probably forget about this if we close it, but I don't have time to fix the problems right now.

vyasr:
I don't think these are problems that you can fix right now. They require substantial refactoring that should be chunked into many PRs, with many intermediate steps along the way. I could try and make a roadmap to that point, but it would take time that I don't have right now. I'm not concerned about forgetting this, because these issues are all core to the refactoring that we want to do. It's just a matter of when we get to a point where the issue you reported here can be solved. If you're concerned about tracking, feel free to open an issue pointing here though.

Issue #281, opened 2020-04-04
Speed up parameter formatting in status


vyasr:
### Feature description

I have a somewhat large signac project (~500) jobs with some complicated conditions. I've profiled and cleaned up the conditions, but I was still finding the status updates to be slow. I was able to isolate the problem to the part of the code that handles gathering parameter information (if you run `python project.py status -d -p`). This part of the code is taking ~250 s (about 4 minutes). We should be able to do it faster than that, so I would like to optimize this part of the code.

### Proposed solution

My very rough benchmark was done by just printing the amount of time spent in the code between line 1873 (`if parameters is self.PRINT_STATUS_ALL_VARYING_PARAMETERS`) and line 1902 (`parameters[i] = shorten(self._alias(str(para)), param_max_width)`) in project.py. Offhand I'd guess that the issue either comes from opening many jobs and performing the conversion of the statepoint to a dictionary (line 1882), or from the fact that the `get` function there is implemented as splitting a string and then rejoining it on each recursive call. Since my schema is flat, the former seems more likely. I would recommend creating a large project with a simple schema (something like 5 keys with 4 values each for 1024 total jobs) and then running a profile specifically on that segment of the code.

kidrahahjo:
@vyasr won't this will speed up once the parallelization for fetching status is implemented?

bdice:
Adding a permalink to this code section on the current commit of `master`, for posterity and convenience: https://github.com/glotzerlab/signac-flow/blob/215b252bbda798d2fcb9f0b44917c4fd70e996c5/flow/project.py#L1872-L1902

vyasr:
@kidrahahjo no, the parallelization will speed up the call to `_fetch_status`, but the code that I'm referring to here is run after `_fetch_status`. You can see the difference by running a status update with and without the `-p` option, which triggers the parameter printing but does not change the process of determining the current status of each job.

bdice:
This code is also the only place that `_to_hashable` is used (see PR #384). It seems like this portion of the code could use a thorough refactoring. I have a couple questions. Note that this code is refactored some in #374, but its performance characteristics won't change.

[...]

PR #384, opened 2020-11-30
Make _to_hashable a private function.

bdice:
## Description
The `_to_hashable` helper function should be private, and should match the content of the same helper function in signac.
@atravitz noted in #378 that some type checks are being done in `to_hashable`. I don't know if this behavior should be changed to use `isinstance` but I'm leaving that alone since it matches the behavior in signac.

I have also commented on issue #281 which discusses the parameter formatting code where this helper function is used.

## Motivation and Context
Reducing public API.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -58,10 +58,10 @@ from .util import template_filters as tf
 from .util.misc import (
     TrackGetItemDict,
     _positive_int,
+    _to_hashable,
     add_cwd_to_environment_pythonpath,
     roundrobin,
     switch_to_directory,
-    to_hashable,
 )
 from .util.translate import abbreviate, shorten
 from .version import __version__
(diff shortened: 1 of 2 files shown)


mikemhenry (approved):
LGTM

b-butler:
@bdice This technically makes a public method private, but I doubt anyone is using `misc.to_hashable` at least I hope they aren't  since it isn't documented. Should we just deprecate the non-underscored function?

bdice:
@b-butler I wouldn’t worry about API breakage in this case, since it’s an undocumented helper function in a pre-1.0 codebase.

PR #387, opened 2020-12-06
Add all missing docstrings, enforce pydocstyle.

bdice:
## Description
This PR adds docstrings to all classes and methods that were missing them.
It also enforces pydocstyle (with the same config as `signac`).

## Motivation and Context
Improves docs by ...a lot.

## Types of Changes
- [x] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -15,6 +15,7 @@ Added
 +++++
 - Code is formatted with ``black`` and ``isort`` pre-commit hooks (#365).
 - Add official support for Python version 3.9 (#365).
+- Documentation has been added for all public classes and methods (#387).
 
 Changed
 +++++++
--- a/flow/project.py
+++ b/flow/project.py
@@ -173,7 +173,7 @@ class _condition:
 
     @classmethod
     def true(cls, key):
-        """Evaluate job document key for True value.
+        """Evaluate if a document key is True for the job(s).
 
         Returns True if the specified key is present in the job document(s) and
         evaluates to True.
--- a/flow/project.py
+++ b/flow/project.py
@@ -749,12 +749,14 @@ class FlowOperation(BaseFlowOperation):
 class FlowGroupEntry:
     """A FlowGroupEntry registers operations for inclusion into a :py:class:`FlowGroup`.
 
-    Has two methods for adding operations: the call operator ``self()`` and
-    :meth:`~.with_directives`.  Using :meth:`~.with_directives` takes one
-    argument, ``directives``, which are applied to the operation when running
-    through the group. This overrides the default directives specified by
-    :meth:`flow.directives`. Calling the object directly will then use the
-    default directives.
+    Operation functions can be marked for inclusion into a :class:`FlowGroup`
+    by decorating the functions with a corresponding :class:`FlowGroupEntry`.
+    If the operation requires specific directives, :meth:`~.with_directives`
+    accepts keyword arguments that are mapped to directives and returns a
+    decorator that can be applied to the operation to mark it for inclusion in
+    the group and indicate that it should be executed using the specified
+    directives. This overrides the default directives specified by
+    :meth:`flow.directives`.
 
     :param name:
         The name of the :py:class:`FlowGroup` to be created.
--- a/flow/project.py
+++ b/flow/project.py
@@ -736,6 +736,13 @@ class FlowOperation(BaseFlowOperation):
         return f"{type(self).__name__}(op_func='{self._op_func}')"
 
     def __call__(self, *jobs):
+        r"""Call the operation on the provided jobs.
+
+        :param \*jobs:
+            The jobs passed to the operation.
+        :type \*jobs:
+            One or more instances of :class:`.Job`.
+        """
         return self._op_func(*jobs)
 
 
@@ -773,7 +780,7 @@ class FlowGroupEntry:
     def __call__(self, func):
         """Add the function into the group's operations.
 
-        This call operator allows the group to be used as a decorator.
+        This call operator allows the class to be used as a decorator.
 
         :param func:
             The function to decorate.
--- a/flow/project.py
+++ b/flow/project.py
@@ -186,7 +186,7 @@ class _condition:
 
     @classmethod
     def false(cls, key):
-        """Evaluate job document key for False value.
+        """Evaluate if a document key is False for the job(s).
 
         Returns True if the specified key is present in the job document(s) and
         evaluates to False.
--- a/flow/project.py
+++ b/flow/project.py
@@ -154,7 +154,6 @@ class _condition:
 
     def __init__(self, condition, tag=None):
         """Add tag to differentiate built-in conditions during graph detection."""
-
         if tag is None:
             try:
                 tag = condition.__code__.co_code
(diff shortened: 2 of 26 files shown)


bdice:
@b-butler @kidrahahjo Some of these changes update (or create) documentation for the groups/aggregation features, feel free to weigh in if you would like.

changelog.txt
```diff
@@ -15,6 +15,7 @@ Added
 +++++
 - Code is formatted with ``black`` and ``isort`` pre-commit hooks (#365).
 - Add official support for Python version 3.9 (#365).
+- Documentation has been added for all public classes and methods (#387).
```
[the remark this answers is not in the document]
bdice: That’s okay — the changelog will reflect the current state of the `next` branch, and it will be merged into `master` before a release is made.

flow/project.py
```diff
@@ -174,8 +173,11 @@ def _isfile(*jobs):
 
     @classmethod
     def true(cls, key):
-        """True if the specified key is present in the job document(s) and
-        evaluates to True."""
+        """Evaluate job document key for True value.
```
vyasr, line 176:
```suggestion
        """Evaluate job document key(s) for True value.
```

flow/project.py
```diff
@@ -184,8 +186,11 @@ def _document(*jobs):
 
     @classmethod
     def false(cls, key):
-        """True if the specified key is present in the job document(s) and
-        evaluates to False."""
+        """Evaluate job document key for False value.
```
vyasr, line 189:
```suggestion
        """Evaluate job document key(s) for False value.
```

flow/project.py
```diff
@@ -619,22 +627,23 @@ def eligible(self, job, ignore_conditions=IgnoreConditions.NONE):
         :type job:
             :class:`~signac.contrib.job.Job`
         :param ignore_conditions:
-            Specify if pre and/or post conditions check is to be ignored for eligibility check.
-            The default is :py:class:`IgnoreConditions.NONE`.
+            Specify if pre and/or post conditions check is to be ignored for
+            eligibility check.  The default is
+            :py:class:`IgnoreConditions.NONE`.
```
vyasr, line 632:
```suggestion
            Specify if the pre and/or post condition checks are to be ignored
            for determining operation eligibility. The default is
            :class:`IgnoreConditions.NONE`.
```

flow/project.py
```diff
@@ -727,17 +736,25 @@ def __str__(self):
...
+        :param \*jobs:
+            The jobs passed to the operation.
+        :type \*jobs:
+            One or more instances of :class:`.Job`.
+        """
         return self._op_func(*jobs)
 
 
 class FlowGroupEntry:
     """A FlowGroupEntry registers operations for inclusion into a :py:class:`FlowGroup`.
```
vyasr, line 750:
```suggestion
    """A FlowGroupEntry registers operations for inclusion into a :class:`FlowGroup`.
```

flow/project.py
```diff
@@ -727,17 +736,25 @@ def __str__(self):
...
-    Using ``with_directives`` takes one argument, ``directives``, which are
-    applied to the operation when running through the group. This overrides
-    the default directives specified by ``@flow.directives``. Calling the object
-    directly will then use the default directives.
+    Has two methods for adding operations: the call operator ``self()`` and
+    :meth:`~.with_directives`.  Using :meth:`~.with_directives` takes one
+    argument, ``directives``, which are applied to the operation when running
+    through the group. This overrides the default directives specified by
+    :meth:`flow.directives`. Calling the object directly will then use the
+    default directives.
```
vyasr, line 757:
```suggestion
    Operation functions can be marked for inclusion into a :class:`FlowGroup`
    by decorating the functions with a corresponding :class:`FlowGroupEntry`.
    If the operation requires specific directives, :meth:`~.with_directives`
    accepts keyword arguments that are mapped to directives and returns a
    decorator that can be applied to the operation to mark it for inclusion in
    the group and indicate that it should be executed using the specified
    directives. This overrides the default directives specified by
    :meth:`flow.directives`.
```

vyasr (changes requested):
Some changes, hopefully they can largely be applied directly.

flow/project.py
bdice: This suggestion seems incorrect. The `get` method is called for a single key. The `all(...)` is over jobs, not keys.

flow/project.py
bdice: I made this same change in a later PR -- I'll ignore it here to avoid conflicts.

flow/project.py
bdice: I also stripped `:py:` prefixes in a later PR.

flow/scheduling/simple_scheduler.py
```diff
@@ -10,20 +11,43 @@
 
 
 class SimpleScheduler(Scheduler):
+    """Implementation of the abstract Scheduler class for SimpleScheduler.
```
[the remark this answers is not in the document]
bdice: This suggestion seems incorrect. This is not just a "simple" implementation, it's designed for the included SimpleScheduler. I also may have revised this docstring further in later PRs.

flow/project.py
vyasr: Yeah I agree that my proposed suggested wording is problematic. Perhaps "Check if key is True in all job documents."? The plural is important, but it also needs to qualify the correct object.

flow/scheduling/simple_scheduler.py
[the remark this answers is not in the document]
vyasr: I'm not sure what you mean, this _is_ the included `SimpleScheduler`. Are you referring the `SimpleSchedulerEnvironment` that is used to spoof the detection of this scheduler? Or are you concerned that just calling it a simple implementation of a scheduler doesn't adequately describe what it does?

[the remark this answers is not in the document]
bdice: I mean that this is a scheduler interface for the program [`bin/simple-scheduler`](https://github.com/glotzerlab/signac-flow/blob/master/bin/simple-scheduler). That isn't "a simple implementation of the abstract interface." It's a specific implementation of the scheduler interface for `bin/simple-scheduler`. We should keep the word "simple" next to "scheduler" in some way so that it's clearly a compound noun and not an adjective.

flow/project.py
bdice: I agree with what you're aiming for here. I suggest:
```suggestion
        """Evaluate if key is True in all job documents.
```
but the "all" may be unclear (project-wide vs. just this aggregate).

I like the statement below best, what do you think? If you agree, feel free to accept this suggestion and do the same for `false` below.
```suggestion
        """Evaluate if a document key is True for the job(s).
```

flow/scheduling/simple_scheduler.py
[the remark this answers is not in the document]
vyasr: Oh of course sorry I somehow didn't make that connection from your comment, I now see that you've worded this to match other scheduler classes "Implementation of ...". How about "Implementation of the abstract Scheduler class for the bundled simple-scheduler."? Then add extra lines to the docstring that point to bin/simple-scheduler and give a brief explanation of its purpose.

flow/project.py
vyasr: Yes that seems like the best of the options we've considered.

flow/scheduling/simple_scheduler.py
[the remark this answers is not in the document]
bdice: Done in 1fb7b42a84f2bfe2c1b830741f7cba37e6c54ce1. Thanks for the feedback!

vyasr (approved):
LGTM, thanks!

PR #389, opened 2020-12-07
Revise Sphinx docs.

bdice:
## Description
Builds on #387 and fills in a lot of missing classes/methods in the Sphinx documentation.

Almost all the changes here are intended to make the Sphinx docs more complete, more inter-linked, or render better.

I made a few small API breaks along the way:
- Renamed `JobStatus.dummy` to `JobStatus.placeholder` to better indicate its purpose with a kinder word.
- Removed all APIs imported into the subpackage `flow.scheduling`. These short-hand imports in `scheduling/__init__.py` were never used internally, the contents were inconsistent (not all schedulers or public methods were defined), and the Sphinx docs did not render references to the correct classes from those short-hand names.

**Reviewers:** I recommend looking at the rendered Sphinx docs, you'll see a lot of new content there in the API section.

## Motivation and Context
Resolves ... some open issues, probably.

## Types of Changes
- [x] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -1845,7 +1845,7 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
         for status, symbol in _FMT_SCHEDULER_STATUS.items()
         if status != JobStatus.dummy
     }
-    # These are default aliases used within the status output.
+    """Default aliases used within the status output."""
 
     @classmethod
     def _alias(cls, name):
--- a/flow/project.py
+++ b/flow/project.py
@@ -228,12 +228,13 @@ def _make_bundles(operations, size=None):
     This function splits an iterable of operations into equally
     sized bundles and a possibly smaller final bundle.
     """
-    n = None if size == 0 else size
+    if size == 0:
+        size = None
     operations = iter(operations)
     while True:
-        b = list(islice(operations, n))
-        if b:
-            yield b
+        bundle = list(islice(operations, size))
+        if bundle:
+            yield bundle
         else:
             break
 
@@ -574,8 +575,8 @@ class BaseFlowOperation:
         if post is None:
             post = []
 
-        self._prereqs = [_FlowCondition(cond) for cond in pre]
-        self._postconds = [_FlowCondition(cond) for cond in post]
+        self._preconditions = [_FlowCondition(cond) for cond in pre]
+        self._postconditions = [_FlowCondition(cond) for cond in post]
 
     def __str__(self):
         return "{type}(cmd='{cmd}')".format(type=type(self).__name__, cmd=self._cmd)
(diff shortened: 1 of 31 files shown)


flow/scheduling/__init__.py
```diff
@@ -2,14 +2,3 @@
 # All rights reserved.
 # This software is licensed under the BSD 3-Clause License.
 """Defines the API for the scheduling system."""
```
vyasr, line 4: Is the intent here to completely remove schedulers from the public API? That seems dramatically out of scope for this PR.

flow/scheduling/base.py
```diff
@@ -10,16 +10,41 @@ class JobStatus(enum.IntEnum):
...
     active = 7
+    """The cluster job is actively running."""
+
     error = 8
-    dummy = 127
-    # All user statuses are >= 128.
+    """The cluster job is in an error or failed state."""
+
+    placeholder = 127
+    """A placeholder state that is used internally."""
```
vyasr, line 44: I would specify the purpose of this status here since it exists for only one purpose and we don't want to make it seem like something random that just gets added.

flow/project.py
```diff
@@ -1507,18 +1507,18 @@ def hello(job):
...
     :type environment:
         :class:`flow.environment.ComputeEnvironment`
     :param entrypoint:
-        A dictionary with two possible keys: executable and path. Path
-        represents the filepath location of the script file (the script file
-        must call ``main``). Executable represents the location of the Python
-        interpreter used for the executable of `FlowOperation` that are Python
-        functions.
+        A dictionary with two possible keys: ``'executable'`` and ``'path'``.
+        The path represents the filepath location of the script file (the
```
vyasr, line 1518:
```suggestion
        The path represents the location of the script file (the
```

flow/project.py
```diff
@@ -2441,7 +2441,7 @@ def print_status(
         :param jobs:
             Only execute operations for the given jobs, or all if the argument is omitted.
         :type jobs:
-            Sequence of instances of :class:`.Job`.
+            Sequence of instances of :class:`~signac.contrib.job.Job`.
```
vyasr, line 2444: Why do we need the fully qualified name here (and elsewhere that you've changed it)?

flow/project.py
```diff
@@ -3101,13 +3101,12 @@ def run(
...
 
-        See also: :meth:`~.run_operations`
-
         :param jobs:
             Only execute operations for the given jobs or aggregates of jobs,
             or all if the argument is omitted.
         :type jobs:
-            Sequence of instances of :class:`.Job` or aggregate of instances of :class:`.Job`.
+            Sequence of instances of :class:`~signac.contrib.job.Job` or
+            aggregate of instances of :class:`~signac.contrib.job.Job`.
```
vyasr, line 3109:
```suggestion
            Sequence of instances of :class:`~signac.contrib.job.Job` or
            sequence of aggregates where each aggregate is a sequence
            of :class:`~signac.contrib.job.Job`.
```

vyasr (changes requested):

flow/scheduling/__init__.py
bdice: They're still in the public API -- but users must specify them with a full name: `flow.scheduling.torque.TorqueScheduler`. The reason I did this is that importing the Scheduler subclasses into the `scheduling` namespace (like `flow.scheduling.TorqueScheduler`) causes Sphinx to fail to find references to the right classes. The references would find `flow.scheduling.TorqueScheduler` but wouldn't link to the corresponding class docs. These classes were not mentioned in the documentation _at all_ before this PR, so while I recognize it's an API break, I'm not too worried about making this change. All internal code already used the full namespace and not the short form. If you know how to solve this issue without the API break, I'd be happy to learn more.

flow/scheduling/base.py
bdice:
```suggestion
    """A placeholder state that is used for status rendering when no operations are eligible."""
```

flow/project.py
bdice: I think it needs a full path for the intersphinx reference, right? I found some cases where the references weren't rendering correctly, but there may be a shorter way to write this. I decided to make every reference to the signac Job class match the working reference that I found. We can find-and-replace this with a shorter form if you have suggestions.

flow/scheduling/__init__.py
vyasr: *sigh OK yeah I have fixed this problem before but I don't recall how. Fiddling with Sphinx usually takes me 5 mins and a few tries to get just right. For now I'm OK with this solution, I agree that it's not too meaningful an API break. Since deprecating schedulers from the public API is something we've discussed recently I didn't want that to just happen without further discussion, but this particular change is fine with me for now.

flow/project.py
vyasr: Hmm OK this seems related to the scheduler question. I think there are better ways to disambiguate correctly without requiring fully qualified names, but since there's a string of these PRs right now I'd prefer to just merge these as is and improve linking separately.

vyasr (approved):
Once #387 is done and conflicts with this PR are resolved, I think this is ready.

PR #390, opened 2020-12-08
Fix Iteration Bug on next and add AggregatesCursor to flow

kidrahahjo:
## Description
New JobOperations created due to dynamic job addition were not considered as "eligible" when we filtered the the `_JobOperations` while executing `run`. This was because of the way we selected jobs from CLI using the method `_select_jobs_from_args`. It returned an instance of `JobsCursor` when it was supposed to return `None`. 

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [x] New feature

## Checklist:

If necessary:


kidrahahjo:
I feel the current behaviour is kind-of hacky and can easily fix this by refactoring `JobsCursor`. Since I'm not very familiar with the class, I had a question that up to what extent is this class used by our users?

vyasr:
> I feel the current behaviour is kind-of hacky and can easily fix this by refactoring `JobsCursor`. Since I'm not very familiar with the class, I had a question that up to what extent is this class used by our users?

Users never directly instantiate `JobsCursor` objects, but they interact with them frequently via `project.find_jobs`. [There's a discussion](https://github.com/glotzerlab/signac-flow/pull/335/files#r504274518) on #335 where I proposed updating the `JobsCursor` to an `AggregateCursor` to avoid having the hacky `return None` solutions that aggregation currently requires. Since `JobsCursor` is a `signac` object (not a `flow` object), we'd have to define a new class to handle this in `flow` anyway. I would be fine with doing that as part of our ongoing refactor, many of the methods could hopefully make extensive use of `yield from JobsCursor...` and then `yield from aggregates` or a similar pattern to avoid too much code repetition. However, we cannot change the underlying `JobsCursor` API.

kidrahahjo:
@vyasr If I get it correctly, do you mean that we could handle this using an internal class `_AggregatesCursor` (acting as a `flow` object) where we could take some advantage of the functionality provided by `JobsCursor` (filter/docfilter)?

It'd be something like

```
class _AggregatesCursor(JobsCursor):
    # code goes here where we fetch and iterate over aggregates according to our need
```

vyasr:
Yes basically that, except that it would probably have to be
```
class AggregatesCursor:
    self.jobs_cursor = JobsCursor
```

i.e. it should 1) probably be publicly visible because users may directly interact with methods like `groupby`, and 2) it would no longer _be_ a `JobsCursor`, but rather something that can either iterate over jobs or aggregates.

kidrahahjo:
sounds good to me!
I'll start the implementation part of it.

kidrahahjo:
@vyasr I added the `AggregatesCursor` class, please have a look at it once you get time.

vyasr:
@kidrahahjo this solution seems fine and resolves the current problem. I anticipate that the AggregatesCursor class will need to become substantially more sophisticated as we continue the refactoring of flow, but for now this resolves the problem. It looks like right now the AggregatesCursor is only used internally, you haven't changed any methods returning JobsCursors to the user, right? Can we just make the class private for now?

kidrahahjo:
@vyasr I believe no method in flow returns the instance of `JobsCursor` hence I'm also in favour of making this class private.
From a broader perspective, I think the `jobs_cursor` (the attribute in `AggregatesCursor` which points to all the aggregates present) will eventually point out to the `_stored_aggregated` attribute in flow so that it could iterate over all the aggregates. And apart from this, features like aggregate filters can also be added to this class.

bdice (commented):
I forgot to submit this review earlier. I will take another look at this code -- it's affecting performance for a project of mine.

flow/project.py
```diff
@@ -260,6 +260,44 @@ def _make_bundles(operations, size=None):
             break
 
 
+class _AggregatesCursor:
+    """Utility class to iterate over aggregates stored in a FlowProject.
+
+    Parameters
+    ----------
+    project : :class:`~.FlowProject`
+        A FlowProject
```
bdice, line 269:
```suggestion
        A FlowProject whose jobs are aggregated.
```

flow/project.py
```diff
@@ -2605,15 +2646,9 @@ def print_status(
...
-            # Fetch all the distinct jobs from all the jobs or aggregate passed by the user.
-            distinct_jobs = set()
-            for aggregate in aggregates:
-                for job in aggregate:
-                    distinct_jobs.add(job)
-        else:
-            distinct_jobs = self
+        aggregates = self._convert_jobs_to_aggregates(jobs)
+        # Fetch all the distinct jobs from all the jobs or aggregate passed by the user.
+        distinct_jobs = set(job for job in aggregate for aggregate in aggregates)
```
bdice, line 2651:
```suggestion
        distinct_jobs = set(job for aggregate in aggregates for job in aggregate)
```

vyasr:
> I forgot to submit this review earlier. I will take another look at this code -- it's affecting performance for a project of mine.

I assume this is referring to the change in glotzerlab/signac#449?

vyasr:
Aside from the performance considerations that @bdice is working on I think this PR is fine. I can provide additional help if there are further questions on functionality, but I'll leave benchmarking to you and @bdice since he's working on it actively right now.

bdice:
> 
> 
> > I forgot to submit this review earlier. I will take another look at this code -- it's affecting performance for a project of mine.
> 
> I assume this is referring to the change in [glotzerlab/signac#449](https://github.com/glotzerlab/signac/pull/449)?

Not related. I was just seeing poor performance on `next` and this PR's implementation of the aggregate cursor fixes it. The work in [glotzerlab/signac#449](https://github.com/glotzerlab/signac/pull/449) is just an additional optimization.

vyasr:
Oh gotcha. Yes, this PR will fix some issues that are currently present on next. If this PR is improving performance for you then I think we're good. Let me (or of course @kidrahahjo) know if you need clarification or help on anything while reviewing.

bdice (approved):
I have worked with this PR for quite a while and it's ready to merge into `next`. @vyasr said this would be fine, in the interest of moving faster towards a release and fixing the performance degradations on `next`.

PR #394, opened 2020-12-09
Reduce calls to getfqdn

### LINKED ISSUE #339: Slow command line operations on macOS traced to external function in environment detection
## Takeaway

This isn&rsquo;t a problem with signac-flow, but we could mitigate it by calling `socket.get_fqdn()` less. [ Jump to workaround.](#orgbc1f967)

<a id="orga8dc9b5"></a>

## Problem

On a newly created signac project, calls to `project.py [run | status | ...]` were taking at least 25 seconds to run.

I profiled the code by running `python -m cProfile -o out.prof project.py status` and found that the function `socket.getfqdn()` (get fully qualified domain name) was the root cause. It&rsquo;s called 5 times as part of `ComputeEnvironment.is_present()`, once each for each defined environment, as part of automatic environment determination in `get_environment()`.

I isolated this issue with:

    $ time python -c 'import socket; print(socket.getfqdn())'
    
    myComputerName.local
    
    real	0m5.050s
    user	0m0.029s
    sys	0m0.011s

The problem has been documented [on Stack Exchange](https://apple.stackexchange.com/questions/175320/why-is-my-hostname-resolution-taking-so-long) and could be related to [this Python bug report](https://bugs.python.org/issue35164#msg359718).

I&rsquo;m on macOS 10.14.6 running signac 1.4.0, flow 0.10.0 and Python 3.7.3.

<a id="org3bdee4e"></a>

## Attempted solution failed

I had the problem where the output of `scutil --get HostName` returned  `HostName: not set`. Note that the command `hostname` worked normally. Since LocalHostName was set, I followed the recommend procedure to fix this with `scutil --set HostName $(scutil --get LocalHostName)`. This command produced the expected result, but `getfqdn()` was still slow.

<a id="orgbc1f967"></a>

## Workaround

Since `get_environment()` checks environment variables before checking for the fully qualified domain name, I set the environment variable `SIGNAC_FLOW_ENVIRONMENT` to StandardEnvironment using

    export SIGNAC_FLOW_ENVIRONMENT=StandardEnvironment

<a id="orgacdb037"></a>

## Possible mitigations

1.  We could mitigate this issue by storing the output of `socket.getfqdn()`, so it wouldn&rsquo;t have to be called 5 or more times.

2.  Just like we have a warning for labels that take a long time to evaluate, perhaps we could time the call to `getfqdn()` and point users to this issue if it takes more than 1 s.

cbkerr:
Adds `@lru_cache`ing of socket.getfqdn to reduce wait when it times out.

## Description

Thanks to Bradley for help along the way

## Motivation and Context

Resolves #339

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -23,7 +23,7 @@ Changed
 Fixed
 +++++
 
-- Automatic environment detection no longer hangs as long if socket.fqdn times out (#339)
+- Cache the fully qualified domain name during environment detection to fix a performance issue on macOS (#339, #394)
 
 Version 0.11
 ============
--- a/changelog.txt
+++ b/changelog.txt
@@ -20,6 +20,10 @@ Changed
 +++++++
 - Command line interface for ``exec`` changed parameter name from ``jobid`` to ``job_id`` (#363).
 
+Fixed
++++++
+
+- Automatic environment detection no longer hangs as long if socket.fqdn times out (#339)
 
 Version 0.11
 ============
--- a/changelog.txt
+++ b/changelog.txt
@@ -23,7 +23,7 @@ Changed
 Fixed
 +++++
 
-- Cache the fully qualified domain name during environment detection to fix a performance issue on macOS (#339, #394)
+- Cache the fully qualified domain name during environment detection to fix a performance issue on macOS (#339, #394).
 
 Version 0.11
 ============
--- a/flow/environment.py
+++ b/flow/environment.py
@@ -44,8 +44,11 @@ logger = logging.getLogger(__name__)
 
 @lru_cache(maxsize=1)
 def _cached_fqdn():
-    """We cache this since fetching the fully qualified domain name can be
-    slow on macOS."""
+    """Return the fully qualified domain name.
+    
+    This value is cached because fetching the fully qualified domain name can
+    be slow on macOS.
+    """
     return socket.getfqdn()
 
 
--- a/flow/environment.py
+++ b/flow/environment.py
@@ -15,8 +15,8 @@ import logging
 import os
 import re
 import socket
-from functools import lru_cache
 from collections import OrderedDict
+from functools import lru_cache
 
 from signac.common import config
 
--- a/flow/environment.py
+++ b/flow/environment.py
@@ -40,6 +40,20 @@ from .util import config as flow_config
 
 logger = logging.getLogger(__name__)
 
+_fqdn = None
+def _cached_fqdn():
+    """Returns the fully qualified domain name.
+
+    If called for the first time in a session, call socket.getfqdn
+    and cache the value. If called again, return the cached value.
+    Solves #339
+    """
+    if _fqdn is None:
+        _fqdn = socket.getfqdn()
+        import pdb
+        pdb.set_trace()
+    return _fqdn
+
 
 def setup(py_modules, **attrs):
     """Setup function for environment modules.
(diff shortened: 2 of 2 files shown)


flow/environment.py
```diff
@@ -15,6 +15,7 @@
 import os
 import re
 import socket
+from functools import lru_cache
```
bdice, line 18: The style checks failed here. We use a tool called pre-commit to manage code style, and we have configured a pre-commit hook that uses the package `isort` to make sure that import statements are sorted (instead of randomly arranged). If you look at the CI output, it says to move this down (it alphabetizes the statements `from collections...` and `from functools...`).

See this guide to install pre-commit: https://docs.signac.io/projects/flow/en/latest/support.html#set-up-a-development-environment

changelog.txt
```diff
@@ -20,6 +20,10 @@ Changed
 +++++++
 - Command line interface for ``exec`` changed parameter name from ``jobid`` to ``job_id`` (#363).
 
+Fixed
++++++
+
+- Automatic environment detection no longer hangs as long if socket.fqdn times out (#339)
```
bdice, line 26:
```suggestion
- The fully qualified domain name is cached during environment detection, fixing a performance issue on macOS (#339, #394).
```

flow/environment.py
```diff
@@ -41,6 +42,17 @@
 logger = logging.getLogger(__name__)
 
 
+@lru_cache(maxsize=1)
+def _cached_fqdn():
+    """Returns the fully qualified domain name.
+
+    If called for the first time in a session, call socket.getfqdn
+    and cache the value. If called again, return the cached value.
+    Solves #339
```
bdice, line 51: We typically do not put issue/PR references in the code, just the changelog. You could instead write:
```suggestion
    This value is cached because fetching the fully qualified domain name can
    be slow on macOS.
```

bdice (changes requested):
Thanks @cbkerr. Some change requests attached.

changelog.txt
cbkerr: Added in active voice!

flow/environment.py
cbkerr: Didn't catch what the diff was telling me, thanks!

changelog.txt
```diff
@@ -20,6 +20,10 @@ Changed
 +++++++
 - Command line interface for ``exec`` changed parameter name from ``jobid`` to ``job_id`` (#363).
 
+Fixed
++++++
+
+- Cache the fully qualified domain name during environment detection to fix a performance issue on macOS (#339, #394)
```
bdice, line 26:
```suggestion
- Cache the fully qualified domain name during environment detection to fix a performance issue on macOS (#339, #394).
```

flow/environment.py
```diff
@@ -41,6 +42,13 @@
 logger = logging.getLogger(__name__)
 
 
+@lru_cache(maxsize=1)
+def _cached_fqdn():
+    """We cache this since fetching the fully qualified domain name can be
+    slow on macOS."""
```
bdice, line 48: @cbkerr Sorry, I probably wasn't clear about what I meant. The docstring you had before was mostly correct. I'm going to apply the following changes. You can see these [recommendations from pydocstyle with the corresponding codes](http://www.pydocstyle.org/en/stable/error_codes.html):
- Summary line (first line) shouldn't wrap. D205.
- Summary line should be in imperative mood (starts with a verb like "Gets" or "Computes"). D401.
- Blank line after the first line, then additional info. D205.
- Avoid first or second person words like we/you (not a pydocstyle recommendation, just my preference.)

We don't run pydocstyle right now but we will require pydocstyle to pass in the near future (#387).
```suggestion
    """Return the fully qualified domain name.
    
    This value is cached because fetching the fully qualified domain name can
    be slow on macOS.
    """
```

bdice (approved):

flow/environment.py
```diff
@@ -41,6 +42,16 @@
 logger = logging.getLogger(__name__)
 
 
+@lru_cache(maxsize=1)
+def _cached_fqdn():
+    """Return the fully qualified domain name.
+
```
bdice, line 48:
```suggestion

```

PR #399, opened 2020-12-15
Add benchmarks / profiling code.

bdice:
## Description
Added benchmarking / profiling code. This code is largely copied from signac.

The sample workflow involves two labels and three operations, and it's based on the built-in template for a test project.

I did not include aggregation or groups, because I want to be able to test this workflow against older versions of signac-flow so we know that the performance is not regressing compared to those older (simpler) versions.

I might want to refactor the benchmarks/profiling a bit so that they use the same project definition. I wanted to copy most of the CLI infrastructure from the benchmarking code to the profiling code. I can't use the benchmarking code directly for profiling, because it uses `timeit` which appears to run separately somehow (thread? process? other?) and shows no call stack when using cProfile.

## Motivation and Context
I needed a standard testing workflow to use for profiling and performance improvement.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


vyasr:
@bdice do you think you'll be able to finish up this PR sometime soon? Some of the benchmarks (such as iteration over the project) may also be helpful to add directly to signac's benchmarks (instead of flow). You're already evaluating a number of PRs based on running such benchmarks locally, so finishing this up may save you that effort.

bdice:
@vyasr I stalled on this because I don't really know what I want "finished" to look like for this PR. It feels more like a handful of random scripts and I didn't see a clear path to make it generally usable. Would someone want to offer thoughts on what is worth including or removing?

bdice:
It seems I have upset CircleCI by rewriting all of the history on this branch. We might need to open a new PR for this...

vyasr:
Could you elaborate on what you think the problems are wrt making this generally useful? Is the issue defining a test FlowProject? What makes this different from the benchmarks in signac?

I'm also wondering whether we should add benchmarks back to CI (for signac as well). We removed it because the numbers on CI are very inconsistent, but maybe it's worth adding back as a step not required for merging so that users can at least see when their changes will slow the code down. All your recent work on benchmarking and optimization highlightst that we've probably been ignoring performance for too long.

bdice:
> Could you elaborate on what you think the problems are wrt making this generally useful? Is the issue defining a test FlowProject? What makes this different from the benchmarks in signac?

The benchmarking code and profiling code are basically copy-pasted but have different execution paths (the profiled code can't run through `timeit` because that hides the profile -- it runs as a different process, maybe? idk).

I only used the profiling code for the optimizations I worked on. Since I didn't even use the `timeit` benchmarks and I only have a "profiling script" that I think is useful, I'm unsure what is most valuable and/or worth maintaining.

> I'm also wondering whether we should add benchmarks back to CI (for signac as well). We removed it because the numbers on CI are very inconsistent, but maybe it's worth adding back as a step not required for merging so that users can at least see when their changes will slow the code down. All your recent work on benchmarking and optimization highlightst that we've probably been ignoring performance for too long.

I have basically no trust in the CI benchmarks, there have been so many false failures in the past. I think the maintenance cost of dealing with CI inconsistency has been too high in the past to justify the benefit of catching slowdowns. I'm fine with keeping it a local-only check that is done infrequently and/or for performance-critical refactoring.

benchmarks/benchmark.py
```diff
@@ -0,0 +1,699 @@
...
+
+    with TemporaryDirectory(dir=root) as tmp:
+        project = signac.init_project(f"benchmark-N={N}", root=tmp)
+        generate_random_data(project, N, num_keys, num_doc_keys, data_size, data_std)
+        yield project
+
+
+class Timer(timeit.Timer):
+    def timeit(self, number=10):
+        return number, timeit.Timer.timeit(self, number=number)
```
vyasr, line 184: Why not user `super().timeit`?

vyasr:
If you defined a custom `Timer` class in `profiling.py` that replaces the `timeit` operations with simple loops, would you be able to reuse all the other logic from `benchmarking.py`? I'm imagining something like:

```
class Timer:
    def __init__(self, func...):
        ...

    def timeit(self, number...):
        for _ in range(number):
            self._func(...)

def benchmark_SOMETHING(..., timer=[timeit.Timer|Timer]):
    ...
```

I agree, CI runs aren't an option without dedicated hardware, otherwise I'd love to get something like airspeed velocity set up to maintain running performance benchmarks that would let us catch regressions. For now, I would settle for adapting the `benchmarker.py` script in `freud`, which allows us to compare git revisions, then including a run of some predefined benchmarks (which we will iron out in this PR) as part of the release process. That way we could catch regressions before releasing. How does that sound? I think the most important thing to avoid is relying on the institutional knowledge of a few contributors to guide what code paths are performance-critical. If we think something is important enough to benchmark the effects of PRs against it, we should codify it and make it easy to verify before releasing a serious regression into the wild. Whatever process we decide on for catching regressions should also be used for `signac`.

With that in mind, here's what I would suggest once you've removed the benchmarks that are duplicated from signac (A lot of this logic already exists in some of your files, I'm just laying out how I think the final product could look):
1. Extract the `BenchmarkProject` logic into a separate file.
2. Define various `FlowProject` classes with different numbers of operations, labels, etc, e.g. `BenchmarkProject` has one op, `BenchmarkProject10` has 10 ops, and `BenchmarkProject100` has 100 operations). You should be able to do this programmatically without too much difficulty, something like
```
def make_flow_project(num_operations, num_labels, ...):
    class BenchmarkProject(FlowProject):
        pass

    for i in range(num_operations):
        @BenchmarkProject.operation
        # some label-adding logic
        def op(job):
            ...
    return BenchmarkProject
```
3. Define signac projects with various numbers of jobs.
4. Define a custom `Timer` that has the same API as `timeit.Timer` (or close enough for our purposes).
5. Move all shared logic for benchmarking/profiling into a common file.
6. Right now, your `run` function takes an instance of `Timer` as an argument. Similarly, you can make the function that contains it (the `benchmark_project` function) accept a class argument for what to use as the `Timer`.
6. In `benchmarking.py` use all the common functions while passing `timeit.Timer` to `benchmark_project`.
7. In `profiling.py` use all the common functions while passing `timeit.Timer` to `benchmark_project`.

Does that sound reasonable? Let me know if I'm missing something, or if there are important problems that I'm overlooking.

bdice:
@vyasr I agree with everything you suggested. However, I don’t know when I will get back to working on this PR. I don’t see it as very important or urgent, so I’m probably going to postpone my efforts on it for now.

vyasr:
That's fine with me. However, I would strongly advocate working on this the next time someone makes a PR that you think requires benchmarking (in which case you might also be able to get somebody to help you finish this up). That will allow us to move towards a more systematic process for evaluating the performance implications of such PRs than what we have now.

bdice:
@vyasr Yes, I agree. I would have a greater sense of importance/urgency for this when dealing with future PRs that affect performance. I just don’t see myself putting this ahead of anything else I want to get done in the next month or so.

benchmarks/benchmark.py
bdice:
```suggestion
        return number, super().timeit(self, number=number)
```

bdice:
I'm going to close this PR. I haven't had time to work on it, and I would prefer a solution based on asv (airspeed velocity) like what we have currently in signac.

PR #400, opened 2020-12-15
Optimization: Don't use min_len_unique_id in str(_JobOperation).

bdice:
## Description
Don't call `Project.min_len_unique_id` when converting `_JobOperation` to a string (presumably for logging or script generation). It's O(N) and is called O(N) times, leading to O(N^2) behavior and significant performance penalties.

Related to https://github.com/glotzerlab/signac/pull/434.

The regression was introduced in #334. **This does not need a changelog line because the code where the regression was introduced has not been released.**

## Motivation and Context
In the "run" benchmarks I'm using (1000 jobs, 3 operations with preconditions/postconditions, 2 labels), `min_len_unique_id` takes 19.4 seconds out of a total runtime of 37.8 seconds (over 50%!). Removing this call and printing the full job id decreases the total runtime to 18.1 seconds.

The performance cost of this call may also be reduced by PR #396, if the `__str__` method of `_JobOperation` is called less frequently.

### Profile before:
image

### Profile after:
image

(There is a tiny, tiny highlighted pink sliver on the right side. That's the `__str__` operation without the expensive call to `min_len_unique_id`.)

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:


atravitz:
This looks very worthwhile to me! Was the reason for this in the first place just for concise output?

cbkerr:
Looks good to me too

b-butler (approved):
LGTM

bdice:
@atravitz Yes, it was just used for concise output.

PR #402, opened 2020-12-17
Get executable directive at call time instead of definition time

### LINKED ISSUE #398: Executable path links to the local environment when generating template reference data
### Description
In generate_template_reference_data.py we have a method called get_masked_flowproject which we use to create the project. There, we mock `sys.executable` before instantiating the project.

It seems like this is a problem related to the `directives` module. We initialize _EXECUTABLE in that module with the default value of `sys.executable`. I think that value is stored before we mock `sys.executable` hence the default value of an executable path always refers to the local environment.

### To reproduce
Execute 
`python -f test/generate_template_reference_data.py`
And test it with
`pytest test/test_templates.py`

### Possible Solution
Reset default value of the `_EXECUATABLE` variable to `sys.executable` when we return the default values of all the directives in `environment.py`
@glotzerlab/signac-maintainers please weigh in.

PS. thanks to @ramanishsingh for pushing a PR, that way I got to learn about this problem.

kidrahahjo:
## Description

This pull request evaluates the executable at call time which previously was evaluated at definition time.

## Motivation and Context
Fixes #398 

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/directives.py
+++ b/flow/directives.py
@@ -354,23 +354,6 @@ _OMP_NUM_THREADS = _Directive("omp_num_threads", validator=_nonnegative_int, def
 Expects a nonnegative integer.
 """
 
-_EXECUTABLE = _Directive(
-    "executable",
-    validator=_OnlyType(str),
-    default=sys.executable,
-    serial=_no_aggregation,
-    parallel=_no_aggregation,
-)
-"""The path to the executable to be used for this operation.
-
-Expects a string pointing to a valid executable file in the
-current file system.
-
-By default this should point to a Python executable (interpreter); however, if
-the :class:`FlowProject` path is an empty string, the executable can be a
-path to an executable Python script. Defaults to ``sys.executable``.
-"""
-
 _WALLTIME = _Directive(
     "walltime",
     validator=_nonnegative_real,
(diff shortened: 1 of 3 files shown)


bdice (approved):
This is an interesting discovery and a good solution, @kidrahahjo. Two alternatives:
- We could explicitly declare the `executable` directive for the test project (might need to assign it to every operation?)
- We could mock both `sys.executable` and `flow.directives._EXECUTABLE` in the test script generation.
This PR looks fine to me unless one of those alternatives is preferred by other reviewers.

vyasr:
> This is an interesting discovery and a good solution, @kidrahahjo. Two alternatives:
> 
>     * We could explicitly declare the `executable` directive for the test project (might need to assign it to every operation?)
> 
>     * We could mock both `sys.executable` and `flow.directives._EXECUTABLE` in the test script generation.
> 
> 
> This PR looks fine to me unless one of those alternatives is preferred by other reviewers.

@bdice these are good alternatives to my solution. The problem with option 2 is similar what I raised in #398: it propagates to any further runs, including if you run other tests that depend on having the correct default executable. However, your suggestions do raise what I think is an important question to ask: is any reason to support modification of `sys.executable` by the user? If the answer is no (probably the right answer), then maybe we should prefer a solution that more directly modifies the internals and doesn't change how the code behaves if users modify the executable.

This is also a minor problem and we could just accept the existing solution unless we see it cause problems (I don't think it should, user modification of sys.executable is an edge case we probably don't need to worry about anyway).

bdice:
> > This PR looks fine to me unless one of those alternatives is preferred by other reviewers.

I think this is fine to merge as-is. It fixes the problem and its purpose is clearly documented. I'll accept this as an approval and merge this PR.

PR #403, opened 2020-12-17
Deprecate script command line argument

### LINKED ISSUE #311: Deprecate the script command

kidrahahjo:
## Description

This pull request deprecates the script argument from the command line interface. This argument will be removed in `flow` version _0.14_.
Users are now advised to use `python project.py submit --pretend` instead.

## Motivation and Context

Fixes #311 
The `script` method of python API was deprecated in `flow` version _0.11_.
For _0.12_, in #311 it was decided to deprecate this argument.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -4116,7 +4116,7 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
             "WARNING: "
             "The script argument is deprecated as of 0.12 "
             "and will be removed in 0.14. "
-            "Use `python project.py submit --pretend` instead.",
+            "Use \"submit --pretend\" instead.",
             file=sys.stderr,
         )
 
--- a/flow/project.py
+++ b/flow/project.py
@@ -4112,6 +4112,14 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
 
     def _main_script(self, args):
         "Generate a script for the execution of operations."
+        print(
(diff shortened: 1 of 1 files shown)


atravitz (approved):

flow/project.py
```diff
@@ -4112,6 +4112,14 @@ def _main_run(self, args):
 
     def _main_script(self, args):
         "Generate a script for the execution of operations."
+        print(
+            "WARNING: "
+            "The script argument is deprecated as of 0.12 "
+            "and will be removed in 0.14. "
+            "Use `python project.py submit --pretend` instead.",
```
bdice, line 4119: Not every user will have "python project.py." I'll commit this.
```suggestion
            "Use \"submit --pretend\" instead.",
```

bdice (approved):
Thanks @kidrahahjo!

PR #409, opened 2020-12-22
Evaluate directives in _JobOperation Constructor

### LINKED ISSUE #408: Bug in directives nranks/np
`@directives` for nranks and np fail to be set when lambda function is used.
@b-butler  maybe able to help with this.

### Description

In version 0.11.0 using submit --pretend in Slurm environments with `@directives(nranks = lambda job: job.doc.nranks)` gives the following error (from --debug):
 ```File "/home1/07885/dfijan/.local/lib/python3.7/site-packages/flow/util/template_filters.py", line 74, in <listcomp>
    op.directives.get('processor_fraction', 1) for op in operations]
TypeError: unsupported operand type(s) for *: 'function' and 'float'
```
When using `@directives(nranks = 16)` everything works fine.

When using on SimpleSchedulerEnvironment supplied by signac everything works. Issue seems to arise when using SLURM (tested 2 different clusters).

Version 0.10.1. version of flow does not have this problem. 

### To reproduce
```
# init.py:
import signac

project = signac.init_project('MyBug')

sp = {"test": 1}

job = project.open_job(sp).init()
job.doc.nranks=2
```
```
# project.py:
import flow
import signac
from flow import FlowProject, directives, cmd, environments

class Project(FlowProject):
    pass

@Project.operation
@directives(nranks= lambda job: job.doc.nranks)
def initialise(job):
    pass


if __name__ == "__main__":
    pr = Project()
    pr.main()
```
```
python3 init.py
python3 project.py submit --pretend
```
### Error output

```File "/home1/07885/dfijan/.local/lib/python3.7/site-packages/flow/util/template_filters.py", line 74, in <listcomp>
    op.directives.get('processor_fraction', 1) for op in operations]
TypeError: unsupported operand type(s) for *: 'function' and 'float'
```
### System configuration

Cluster: Stampede2
OS: Linux-3.10.0-957.5.1.el7.x86_64-x86_64-with-centos-7.6.1810-Core
Python: 3.7.0 (default, Feb  6 2019, 21:24:19)
             [GCC Intel(R) C++ gcc 6.3 mode]
Signac:1.5.1
Signac-flow:0.11.0

b-butler:
## Description

Evaluate the directives in the `_JobOperation` constructor. This should fix #408. This does mean that all directives will be evaluated at the construction of `_JobOperation`. I don't think this should pose any problems, and indeed it should have been evaluated there to begin with since we change class in `_JobOperation`.

## Motivation and Context

Fixes #408.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -26,7 +26,7 @@ Changed
 Fixed
 +++++
 
-- Ensure that directives are always evaluated before running or submitting (#408).
+- Ensure that directives are always evaluated before running or submitting (#408, #409).
 - Cache the fully qualified domain name during environment detection to fix a performance issue on macOS (#339, #394).
 
 Removed
--- a/changelog.txt
+++ b/changelog.txt
@@ -26,6 +26,7 @@ Changed
 Fixed
 +++++
 
+- Ensure that directives are always evaluated before running or submitting (#408).
 - Cache the fully qualified domain name during environment detection to fix a performance issue on macOS (#339, #394).
 
 Removed
--- a/flow/project.py
+++ b/flow/project.py
@@ -271,7 +271,7 @@ class _JobOperation:
         A :class:`flow.directives._Directives` object of additional parameters
         that provide instructions on how to execute this operation, e.g.,
         specifically required resources. This is expected to be evaluated (i.e.
-        all callable directives should be evalauted to concrete value).
+        all callable directives should evaluate to concrete values).
     :type directives:
         :class:`flow.directives._Directives`
     """
--- a/flow/project.py
+++ b/flow/project.py
@@ -270,7 +270,8 @@ class _JobOperation:
     :param directives:
         A :class:`flow.directives._Directives` object of additional parameters
         that provide instructions on how to execute this operation, e.g.,
-        specifically required resources.
+        specifically required resources. This is expected to be evaluated (i.e.
+        all callable directives should be evalauted to concrete value).
     :type directives:
         :class:`flow.directives._Directives`
     """
@@ -282,8 +283,6 @@ class _JobOperation:
         if not (callable(cmd) or isinstance(cmd, str)):
             raise ValueError("JobOperation cmd must be a callable or string.")
         self._cmd = cmd
-
-        directives.evaluate(jobs)
 
         # Keys which were explicitly set by the user, but are not evaluated by the
         # template engine are cause for concern and might hint at a bug in the template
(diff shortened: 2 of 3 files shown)


bdice (changes requested):
@b-butler It looks like this might fix the problem, but I am worried about the duplication between `_JobOperation` and `JobOperation`. I asked a while back about whether we could safely merge the two classes (so that the deprecated public class would call the private class's constructor via `super().__init__(...)` instead of having repeated logic. I believe that may be important, since I don't see a similar `directives.evaluate(jobs)` call in the public class's constructor. Would you be able to take a look into consolidating that?

b-butler:
@bdice Can do.

bdice (changes requested):
@b-butler These changes look good. One more request; would it be possible to add a test that catches the problem in #408?

b-butler:
@bdice, the code coverage tests are failing due to the change in `JobOperation`'s constructor. I don't think like adding tests for specifically for a deprecated class that no one is likely using necessary, but if you feel it is, I can add tests specifically for that so the code coverage tests pass.

bdice:
> @bdice, the code coverage tests are failing due to the change in `JobOperation`'s constructor. I don't think like adding tests for specifically for a deprecated class that no one is likely using necessary, but if you feel it is, I can add tests specifically for that so the code coverage tests pass.

@b-butler I wasn’t worried about code coverage as a number - I was concerned that the issue #408 might not be covered by tests. This fix is good but only if we can prevent the problem from happening again. In my understanding that’s independent of whether the public/deprecated class or private class is used. Is that right?

changelog.txt
```diff
@@ -26,6 +26,7 @@ Changed
 Fixed
 +++++
 
+- Ensure that directives are always evaluated before running or submitting (#408).
```
bdice, line 29:
```suggestion
- Ensure that directives are always evaluated before running or submitting (#408, #409).
```

flow/project.py
```diff
@@ -270,12 +270,13 @@ class _JobOperation:
     :param directives:
         A :class:`flow.directives._Directives` object of additional parameters
         that provide instructions on how to execute this operation, e.g.,
-        specifically required resources.
+        specifically required resources. This is expected to be evaluated (i.e.
+        all callable directives should be evalauted to concrete value).
```
bdice, line 274:
```suggestion
        all callable directives should evaluate to concrete values).
```

tests/test_project.py
```diff
@@ -476,6 +485,64 @@ def a(job):
...
+        )
+        def b(job):
+            return "world"
+
+        project = self.mock_project(A)
+
+        # test for proper evaluation of all directives.
+        job = next(iter(project))
+        for submit_job_operation in project._get_submission_operations(
+            [job], project._get_default_directives()
```
bdice, line 538: The internals work differently on `next`, so we'll need to adjust this test when merging it with `next`. This PR currently targets `master` so this shouldn't be applied yet.
```suggestion
            [(job,)], project._get_default_directives()
```

bdice (approved):
I have a couple small suggestions that I will apply directly. This PR looks good, thanks for the fix!

PR #410, opened 2020-12-30
Optimization: use cached status everywhere.

### LINKED ISSUE #368: Remove FlowGroup._get_status
As a possible simplification of the code base, we could remove `FlowGroup._get_status`. Currently only `FlowProject._eligible_for_submission` relies on the method. We could leverage `FlowProject._fetch_scheduler_status` instead to remove the need for `FlowGroup._get_status` to check its own status.

bdice:
## Description
This PR ensures that all status-fetching operations use a cached status dictionary.

The cached status information was already being used when printing status info, but not for submission.

I refactored the existing code to reduce duplication. I changed the APIs of several internal methods to *require* a `cached_status` argument, instead of allowing None.

## Motivation and Context
The benchmarks I was running locally kept fetching `project.document["_status"]` from the disk and converting from a synced dict to a plain dict. This loading and recursive conversion to a plain dict is slow for large workspaces and only needs to be called once at the beginning.

Also resolves #368.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:
- [x] My code follows the [code style guideline](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md%23code-style) of this project.

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -4663,7 +4663,7 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
         return {}
 
     def _eligible_for_submission(self, flow_group, jobs, cached_status):
-        """Check group eligibility for submission with a job/aggregate.
+        """Check group eligibility for submission with a job or aggregate.
 
         By default, an operation is eligible for submission when it
         is not considered active, that means already queued or running.
--- a/flow/project.py
+++ b/flow/project.py
@@ -1192,12 +1192,10 @@ class FlowGroup:
         # By appending the unique job_op_id, we ensure that each id is truly unique.
         return readable_name + job_op_id
 
-    def _get_status(self, jobs):
+    def _get_status(self, jobs, cached_status):
         """For a given job-aggregate, check the group's submission status."""
         try:
-            return JobStatus(
-                jobs[0]._project.document["_status"][self._generate_id(jobs)]
-            )
+            return JobStatus(cached_status[self._generate_id(jobs)])
         except KeyError:
             return JobStatus.unknown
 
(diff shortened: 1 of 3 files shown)


tests/test_project.py
```diff
@@ -1288,30 +1292,49 @@ def test_script(self):
...
     def test_directives_hierarchy(self):
         project = self.mock_project()
+        cached_status = project._get_cached_status()
         for job in project:
             # Test submit JobOperations
-            job_ops = project._get_submission_operations(
-                (job,), project._get_default_directives(), names=["group2"]
+            job_ops = list(
+                project._get_submission_operations(
+                    aggregates=[(job,)],
```
bdice, line 1300: This change was from a bug in the tests. When `_get_submission_operations` is called, it requires a _sequence of aggregates_, not a _single aggregate_. We validate this in the public method that calls `_get_submission_operations`, but this test calls an internal private method. As a result, this test was being fed invalid data and was passing for the wrong reason. Somewhere (perhaps `_is_selected_aggregate`), a single job was being compared to an aggregate containing only that job, and that made the test pass because no job operations were generated. I only found this problem because of a mistake in the `Job.__eq__` method that was introduced in [PR 442](https://github.com/glotzerlab/signac/pull/442) is being fixed in [PR 455](https://github.com/glotzerlab/signac/pull/455). I have expanded the test to ensure that non-zero (submission/run) job operations are created and fixed the problem.

flow/project.py
```diff
@@ -4612,20 +4662,44 @@ def _get_aggregate_store(self, group):
                 return aggregate_store
         return {}
 
-    def _eligible_for_submission(self, flow_group, jobs):
-        """Check flow_group eligibility for submission with a job-aggregate.
+    def _eligible_for_submission(self, flow_group, jobs, cached_status):
+        """Check group eligibility for submission with a job/aggregate.
```
bdice, line 4666:
```suggestion
        """Check group eligibility for submission with a job or aggregate.
```

flow/project.py
```diff
@@ -2042,6 +2033,33 @@ def scheduler_jobs(self, scheduler):
         """
         yield from self._expand_bundled_jobs(scheduler.jobs())
 
+    def _get_cached_status(self, cached_status=None):
```
b-butler, line 2036: The only place where the `cached_status` argument is currently passed is in `get_job_status`. I would be inclined to just have an `if cached_status is None` check in that function rather than optionally pass a `cached_status` argument solely for `get_job_status`.

b-butler (changes requested):
I only have one comment to address/comment on. The PR looks good and it will be nice to reduce file IO some more.

flow/project.py
bdice: That's a good idea. I was hoping to do that but I didn't recognize how easy it would be to simplify. :smile:

bdice:
@b-butler Thanks for the review. I simplified the method as you suggested. Feel free to merge this once tests pass, if you're happy with it.

b-butler (approved):
LGTM

PR #415, opened 2021-01-03
Unify group and aggregate storage with a bi-map data structure.

bdice:
## Description
Introduces a bidirectional map data structure that allows for a unification of group / aggregate registration and storage mechanisms in the `FlowProject`.

## Motivation and Context
Using a `bidict` data structure reduces redundancy. It eliminates superfluous private methods and makes all accesses O(1) whether by group or by aggregator.

This is an intermediate step for unifying and optimizing the status fetching logic with run/submit logic.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/aggregates.py
+++ b/flow/aggregates.py
@@ -472,15 +472,15 @@ class _DefaultAggregateStore(Mapping):
 
     def __init__(self, project):
         self._project = project
-        # Below, we store repr(project), which defines the hash of this class.
-        # This class must be hashable because it is used as a dict key.
-        # However, when unpickling a FlowProject, this object's hash must be
-        # computed *before* the FlowProject is fully initialized. Thus, it is
-        # not possible to execute repr(project) when hashing the instance at
-        # the time of unpickling. This means that this class cannot be
-        # unpickled unless we pre-emptively compute and store the repr used to
-        # distinguish this instance's hash.
-        self._hash_repr = hash(repr(project))
+        # Below, we store repr(project), which defines the hash and equality
+        # operators of this class. This class must be hashable because it is
+        # used as a dict key. However, when unpickling a FlowProject, this
+        # object's hash must be computed *before* the FlowProject is fully
+        # initialized. Thus, it is not possible to execute repr(project) when
+        # hashing the instance at the time of unpickling. This means that this
+        # class cannot be unpickled unless we pre-emptively compute and store
+        # the repr.
+        self._project_repr = repr(project)
 
     def __iter__(self):
         yield from self.keys()
@@ -521,10 +521,10 @@ class _DefaultAggregateStore(Mapping):
         return len(self._project)
 
     def __eq__(self, other):
-        return type(self) == type(other) and self._project == other._project
+        return type(self) == type(other) and self._project_repr == other._project_repr
 
     def __hash__(self):
-        return self._hash_repr
+        return hash(self._project_repr)
 
     def keys(self):
         for job in self._project:
--- a/flow/project.py
+++ b/flow/project.py
@@ -61,6 +61,7 @@ from .util.misc import (
     _positive_int,
     _to_hashable,
     add_cwd_to_environment_pythonpath,
+    bidict,
     roundrobin,
     switch_to_directory,
 )
@@ -1629,14 +1630,10 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
         self._operations = {}
         self._register_operations()
 
-        # Register all groups with this project instance.
+        # Register all groups and aggregates with this project instance.
         self._groups = {}
-        self._aggregator_per_group = {}
+        self._group_to_aggregator = bidict()
         self._register_groups()
-
-        # Register all aggregates which are created for this project
-        self._stored_aggregates = {}
-        self._register_aggregates()
 
     def _setup_template_environment(self):
         """Set up the jinja2 template environment.
@@ -2082,7 +2079,7 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
 
         groups = [self._groups[name] for name in self.operations]
         for group in groups:
-            if get_aggregate_id(jobs) in self._get_aggregate_store(group.name):
+            if get_aggregate_id(jobs) in self._group_to_aggregator[group]:
                 completed = group._complete(jobs)
                 eligible = not completed and group._eligible(jobs)
                 scheduler_status = cached_status.get(
@@ -2186,7 +2183,7 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
                 total=len(self._groups),
                 file=file,
             ):
-                aggregate_store = self._get_aggregate_store(group.name)
+                aggregate_store = self._group_to_aggregator[group]
                 for aggregate in tqdm(
                     aggregate_store.values(),
                     total=len(aggregate_store),
@@ -2235,7 +2232,7 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
         group = self._groups[group_name]
         status_dict = {}
         errors = {}
-        aggregate_store = self._get_aggregate_store(group.name)
+        aggregate_store = self._group_to_aggregator[group]
         for aggregate_id, aggregate in tqdm(
             aggregate_store.items(),
             desc=f"Collecting aggregate status info for operation {group.name}",
@@ -3314,9 +3311,8 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
                 with self._potentially_buffered():
                     operations = []
                     for flow_group in flow_groups:
-                        for aggregate in self._get_aggregate_store(
-                            flow_group.name
-                        ).values():
+                        aggregate_store = self._group_to_aggregator[flow_group]
+                        for aggregate in aggregate_store.values():
                             operations.extend(
                                 flow_group._create_run_job_operations(
                                     self._entrypoint,
@@ -3461,20 +3457,32 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
             been collected appropriately from its contained operations.
 
         """
-        for group in self._gather_flow_groups(names):
-            for aggregate in self._get_aggregate_store(group.name).values():
-                if (
-                    group._eligible(aggregate, ignore_conditions)
-                    and self._eligible_for_submission(group, aggregate, cached_status)
-                    and self._is_selected_aggregate(aggregate, aggregates)
-                ):
-                    yield group._create_submission_job_operation(
-                        entrypoint=self._entrypoint,
-                        default_directives=default_directives,
-                        jobs=aggregate,
-                        index=0,
-                        ignore_conditions_on_execution=ignore_conditions_on_execution,
-                    )
+        submission_groups = set(self._gather_flow_groups(names))
+        for (
+            aggregate_store,
+            aggregate_groups,
+        ) in self._group_to_aggregator.inverse.items():
+            matching_groups = set(aggregate_groups) & submission_groups
+            if not matching_groups:
+                # Skip iteration over aggregate store if no groups match the
+                # specified names
+                continue
+            for aggregate in aggregate_store.values():
+                for group in matching_groups:
+                    if (
+                        group._eligible(aggregate, ignore_conditions)
+                        and self._eligible_for_submission(
+                            group, aggregate, cached_status
+                        )
+                        and self._is_selected_aggregate(aggregate, aggregates)
+                    ):
+                        yield group._create_submission_job_operation(
+                            entrypoint=self._entrypoint,
+                            default_directives=default_directives,
+                            jobs=aggregate,
+                            index=0,
+                            ignore_conditions_on_execution=ignore_conditions_on_execution,
+                        )
 
     def _get_pending_operations(
         self, jobs=None, operation_names=None, ignore_conditions=IgnoreConditions.NONE
@@ -3501,7 +3509,7 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
         """Verify that the aggregate belongs to this project."""
         return any(
             get_aggregate_id(aggregate) in aggregates
-            for aggregates in self._stored_aggregates
+            for aggregates in self._group_to_aggregator.inverse
         )
 
     @staticmethod
@@ -3512,7 +3520,7 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
     def _get_aggregate_from_id(self, id):
         # Iterate over all the instances of stored aggregates and search for the
         # aggregate in those instances.
-        for aggregate_store in self._stored_aggregates:
+        for aggregate_store in self._group_to_aggregator.inverse:
             try:
                 # Assume the id exists and skip the __contains__ check for
                 # performance. If the id doesn't exist in this aggregate_store,
@@ -4376,7 +4384,8 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
             jobs = _AggregatesCursor(self)
         for name in self.operations:
             group = self._groups[name]
-            for aggregate in self._get_aggregate_store(group.name).values():
+            aggregate_store = self._group_to_aggregator[group]
+            for aggregate in aggregate_store.values():
                 if not self._is_selected_aggregate(aggregate, jobs):
                     continue
                 yield from group._create_run_job_operations(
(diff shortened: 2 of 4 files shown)


flow/aggregates.py
```diff
@@ -521,10 +521,10 @@ def __len__(self):
         return len(self._project)
 
     def __eq__(self, other):
-        return type(self) == type(other) and self._project == other._project
+        return type(self) == type(other) and self._project_repr == other._project_repr
```
atravitz, line 524: Is it possible for the `repr`s to be equivalent but not the `type`s?

flow/project.py
```diff
@@ -2082,7 +2079,7 @@ def _get_operations_status(self, jobs, cached_status):
 
         groups = [self._groups[name] for name in self.operations]
         for group in groups:
-            if get_aggregate_id(jobs) in self._get_aggregate_store(group.name):
+            if get_aggregate_id(jobs) in self._group_to_aggregator[group]:
```
atravitz, line 2082: this and line 2186 are where we get the performance boost, yes?

flow/project.py
```diff
@@ -3313,18 +3308,26 @@ def select(operation):
                 # Change groups to available run _JobOperation(s)
```
atravitz, line 3308: What does this comment mean?

flow/project.py
```diff
@@ -3512,7 +3527,7 @@ def _is_selected_aggregate(aggregate, jobs):
     def _get_aggregate_from_id(self, id):
         # Iterate over all the instances of stored aggregates and search for the
         # aggregate in those instances.
```
atravitz, line 3529: Is this just iterating over groups of aggregate ids? is 
```suggestion
        # Iterate over all the instances of stored aggregate ids and search for the
        # aggregate id in those instances.
```
more accurate, or am I misunderstanding?

atravitz (commented):
This seems much more intuitive and efficient! I just have a few things I'd like to understand

flow/aggregates.py
bdice: If we dropped the type check, we would get an `AttributeError` (can't access `other._project_repr`) when comparing with another type, instead of the expected `False`.

However, the most correct way to implement `__eq__` (accounting for subclasses) is probably closer to what I implemented in https://github.com/glotzerlab/signac/pull/455. I'll adopt that here.

```suggestion
        if not isinstance(other, type(self)):
            return NotImplemented
        return self._project_repr == other._project_repr
```

flow/project.py
bdice: Good question. I'm refactoring this code in a separate PR (not done yet). I will improve the code comments there.

flow/project.py
bdice: Your comment suggestion is partially correct. The aggregate stores are mappings from `aggregate_id` to `aggregate`. I suggest we change the second word (we search for an aggregate _id_) but not the first (the instances are "stored aggregates", not just ids).
```suggestion
        # Iterate over all the instances of stored aggregates and search for the
        # aggregate id in those instances.
```

bdice:
There are some template tests that are failing due to (nondeterministic?) re-ordering of operations that run in parallel. From what I can tell, we don't enforce a sorting order for the output script when operations are launched in parallel. I think this choice of ordering is arbitrary (it shouldn't matter for execution) but we may want to enforce a sorting order (e.g. sorted by the `_JobOperation` id) to prevent template test failures.

kidrahahjo:
@bdice this is fantastic. I loved the way it is implemented.

One question, the reason to store values like before was due to memory issues that can come into play. Now we're storing the mapping of the group to its aggregate store (`{group1: default_store. group2: default_store}`) as well as the mapping of an aggregate store to all the groups it's associated to (`{default_store: [group1, group2]}`).

Wouldn't this be a problem?

kidrahahjo:
> There are some template tests that are failing due to (nondeterministic?) re-ordering of operations that run in parallel. From what I can tell, we don't enforce a sorting order for the output script when operations are launched in parallel. I think this choice of ordering is arbitrary (it shouldn't matter for execution) but we may want to enforce a sorting order (e.g. sorted by the `_JobOperation` id) to prevent template test failures.

I agree with @bdice. While analysing the test failure output, I saw that when bundling, the order of operations is not the same as intended. A quick fix would be something like this:
```
for bundle in _make_bundles(operations, bundle_size):
  def key_func(op):
      return op._id
  bundle = sorted(bundle, key=key_func)
```

But I am not sure if we should go in this direction but it's a small start.

flow/util/misc.py
```diff
@@ -251,3 +251,46 @@ def _cached_partial(func, *args, maxsize=None, **kwargs):
...
+    def __setitem__(self, key, value):
+        """Assign a value to the provided key."""
+        if key in self:
+            self.inverse[self[key]].remove(key)
+        super().__setitem__(key, value)
+        self.inverse.setdefault(value, []).append(key)
+
+    def __delitem__(self, key):
+        """Delete the provided key."""
+        self.inverse.setdefault(self[key], []).remove(key)
```
b-butler, line 293: Do we need `setdefault` here? If we set `self.inverse[self[key]] = []`, then `remove` will always just raise an `ValueError`. `self.inverse[self[key]]` should exist or the class would be in an invalid state.

flow/util/misc.py
```diff
@@ -251,3 +251,46 @@ def _cached_partial(func, *args, maxsize=None, **kwargs):
...
+
+class bidict(dict):
+    r"""A bidirectional dictionary.
+
+    The attribute ``inverse`` contains the inverse mapping, where the inverse
+    values are :class:`set`\ s of keys with that value.
+
+    Both keys and values must be hashable.
+    A key is associated with one or more values.
+    A value is associated with exactly one key.
```
b-butler, line 264: I believe this is backwards. One value (e.g. aggregate store) can be associated with multiple keys (e.g. groups).

flow/project.py
```diff
@@ -4593,20 +4602,26 @@ def foo(job):
...
+        # associated with each group.
         for entry in group_entries:
-            self._groups[entry.name] = FlowGroup(entry.name, options=entry.options)
-            aggregators[entry.aggregator].append(entry.name)
-        self._aggregator_per_group = dict(aggregators)
+            group = FlowGroup(entry.name, options=entry.options)
+            self._groups[entry.name] = group
+            self._group_to_aggregator[group] = entry.aggregator._create_AggregatesStore(
+                self
+            )
```
b-butler, line 4624: Here we aren't checking any more that any repeated aggregate stores aren't created. We should check first to see if the entry's aggregate already exists in the bimap.

flow/project.py
```diff
@@ -3461,20 +3464,32 @@ def _get_submission_operations(
...
+                # specified names
+                continue
+            for aggregate in aggregate_store.values():
+                for group in matching_groups:
+                    if (
+                        group._eligible(aggregate, ignore_conditions)
+                        and self._eligible_for_submission(
+                            group, aggregate, cached_status
+                        )
+                        and self._is_selected_aggregate(aggregate, aggregates)
```
b-butler, line 3484: This check should be after the `for` statement. We shouldn't check this for every group since this does not depend on the group.

b-butler (changes requested):
I like the use of the bidirectional mapping.

flow/util/misc.py
bdice: This might be true. I got this implementation from a StackOverflow post and cleaned up a few pieces of it, but I don't have any tests for the class. I think your statement seems reasonable and changing that worked in my small amount of local testing. I also looked at the StackOverflow post comments but nobody had mentioned that piece of the code. Fixed in adf5370e56d88cce85e1d0af6b6c9d5ecfa94d9d.

flow/util/misc.py
bdice: Good catch. I wrote this docstring 3 times to fix other mistakes I'd made. 😆

flow/project.py
bdice: Good observation. Thanks to @b-butler and @kidrahahjo for pointing this out. Fixed in c0be467f615f5559fa88eb1d1761ee498ffa0198.

flow/project.py
bdice: Good catch -- this is fixed by unifying the selection logic in #416, so I'll mark it as resolved here.

flow/project.py
bdice: Fixed in #416.

flow/project.py
bdice: That is a small speedup, yes. The bigger performance boost is from replacing code like:
```python
for group in groups:
    for aggregate in self._get_aggregate_store(group.name).values()
        ...
```
with the selection logic that is partially implemented in this PR and more thoroughly implemented in #416.

flow/util/misc.py
```diff
@@ -251,3 +251,47 @@ def _cached_partial(func, *args, maxsize=None, **kwargs):
...
+            self.inverse[self[key]].remove(key)
+        super().__setitem__(key, value)
+        self.inverse.setdefault(value, []).append(key)
+
+    def __delitem__(self, key):
+        """Delete the provided key."""
+        value = self[key]
+        self.inverse[value].remove(key)
+        if value in self.inverse and not self.inverse[value]:
+            del self.inverse[value]
```
vyasr, line 296: You don't need the first part of this clause, if `value` is not in `self.inverse` then the previous line would result in a `KeyError`.

flow/util/misc.py
vyasr: I agree with Brandon's analysis. `setdefault` _is_ required in `__setitem__. In the constructor, you might see minor performance improvements if you avoid the setdefault call (I'll make a suggestion there).

flow/util/misc.py
bdice: Fixed.

vyasr:
I'm fine with the contents of this PR and with how my comment was resolved, but @atravitz, @b-butler, and @kidrahahjo have given this a more thorough review than I have so I'm going to remove myself as a reviewer and defer to their evaluations rather than approving.

bdice:
I fixed the template tests in #416. Reviewers, please ignore the failing tests on this PR.

flow/util/misc.py
```diff
@@ -251,3 +251,50 @@ def _cached_partial(func, *args, maxsize=None, **kwargs):
 
     """
     return lru_cache(maxsize=maxsize)(partial(func, *args, **kwargs))
+
+
+class _bidict(dict):
```
kidrahahjo, line 256: Is this private because we don't want users to play around with this class?

tests/test_util.py
```diff
@@ -0,0 +1,20 @@
+from flow.util.misc import bidict
```
kidrahahjo, line 1: Assuming this is fixed in #416 
```suggestion
from flow.util.misc import _bidict
```

tests/test_util.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class TestBidict:
+    def test_bidict(self):
+        # The bidict class inherits from dict, so we trust its inherited
+        # methods and only need to validate its overrides and inverse behavior.
+        bd = bidict({"a": 1, "b": 2})
+        bd["c"] = 1
+        assert bd == {"a": 1, "c": 1, "b": 2}
+        assert bd.inverse == {1: ["a", "c"], 2: ["b"]}
```
kidrahahjo, line 11: I am not sure about this, but will the order always be conserved?

flow/project.py
```diff
@@ -3313,18 +3308,26 @@ def select(operation):
...
+                    for (
+                        aggregate_store,
+                        aggregate_groups,
+                    ) in self._group_to_aggregator.inverse.items():
+                        matching_groups = set(aggregate_groups) & run_groups
+                        if not matching_groups:
+                            # Skip iteration over aggregate store if no groups match the
+                            # specified names
+                            continue
+                        for aggregate in aggregate_store.values():
```
kidrahahjo, line 3321:
```suggestion
                        for aggregate in aggregate_store.values():
                            if aggregate not in aggregates:
                                continue
```

This is a really small optimization which makes a difference when only a few aggregates/jobs are provided by the user.

Given this, we can now remove the exact same check in the `select` method above.

flow/project.py
kidrahahjo, line 3484: This is not group-specific so we can consider checking this in the outer loop and continue if it's not the selected aggregate.

Also, I feel like the `_is_selected_aggregate` method is now redundant because we simply just check the availability of `aggregate` in `aggregates`.

flow/project.py
```diff
@@ -3313,18 +3308,26 @@ def select(operation):
...
-                                    self._entrypoint,
-                                    default_directives,
-                                    aggregate,
-                                    ignore_conditions,
+                    run_groups = set(self._gather_flow_groups(names))
+                    for (
+                        aggregate_store,
+                        aggregate_groups,
+                    ) in self._group_to_aggregator.inverse.items():
+                        matching_groups = set(aggregate_groups) & run_groups
```
kidrahahjo, line 3316: Do we have tests to check that the groups don't get executed unless the user specifically asked for them?

kidrahahjo (commented):
I have some minor suggestions, apart from them, the PR is looking good.

flow/util/misc.py
b-butler, line 256: @bdice Also do note that `update` will not call `__setitem__`, `pop` will not call `__delitem__`, `setdefault` will not call `__setitem__` and so on. The CPython `dict` class is optimized to not use these paradigms that exist in `abc.collections`. We should probably either disable these methods `return NotImplementedError` or implement all methods that involve setting or deleting. I forgot this when reviewing last night.

b-butler (changes requested):
I left a comment about the implementation of `_bidict`. Also I would like to test some of the methods mentioned in the comment in your `_bidict` test. Beyond that I am ready to merge.

flow/util/misc.py
bdice: @b-butler That's a really helpful observation, I didn't realize that. I re-implemented the class as a `MutableMapping` with an internal dictionary, so it gets all of those methods "for free." Also, I added tests for all of the methods expected from a `MutableMapping`. Fixed in 0789f2a.

bdice: Yes, I don't want to make this part of the public API. If it's public, it's harder to make breaking changes or remove it if it's not being used anymore.

tests/test_util.py
bdice: Yes, this is fixed in #416. My mistake.

tests/test_util.py
bdice: The inverse values should always be insertion-ordered. I considered using a `set` instead of a `list` to store the inverse mapping's values (which are keys in the forward mapping), but it turns out that we need the insertion-ordering property so that aggregate stores return groups in the order that they are registered. This is necessary for ensuring that groups/operations are always ordered in a deterministic way. I will add a comment explaining this requirement.

flow/project.py
bdice: I will look at this in #416, I refactored this code and may have already made the change you're suggesting.

flow/project.py
bdice: This is handled in #416.

flow/project.py
bdice: I'm not entirely sure but I _think_ we have coverage of that. I'll defer this to a follow-up PR.

bdice:
I'm going to go ahead and merge this into `next`. I will handle the remaining suggestions/comments in #416. The two PRs are very closely related and I think #416 will fix most of the problems that we have identified in this PR. Thank you so much to all reviewers for the thoughtful input!

flow/project.py
bdice: I think the new approach in #416 is similar to what you were looking for here. Please comment on that PR in the method `_generate_selected_aggregate_groups` if not. Here's the logic:
```python
if selected_aggregates is not None:
    # Use selected aggregates in the aggregate store
    for aggregate in aggregate_progress_wrapper(selected_aggregates):
        aggregate_id = get_aggregate_id(aggregate)
        if aggregate_id in aggregate_store:
            for group in matching_groups:
                yield aggregate_id, aggregate, group
else:
    # Use all aggregates in the aggregate store
    for aggregate_id, aggregate in aggregate_progress_wrapper(
        aggregate_store.items()
    ):
        for group in matching_groups:
            yield aggregate_id, aggregate, group
```

flow/project.py
bdice: I confirmed that tests fail on #416 if this check is not performed correctly (if more groups are run/status-fetched/submitted than were requested).

Issue #404, opened 2020-12-17
Add type hints


csadorf:
The source of signac-flow would likely be more readable and more bug-resilient with [type hints](https://docs.python.org/3/library/typing.html).

atravitz:
I strongly agree with this - should it be done on `next`?

bdice:
I would suggest using the approach from this PR to automatically generate some type hints, and then manually refine them. https://github.com/glotzerlab/signac/pull/313

kidrahahjo:
I'll work on implementing this!

vyasr:
@kidrahahjo while this feature would be nice, I would recommend working on something else for the moment if you have the time. There are higher priority items to be worked on at the moment, in particular I would focus on things related to getting actual aggregation working. Working through #362 and updating that checklist with remaining items would be very helpful.

If you want to wait on that for a bit while issues with the current code base get ironed out after the 0.12 release, then I would suggest trying to work through some of the more self-contained issues in our backlog. Considering the amount of code in flow and the fact that this kind of change would require eventually modifying every file (although of course you can stage it to avoid introducing lots of merge conflicts), I would like to get our issue list shorter before working on something bigger like this.

For starters, I'd suggest the following issues:
- #258: I think this might already be handled by directives.py and you'll just need to do a thorough review of existing environments/templates to make sure there's no legacy code that's not using this path.
- #370: Also hopefully mostly addressed by @bdice but there's probably some holdovers.
- #341: This _might_ be a bigger project, but worth revisiting to see if #417 makes this easier. It's at least worth scoping out and documenting next steps.
- #270: This feature is a blocker for real applications, and should be handled.

kidrahahjo:
@vyasr I agree on what you suggested. I had a chat with @bdice on slack, I will start working on introducing actual aggregation as soon as we release flow `0.12.0`. With respect to the point that this should be handled after we're done with some of the already existing issues, I think you're definitely right. I will have a look at the mentioned issues and see what I can do.

PR #416, opened 2021-01-05
Unify group/aggregate selection

bdice:
## Description
Reduces redundancy in several parts of the code that select subsets of the project's groups/aggregates. Also updates some docstrings/etc. These changes will support the rewrite of the status fetching code that I am working on next. This PR lays some initial groundwork to allow "aggregate status fetching" but that's not yet complete and may be changed in the following PR.

I removed the "by-op" run ordering in this PR, which has not been released and was only created to help with the implementation of aggregation.

In my testing FlowProject with 3 operations and 1000 jobs, I saw the runtime for `submit` and `run` reduced by about 1/3. This is because the new iteration order in this PR saves quite a bit of extra work. For jobs with more than 3 operations, I believe this might save even more work.

Per our discussion in dev meeting, I am deferring on adding tests with non-trivial aggregators for now.

## Motivation and Context
Unifying code paths, improving status fetching, better performance.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


flow/project.py
```diff
@@ -1265,18 +1254,18 @@ def _get_run_ops(ignore_ops, additional_ignores_flag):
...
+                    entrypoint=entrypoint,
+                    default_directives=default_directives,
+                    jobs=jobs,
+                    ignore_conditions=ignore_conditions_on_execution
+                    | additional_ignores_flag,
                 )
-                - set(ignore_ops)
-            )
+                if op not in ignore_ops
+            ]
```
bdice, line 1268: This is a crucial change - the previous `set(...) - set(...)` approach does not order the results deterministically, because the `set` class in Python is unordered. This led to inconsistency between runs with template tests. We must use a `list` to iterate over the created run-job-operations, but it's perfectly fine to use a `set` for the ignored operations because it's O(1) to check `if op not in ignore_ops` if `ignore_ops` is a `set`.

flow/project.py
```diff
@@ -2054,13 +2043,103 @@ def _get_cached_status(self):
         except KeyError:
             return {}
 
-    def _get_operations_status(self, jobs, cached_status):
-        """Return a dict with information about job-operations for this aggregate.
+    def _generate_selected_aggregate_groups(
```
bdice, line 2046: This is the most important part of this PR. We need this method to have a fast way to execute the iteration for all the cases below:
- `selected_aggregates` is `None`, a list of aggregates, an `_AggregateCursor`, or any other common way to express an iterable of aggregates
- `selected_groups` is `None` or a set of groups (e.g. groups representing singleton operations)

flow/project.py
```diff
@@ -2077,75 +2156,99 @@ def _get_operations_status(self, jobs, cached_status):
...
-                "Error while determining labels for job '%s': '%s'.", job, error
-            )
-            if ignore_errors:
-                result["labels"] = []
-                result["_labels_error"] = str(error)
-            else:
-                raise
+        return result
+
+    def get_job_status(self, job, ignore_errors=False, cached_status=None):
```
bdice, line 2222: This function was heavily simplified by creating `_get_aggregate_status` above and using the duplicated code from `_get_job_labels`.

flow/project.py
```diff
@@ -3345,25 +3440,23 @@ def key_func_by_job(operation):
             # Optionally re-order operations for execution if order argument is provided:
             if callable(order):
                 operations = list(sorted(operations, key=order))
-            elif order == "random":
-                random.shuffle(operations)
-            elif order in ("by-job", "cyclic"):
+            elif order == "cyclic":
```
bdice, line 3443: This ordering code has been mostly reverted to its pre-aggregation form. by-op was removed, and the rest should look like what is on `master`.

flow/project.py
```diff
@@ -3305,29 +3404,25 @@ def select(operation):
                 )
                 break
             try:
-                # Change groups to available run _JobOperation(s)
+                # Generate _JobOperation instances for selected groups and aggregates.
```
bdice, line 3407: @atravitz requested an improvement on this comment earlier. 👍

flow/project.py
```diff
@@ -4877,11 +4955,11 @@ def operation_function(job):
...
             raise KeyError(f"Unknown operation '{args.operation}'.")
 
-        group = self._groups[args.operation]
-        aggregate_store = self._group_to_aggregator[group]
-        for aggregate in aggregate_store.values():
-            if self._is_selected_aggregate(aggregate, aggregates):
-                operation_function(*aggregate)
+        for aggregate_id, aggregate, group in self._generate_selected_aggregate_groups(
+            selected_aggregates=aggregates,
+            selected_groups={self._groups[args.operation]},
```
bdice, line 4955: Selects just a single group -- this should be efficient for the iteration logic in `_generate_selected_aggregate_groups` because it will heavily filter its iterations. It should only iterate over the aggregate store for this group, and only handle the provided group.

tests/test_project.py
```diff
@@ -770,13 +770,17 @@ class TestExecutionProject(TestProjectBase):
 
     def test_pending_operations_order(self):
```
bdice, line 771: This behavior is reverted to the pre-aggregation behavior.

flow/project.py
```diff
@@ -2054,13 +2043,103 @@ def _get_cached_status(self):
...
+            and len(self._group_to_aggregate_store.inverse) > 1
+        ):
+            # Use only aggregate stores for the selected groups.
+            selected_aggregate_stores = {
+                self._group_to_aggregate_store[group] for group in selected_groups
+            }
+            aggregate_stores = {
+                aggregate_store: self._group_to_aggregate_store.inverse[aggregate_store]
+                for aggregate_store in selected_aggregate_stores
+            }
```
atravitz, line 2099: I'm a little confused, maybe because of naming. Is `selected_aggregate_stores` just the set of aggregates from the selected groups? and then `aggregate_stores` is the dict that maps these aggregates to their groups?

flow/project.py
atravitz: much clearer!!

flow/aggregates.py
```diff
@@ -549,23 +549,23 @@ def _register_aggregates(self, project):
...
 
     """
-    if len(jobs) == 1:
-        return jobs[0].get_id()  # Return job id as it's already unique
+    if len(aggregate) == 1:
+        return aggregate[0].get_id()  # Return job id as it's already unique
 
-    blob = ",".join(job.get_id() for job in jobs)
+    blob = ",".join(job.get_id() for job in aggregate)
     hash_ = md5(blob.encode("utf-8")).hexdigest()
```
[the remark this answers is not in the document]
bdice: Yes, exactly. We use this to uniquely identify aggregates from their component jobs. It generates ids that look like `agg-081759177c477536f882eee3e30b398e`.

flow/project.py
bdice: The problem was nondeterminism in this block of code, which the tests showed. We ended up printing out the job operations in the section `# Eligible to run:` and using a difference of (unordered) sets made that output have no guaranteed order. Switching to a list comprehension that excludes items in the `ignore_ops` set makes the order consistent from run to run. We need ordering for test stability and workflow reproducibility.

flow/project.py
bdice: Good question. You're mostly correct about the descriptions of both variables. To clarify the first one, I might say `selected_aggregate_stores` is the set of _aggregate stores_ (not the set of _aggregates_) that correspond to the selected groups. I agree this is a potentially confusing name. I will rename `selected_aggregate_stores` to `aggregate_stores_of_selected_groups` to try and make it clearer.

atravitz: great, thanks! having `x` and `selected_x` implies to me that theyre the same type and one is a subset of the other.

flow/project.py
```diff
@@ -2054,13 +2043,103 @@ def _get_cached_status(self):
...
+            and len(self._group_to_aggregate_store.inverse) > 1
+        ):
+            # Use only aggregate stores for the selected groups.
+            aggregate_stores_of_selected_groups = {
+                self._group_to_aggregate_store[group] for group in selected_groups
+            }
+            aggregate_stores = {
+                aggregate_store: self._group_to_aggregate_store.inverse[aggregate_store]
+                for aggregate_store in aggregate_stores_of_selected_groups
+            }
```
kidrahahjo, line 2099:
```suggestion
            aggregate_stores = {
                aggregate_store: self._group_to_aggregate_store.inverse[aggregate_store]
                for group, aggregate_store in self._group_to_aggregate_store.items() 
                if group in selected_groups
            }
```

Can the code be replaced with this one?

kidrahahjo: Also, `self._group_to_aggregate_store.inverse[aggregate_store]` would map to all the groups (even if they're not selected).

This means that we'll end up accepting all the groups if one of them is selected.

flow/project.py
```diff
@@ -2054,13 +2043,103 @@ def _get_cached_status(self):
...
+        for (
+            aggregate_store,
+            aggregate_groups,
+        ) in aggregate_stores.items():
+            if selected_groups is not None:
+                # Filter out groups that are not selected. The order of
+                # aggregate_groups must be preserved. Using an intersection of
+                # ordered sets would be preferable but would require a
+                # dependency.
+                matching_groups = [
```
kidrahahjo, line 2113: Aaah I see, please ignore my above comment.

kidrahahjo (commented):
I did not review all of it. I stopped after reviewing `get_job_status`. I love the implementation though. It looks very clean now.

Here are some minor comments.

flow/project.py
```diff
@@ -3305,29 +3414,25 @@ def select(operation):
...
-                                        ignore_conditions,
-                                    )
-                                )
+                        aggregate_id,
+                        aggregate,
+                        group,
+                    ) in self._generate_selected_aggregate_groups(
+                        selected_groups=run_groups,
+                    ):
+                        operations.extend(
```
kidrahahjo, line 3428:
```suggestion
                        if aggregate not in aggregates:
                            continue
                        operations.extend(
```

This might save users a little time if they request `run` for a subset of jobs.

flow/project.py
```diff
@@ -3512,22 +3604,11 @@ def _verify_group_compatibility(cls, groups):
...
-    @staticmethod
-    def _is_selected_aggregate(aggregate, jobs):
-        """Verify whether the aggregate is present in the provided jobs."""
-        return aggregate in jobs
-
     def _get_aggregate_from_id(self, id):
         # Iterate over all the instances of stored aggregates and search for the
         # aggregate id in those instances.
-        for aggregate_store in self._group_to_aggregator.inverse:
+        # aggregate in those instances.
```
kidrahahjo, line 3610: These two comments are confusing.

kidrahahjo (commented):
I finished my review round!

flow/project.py
```diff
@@ -2077,75 +2156,104 @@ def _get_operations_status(self, jobs, cached_status):
...
-                    group._generate_id(jobs), JobStatus.unknown
-                )
-                for operation in group.operations:
-                    if scheduler_status >= status_dict[operation]["scheduler_status"]:
-                        status_dict[operation] = {
-                            "scheduler_status": scheduler_status,
-                            "eligible": eligible,
-                            "completed": completed,
-                        }
+        single_operation_groups = {self._groups[name] for name in self.operations}
```
vyasr, line 2159: Maybe add a TODO indicating that we're just not showing any meaningful status for groups at this stage.

flow/project.py
vyasr: This function is good to have, I agree that we need to cover these cases. Although its internal, and while it probably gets decent run from the various uses, I would like to see some direct tests of this method. It's complex enough, and right now if something goes wrong elsewhere we'd have to trace back pretty far to identify failure cases.

flow/project.py
```diff
@@ -2247,7 +2352,12 @@ def _get_group_status(self, group_name, cached_status, ignore_errors=False):
                 job_op_id = group._generate_id(aggregate)
                 scheduler_status = cached_status.get(job_op_id, JobStatus.unknown)
                 completed = group._complete(aggregate)
-                eligible = not completed and group._eligible(aggregate)
+                eligible_ignore_conditions = (
+                    IgnoreConditions.POST if completed else IgnoreConditions.NONE
+                )
+                eligible = not completed and group._eligible(
+                    aggregate, ignore_conditions=eligible_ignore_conditions
+                )
```
vyasr, line 2360: I'm confused about why this ignore conditions logic is necessary. The checks for `eligible` will short-circuit if ` completed is True`, so `group._eligible` is only called when `not completed`, in which case we should pass `IgnoreConditions.None`, which is the default. Am I missing something? If not, I think I saw the exact same logic in one of your new functions (probably the function getting aggregate/group status?).

flow/project.py
```diff
@@ -2287,21 +2397,21 @@ def _get_job_labels(self, job, ignore_errors=False):
             Dictionary with keys ``job_id``, ``labels``, and ``_labels_error``.
 
         """
-        result = {}
-        result["job_id"] = str(job)
+        result = {
+            "job_id": job.get_id(),
```
vyasr, line 2401: I think we want to use `job.id` everywhere instead of `job.get_id`, as per glotzerlab/signac#382.

vyasr: I believe this also affects code in `get_aggregate_id`.

flow/project.py
```diff
@@ -2316,7 +2426,7 @@ def _fetch_status(
 
         Parameters
         ----------
-        aggregates : list
+        aggregates : sequence of aggregates
```
vyasr, line 2429:
```suggestion
        aggregates : Sequence[aggregates]
```

vyasr (changes requested):
Looks like a nice refactor. This definitely helps unify a bit more of the logic

flow/project.py
bdice: @vyasr We currently require signac-flow compatibility back to signac 1.0. The `id` property was introduced in signac 1.3. If we could bump our signac version requirement to 1.3, that would allow us to make this change.

flow/project.py
bdice: I agree. I will add tests.

flow/aggregates.py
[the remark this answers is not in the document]
bdice: I think we're okay with that. It just extends from 32 characters to 36, I think. Not a problem.

flow/project.py
```diff
@@ -384,7 +383,7 @@ def __repr__(self):
         )
 
     def __hash__(self):
-        return int(sha1(self.id.encode("utf-8")).hexdigest(), 16)
+        return hash(self.id)
```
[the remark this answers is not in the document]
bdice: The `__hash__` method here does not require a function like md5 or sha1. It's sufficient to just return `hash(self.id)`. This appeared to be a legacy pattern that was copied for no apparent reason, and it is the only place left in signac or signac-flow where such a custom hash function is provided.

The purpose of Python's `hash` is to help with hash-based data structures like `dict` and `set` and (potentially) to be used as a required condition for object equality. The requirements of md5 / sha1 are stricter and they are much more expensive to compute than the built-in `hash` function.

flow/project.py
bdice: @kidrahahjo I believe that your single comprehension is equivalent. I chose to do this in two comprehensions instead of one because I felt that one comprehension was a bit hard to read. If you think your suggestion is more readable, we can change it. I don't think there are any performance implications for either choice.

For your second question about mapping to all groups: yes, that is true. I see you noticed the comment on matching groups below, it was my intention to filter those groups (in order) in that later step. I don't think filtering at this point would make it any more efficient. We still perform the filtering before we iterate over the selected groups.

flow/project.py
bdice: Good observation - instead of the suggestion you made, I provided the argument `selected_aggregates=aggregates` to the generator. That should end up doing the same thing you suggested, but using the logic in the generator to filter instead.

flow/project.py
bdice: Haha, good catch. I made a mistake in resolving merge conflicts. Fixed.

flow/aggregates.py
```diff
@@ -549,23 +549,23 @@ def _register_aggregates(self, project):
         self._project = project
 
 
-def get_aggregate_id(jobs):
+def get_aggregate_id(aggregate):
```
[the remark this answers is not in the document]
bdice: I'm glad you asked. I would prefer to respect Python's intended ordering for the resolution of `__eq__` which can only be done via `isinstance` checks. I don't think this choice will present a performance issue. The resolution order is described further here: https://docs.python.org/3/reference/datamodel.html#object.__eq__

flow/aggregates.py
```diff
@@ -549,23 +549,23 @@ def _register_aggregates(self, project):
...
     """Generate hashed id for an aggregate of jobs.
 
     Parameters
     ----------
-    jobs : tuple of :class:`~signac.contrib.job.Job`
-        The signac job handles.
+    aggregate : tuple of :class:`~signac.contrib.job.Job`
```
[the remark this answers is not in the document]
bdice: Even if lists / other sequences would work, I would like the documentation to state that this _expects_ a `tuple`. Anecdotally, I haven't seen any instances in the code base where lists / non-tuples are used.

flow/aggregates.py
```diff
@@ -549,23 +549,23 @@ def _register_aggregates(self, project):
...
         The generated aggregate id.
 
     """
-    if len(jobs) == 1:
-        return jobs[0].get_id()  # Return job id as it's already unique
+    if len(aggregate) == 1:
+        return aggregate[0].get_id()  # Return job id as it's already unique
 
-    blob = ",".join(job.get_id() for job in jobs)
+    blob = ",".join(job.get_id() for job in aggregate)
```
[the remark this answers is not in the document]
bdice: That is reasonable. In signac, about half of the uses of that variable name are strings instead of bytes. This is the only instance of `blob` in signac-flow, so I will go ahead and change it here.

flow/aggregates.py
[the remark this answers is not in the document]
vyasr: Right, that's what I mean by "proper" implementations of `__eq__`. If you are confident that this isn't a performance issue in your current benchmarks then that's fine.

flow/project.py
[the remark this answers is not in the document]
vyasr: Sorry, I left this comment without appropriate context. I was primarily responding to @kidrahahjo, and indicating future work.

I agree that using `hash` is more appropriate here. What I was discussing is the use of sha1 in general w.r.t. security as per the original comment, because I noticed that sha1 is still imported by this file but md5 was added. I think that in general we should use md5 everywhere because that's our historical usage. If we want to change to something else (aggregation uses sha1 in a couple places) that should be a discussion and be done globally. In general we don't have security concerns; we're not storing passwords, or any encrypted data, we just want something reasonably collision-free. If we start seeing collisions and need a better algorithm, that's a schema change in signac and may lead to breaking changes between signac-flow releases (because an aggregate's name changes), so I think we should be consistent for now and make uniform changes as needed later.

flow/project.py
bdice: Here's the logic:
- `completed is False` ▶️ _some postcondition is not satisified_
- `completed is False` ▶️ we need to run the eligibility check.
- Without the `ignore_conditions`, the eligibility checks postconditions _again_ (if all preconditions are met).
- Since we already know that _some postcondition is not satisfied_, it is sufficient to determine eligibility only according to preconditions. This eliminates the need to check postconditions twice.

This will provide a benefit only when all preconditions are met (otherwise the eligibility check would exit at the first failed precondition).

flow/project.py
bdice: We're wildly inconsistent with docstring parameter type syntax. I honestly don't know what the right syntax is supposed to be. Would it be something like `Sequence[tuple[signac.contrib.job.Job]]`? The other problem is that it won't link classes in the Sphinx docs unless we use `:class:` syntax...

flow/aggregates.py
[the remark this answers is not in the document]
vyasr: The code definitely used to, but I haven't done a thorough review since the next branch was created and it's possible that we eventually cleaned up all the use case. As long as we've actually standardized that usage I'm fine with it, we definitely agreed that tuples are the _correct_ storage mechanism.

flow/project.py
vyasr: Ah I see, this is purely an optimization. I thought there was an issue with correctness. Please add a comment describing this logic (more briefly :)) in the code.

flow/project.py
vyasr: I agree, don't use `str`. I would also prefer to stick to public APIs, so I think we are stuck with `get_id` until we drop version compatibility.

flow/project.py
vyasr: Yes, I think that's probably appropriate, but I agree that we're not consistent and I don't want to start trying to figure that out as part of this PR. Just disregard this suggestion, and we'll try and come up with something better at a later date, perhaps as part of #404.

numpydoc is kind of inconsistent about it, but in some contexts like parameter lists it does infer the type without the needed for an explicit cross-reference. Once you write something like "sequence of aggregates" all bets are off though, I'm not sure when that'll work and when not.

vyasr:
I think of my open comments there are two that need to be addressed. One is adding comments to the IgnoreConditions optimization, and the other is adding tests of the new function.

flow/project.py
kidrahahjo: Nah, I'm fine with it. I'll resolve this conversation.

bdice:
I will handle the tests suggested by @vyasr in a subsequent PR (WIP).

Issue #401, opened 2020-12-17
Discussion: Performance improvements and aggregation


bdice:
This issue exists to document a series of performance improvements I have worked on recently. We created the `next` branch to be a testing area for the new aggregation features, where we can focus on eliminating any bugs and improving performance before merging into `master` and releasing 0.12.0.

In PR #399, I have created a "benchmark" project with 3 operations and 2 labels. This project does not utilize groups or aggregation, so that we can use it as a fair performance comparison with older versions of signac.

Comparison of runtimes on the test project (2 labels, 3 operations with pre/postconditions):
| flow             | signac                                 | Runtime for 1000 jobs (seconds) |
|------------------|----------------------------------------|---------------------------------|
| 0.8.0            | 1.2.0                                  | 52.5                            |
| 0.9.0            | 1.2.0                                  | 50.5                            |
| 0.10.0           | 1.2.0                                  | 65.0                            |
| 0.11.0           | 1.2.0                                  | 75.6                            |
| 0.11.0           | 1.5.0                                  | 71.7                            |
| 0.11.0           | master (7db7d57)                       | 42.1                            |
| master (8a33a19) | master (7db7d57)                       | 43.1                            |
| next (2328ba1)   | master (7db7d57)                       | 64.2                            |
| next (2328ba1)   | feature/optimize-dfs-convert (43605a9) | 61.4                            |

We're getting the performance back down to a tolerable level (within the bounds of past releases). It was something like 100-110 seconds before I made the recent round of PRs to signac and signac-flow. Measuring the scaling for larger workspaces will also be important before we make any final conclusions.

It looks like the performance improvements in signac have been extremely helpful, and using signac `master` with the last flow release 0.11.0 or the `master` branch of flow gives the overall best performance thus far (and it would be even better with the `_dfs_convert` optimization in [this PR](https://github.com/glotzerlab/signac/pull/429)). The overall trend is that newer signac versions are faster, while newer flow versions are slower (as we would expect, since signac has received more optimizations over time and flow has received more complex features over time).

The primary difference between flow's `master` and `next` branches is the introduction of aggregation. Breaking down the profile results for those two, the difference in performance appears to be solely affecting status check performance (`print_status`). **Running and submitting jobs appear to have nearly identical performance to past versions**, and with similar profiles before/after aggregation.

[...]

PR #417, opened 2021-01-06
New _fetch_status code.

### LINKED ISSUE #368: Remove FlowGroup._get_status

### LINKED ISSUE #413: Test re-ordering causes failures
### Description
Reversing the order of tests causes some tests to fail - I believe this may be due to some problem with how we mock the system executable. Some of the failing tests show that they cannot find `/usr/local/bin/python`.

### To reproduce

On the `next` branch, execute:
```
pip install pytest-reverse
pytest --reverse
```

### Error output

One error output is copied below. There are many others that may be unrelated to the executable mocking process.
```
E               subprocess.CalledProcessError: Command 'export OMP_NUM_THREADS=1;  /usr/local/bin/python /home/bdice/code/signac-flow/tests/define_test_project.py exec op3 5e4d14d82c320bafb2f1286fe486d1f8' returned non-zero exit status 127. 
```

### System configuration

 - Operating System [e.g. macOS]: Ubuntu 20.04
 - Version of Python [e.g. 3.7]: 3.9
 - Version of signac [e.g. 1.0]: 1.5.1
 - Version of signac-flow: `next` (071389304beba386ba010726d551aa8957a9adda)

bdice:
## Description
This is a full re-implementation of the `_fetch_status` method. It cuts out a lot of the complexity of the old code and makes it significantly faster.

The new approach:
- Iterates only once over all aggregates/groups instead of twice
- Uses a context manager to update cached aggregate/group status transparently. For status fetching and submission, only aggregates/groups that are selected will be updated in the cached status in `project.document["_status"]`.
- Uses a tqdm wrapper around concurrent futures. Note: if we choose to use this approach (which is a lot less total code), we will need to bump our tqdm version requirement to `tqdm>=4.42.0`.
- Removes methods `get_status`, `set_status` from `FlowGroup`. These are inefficient and awkward, since they rely on information from the `FlowProject` and don't have a good way to access that information.

## Motivation and Context
The previous status fetching code was quite complex and did more work than is necessary. This code is an (attempt at) improvement. See also: #401

Resolves #368.

I also resolved #413 in this PR. This bug in the tests was a hindrance, I apologize to reviewers for making those changes in this already-large pull request.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


flow/project.py
```diff
@@ -2443,9 +2459,60 @@ def _fetch_status(
...
+
+            status = {}
+            error_text = None
+            try:
+                status["scheduler_status"] = scheduler_status
+                completed = group._complete(aggregate)
+                status["completed"] = completed
+                eligible_ignore_conditions = (
+                    IgnoreConditions.NONE if completed else IgnoreConditions.POST
+                )
```
bdice, line 2482: When we print the status, we have to evaluate all pre/postconditions to determine both "eligible" (all preconditions met and some unmet postcondition) and "complete" (all postconditions met). With run/submit, we can exit early if some precondition is unmet / etc., but that is not the case for status (we must do more work). I found and implemented a partial optimization that will help in certain cases: If `completed=True` then the eligibility check is skipped. However, if `completed=False`, the eligibility check should ignore postconditions (it has already been established that `completed=False`, so checking postconditions is unnecessary). This is implemented in e0abac5.

flow/util/misc.py
```diff
@@ -303,3 +306,46 @@ def __iter__(self):
...
+    Parameters
+    ----------
+    parallelization : str
+        Parallelization mode. Allowed values are "thread", "process", or
+        "none". (Default value = "thread")
+
+    """
+    if parallelization == "thread":
+        parallel_executor = thread_map
+    elif parallelization == "process":
```
bdice, line 323: I'd like help on this if we decide to keep process parallelism. This code just hangs when the parallel processes fork and I can't figure out why.

tests/define_test_project.py
```diff
@@ -79,7 +79,7 @@ class _DynamicTestProject(_TestProject):
 @_DynamicTestProject.operation
 @group3
 @_DynamicTestProject.pre.after(op1)
-@_DynamicTestProject.post.true("dynamic")
+@_DynamicTestProject.post(lambda job: job.sp.get("dynamic", False))
```
bdice, line 82: I think this was a bug in the tests. The previous condition `post.true("dynamic")` was checking the job _document_ but the purpose of the dynamic execution is to change the _state point_. This led to incorrect behavior that was exposed by a mismatch in the value `expected_number_of_steps`, set below. I believe I have the correct number of steps now (see comment below).

tests/test_project.py
```diff
@@ -1128,7 +1128,7 @@ def mock_project(self, project_class=None):
 
 class TestExecutionDynamicProject(TestExecutionProject):
     project_class = _DynamicTestProject
-    expected_number_of_steps = 10
+    expected_number_of_steps = 7
```
bdice, line 1131: I think this was a bug in the tests (see comment above for explanation of the incorrect postcondition). This should take seven steps. In each step, the scheduler makes a pass and then the project runs `submit(...)`.

1. Scheduler: no jobs. Project: submit op1, op2, op3 for all jobs (those three operations are eligible at the beginning).
2. Scheduler: move op1-3 jobs from submitted to holding. Project: no action.
3. Scheduler: move op1-3 jobs from holding to queued. Project: no action.
4. Scheduler: execute op1-3 jobs, set status to inactive. Project: submit op4 for all jobs (eligible because op1 is complete).
5. Scheduler: move op4 jobs from submitted to holding. Project: no action.
6. Scheduler: move op4 jobs from holding to queued. Project: no action.
7. Scheduler: execute op4 jobs, set status to inactive. Project: no action (all done).

flow/util/misc.py
vyasr: I believe I've resolved this error, although at the moment a lot of tests are failing due to the recent changes in mocking. I had to check out [fa6dad6](https://github.com/glotzerlab/signac-flow/pull/417/commits/fa6dad6a94e5aa899cf8677b7d22d4cfe08cfe60) to get enough status tests to run to identify the issue, so you may need to recover to at least that state before you can validate that my fix resolves your problem.

The problem you're running into is that while you were pickling all the _arguments_ to the function, the function itself that you're trying to execute in parallel is not pickleable. This is due in part to the fact that it's a local function, but also to the fact that even if you remove it from the local scope it will be calling the function passed to `_get_parallel_executor`, which is typically going to call other unpickleable things like class methods of `FlowProject`. I resolved this issue by using cloudpickle to serialize the function itself, then binding that to a partial that can be executed.

flow/project.py
```diff
@@ -2036,6 +2019,30 @@ def _get_cached_status(self):
         except KeyError:
             return {}
 
+    @contextlib.contextmanager
+    def _update_cached_scheduler_status(self):
```
b-butler, line 2023: I can't remember, what did we do before about updating the project's status?

b-butler (commented):
The code looks a lot cleaner. I will review again when the tests are fixed.

flow/project.py
bdice: It was done in a somewhat ad-hoc manner by several methods that modified individual keys in `project.document["_status"]`. The previous code could trigger multiple write operations to the project document, when we really only need one write operation at the end.

flow/util/misc.py
bdice: @vyasr Just wanted to offer a massive THANKS for this! I was having a lot of difficulty identifying that problem. Your explanation totally makes sense. Here's a cookie to express my appreciation. 🍪

flow/util/misc.py
```diff
@@ -303,3 +306,64 @@ def __iter__(self):
...
+                # functions like FlowProject.compute_status.
+                partial(_run_cloudpickled_func, cloudpickle.dumps(unpickled_func)),
+                *pickled_iterables,
+                **kwargs,
+            )
+
+    else:
+
+        def parallel_executor(func, *iterables, **kwargs):
+            return list(map(func, *iterables))
```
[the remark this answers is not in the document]
bdice: Fixed.

flow/project.py
```diff
@@ -2126,6 +2133,59 @@ def aggregate_progress_wrapper(aggregates):
...
+            scheduler_info = {}
+        with self._update_cached_scheduler_status() as status_update:
+            for (
+                aggregate_id,
+                aggregate,
+                group,
+            ) in self._generate_selected_aggregate_groups(*args, **kwargs):
+                scheduler_id = group._generate_id(aggregate)
+                scheduler_status = scheduler_info.get(scheduler_id, JobStatus.unknown)
+                status_update[scheduler_id] = int(scheduler_status)
```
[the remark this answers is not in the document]
bdice: The yielded values are instances of `JobStatus`, so the docs are correct.

The purpose of this `int(...)` cast is to reinforce that the `status_update` dictionary has to be serialized. While the values of `JobStatus` can be implicitly converted to `int` (via the enum mapping) when the project document is updated and serialized, we are currently performing the conversion explicitly just to make that more obvious and prevent unexpected failure if something else should change in the codebase. As we continue to explore adding alternative backends for signac, I think it is best to leave this behavior explicitly in the code. The code must already explicitly parse the `int` into a `JobStatus` when reading the project document.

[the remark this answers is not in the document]
bdice: Sorry. I looked at the wrong part of the docs. I see what you mean. I will update the docs to require `int` values and add manual casts any place that needs it.

flow/util/misc.py
vyasr: Coooookieeeee

b-butler (approved):
LGTM

tests/test_project.py
vyasr: Agreed, this makes sense. And the old behavior just happened to work because `test_resubmit` runs exactly three more steps than necessary (to flush things out of the queue), which happens to be the value originally included. Without an upper bound that test would run forever.

tests/define_test_project.py
vyasr: Approved by @b-butler and myself on the other thread.

tests/generate_template_reference_data.py
```diff
@@ -170,79 +180,84 @@ def main(args):
...
-                                "gpu" in parameters["partition"].lower(),
-                                "gpu" in op.lower(),
-                            ):
-                                continue
-                        fn = f"script_{op}.sh"
+        with get_masked_flowproject(p) as fp:
+            # Here we set the appropriate executable for all the operations. This
+            # is necessary as otherwise the default executable between submitting
+            # and running could look different depending on the environment.
+            executable = "/usr/local/bin/python"
```
vyasr, line 187: Save this as a module-level variable `MOCK_EXECUTABLE` that you can use both here and in the `get_mocked_flowproject` context above.

tests/generate_template_reference_data.py
```diff
@@ -143,16 +145,24 @@ def _store_bundled(self, operations):
...
+    try:
+        old_executable = sys.executable
+        sys.executable = "/usr/local/bin/python"
+        fp = TestProject.get_project(root=p.root_directory())
+        if environment is not None:
+            fp._environment = environment
+        fp._entrypoint.setdefault("path", "generate_template_reference_data.py")
+        fp._mock = True
+        fp._mock_root_directory = lambda: PROJECT_DIRECTORY
+        fp.config.project_dir = PROJECT_DIRECTORY
```
vyasr, line 162: These appear superfluous to me.
```suggestion
        fp.config.project_dir = PROJECT_DIRECTORY
```

tests/generate_template_reference_data.py
```diff
@@ -170,79 +180,84 @@ def main(args):
...
+            executable = "/usr/local/bin/python"
+            for group in fp.groups.values():
+                for op_key in group.operations:
+                    if op_key in group.operation_directives:
+                        group.operation_directives[op_key]["executable"] = executable
+            for job in fp:
+                with job:
+                    kwargs = job.statepoint()
+                    env = get_nested_attr(flow, kwargs["environment"])
+                    env.scheduler_type = FakeScheduler
```
vyasr, line 196: Add a comment for why we need to use the `FakeScheduler` here.

tests/generate_template_reference_data.py
```diff
@@ -170,79 +180,84 @@ def main(args):
...
+                    fp._environment = env
+                    parameters = kwargs["parameters"]
```
vyasr, line 198: Wow Github really has no idea how to interpret this indentation does it? Correct me if I'm wrong, nothing after this line has actually changed (other than being indented), right?

tests/test_templates.py
```diff
@@ -30,78 +31,81 @@ def find_envs():
 @pytest.mark.parametrize("env", find_envs())
 def test_env(env, monkeypatch):
     monkeypatch.setattr(flow.FlowProject, "_store_bundled", gen._store_bundled)
+    monkeypatch.setattr(env, "scheduler_type", FakeScheduler)
```
vyasr, line 34: Copy whatever comment you write in `test_project.py` here, it's useful for future devs to know why the `FakeScheduler` is important (as opposed to the `MockScheduler`).

tests/test_templates.py
```diff
@@ -30,78 +31,81 @@ def find_envs():
...
-                        # Don't try to submit GPU operations to CPU partitions
-                        # and vice versa.  We should be able to relax this
-                        # requirement if we make our error checking more
-                        # consistent.
-                        if operator.xor(
-                            "gpu" in parameters["partition"].lower(),
-                            "gpu" in op.lower(),
-                        ):
-                            continue
+        with gen.get_masked_flowproject(p, environment=env) as fp:
```
vyasr, line 41: Another quick check, nothing actually changed below other than indentation right?

tests/generate_template_reference_data.py
vyasr: Never mind, I see where these are used now. I would rather set `fp.root_directory = lambda: PROJECT_DIRECTORY` here than include code in the main code base that exists just to facilitate this test. Is there a reason that wouldn't work?

flow/project.py
```diff
@@ -1133,9 +1109,17 @@ def _generate_id(self, aggregate, operation_name=None):
...
             op_string = operation_name
 
+        # Ids are generated based on the project root directory. However, while
+        # testing, we must use a mocked value of the root directory. This is a
+        # somewhat hacky solution.
+        root_directory = (
+            project.root_directory()
+            if not getattr(project, "_mock", False)
+            else project._mock_root_directory()
+        )
```
vyasr, line 1119: I proposed an alternative to this in `generate_template_reference_data.py`, would be nice to get rid of if possible but depends on whether my proposal would work.

flow/project.py
```diff
@@ -2411,208 +2388,114 @@ def _fetch_status(
...
+            status_results_combined.append(
+                {
+                    "aggregate_id": aggregate_id,
+                    "groups": {
+                        result["group_name"]: result["status"]
+                        for result in aggregate_results
+                    },
+                    # TODO: fix operations error propagation (bdice is not
+                    # sure what that data is supposed to look like)
+                    "_error": None,
```
vyasr, line 2494: Is the issue here that we'd need to group the error messages from various groups in some consistent fashion?

flow/project.py
```diff
@@ -2896,7 +2775,11 @@ def _incomplete(status_entry):
             # get parameters info
 
             def _add_parameters(status):
-                statepoint = self.open_job(id=status["job_id"]).statepoint()
+                aggregate_id = status["aggregate_id"]
+                if aggregate_id.startswith("agg-"):
+                    # TODO: Fill parameters with empty values (or shared values?)
```
vyasr, line 2780: Is this TODO regarding the error message, or deciding how this actually should behave for aggregates once we implement it?

vyasr (changes requested):
This is definitely a big improvement in the code. There are a few more things to be done before we can merge.

flow/util/misc.py
```diff
@@ -303,3 +309,77 @@ def __iter__(self):
...
+    the function object a priori and bind it as the first argument to a partial
+    application of this function. All subsequent arguments are transparently
+    passed through.
+    """
+    unpickled_func = cloudpickle.loads(func)
+    args = list(map(cloudpickle.loads, args))
+    return unpickled_func(*args)
+
+
+def _get_parallel_executor(parallelization="thread"):
```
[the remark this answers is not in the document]
bdice: Good question. I aimed to make my benchmarks representative-ish, but I would like one or two real-world benchmarks from users before making that decision. Ideally something on an HPC cluster with a "realistic" file system (not my local M.2 SSD). Would you be willing to try both options on a project of yours?

flow/util/misc.py
```diff
@@ -4,12 +4,18 @@
...
 import os
 from collections.abc import MutableMapping
 from contextlib import contextmanager
 from functools import lru_cache, partial
 from itertools import cycle, islice
 
+import cloudpickle
+from tqdm.auto import tqdm
+from tqdm.contrib import tmap
+from tqdm.contrib.concurrent import thread_map
```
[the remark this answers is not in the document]
bdice: No, I didn't use `process_map` because of ... reasons that may no longer exist? I agree that would be good. I'll try again.

tests/generate_template_reference_data.py
bdice: Yes, that won't work because we have to mock the root directory used _for operation id generation_ separately from the real root directory that contains the jobs. We need consistently generated ids (using the mocked value) but we still need the root directory to be correct or else we break signac (maybe instantiating jobs fails? Or the config is inaccessible? Or it can't pickle/unpickle? Something like that, but I forgot the exact reason).

tests/generate_template_reference_data.py
bdice: I believe that's correct, nothing else changed here that I recall.

tests/test_templates.py
bdice: Correct.

flow/project.py
bdice: Ah, I forgot about this TODO. This needs to be handled. If errors occur, that information is propagated through the status somehow. I need to figure out what the "error" data structure should look like and group it, as you indicated.

flow/project.py
bdice: This can be deferred, I just didn't know what to do if the user requests to show state point parameters and aggregates exist. It might make sense to show common parameters if we put aggregates with the rest of the rows of jobs...? Regardless, that's a hairy design question and can be left as TODO for the purposes of this PR.

flow/project.py
bdice: The proposal won't work - see other comment. I will resolve this and continue the conversation on the other thread.

flow/util/misc.py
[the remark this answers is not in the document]
bdice: `process_map` works and it is considerably simpler. Good call.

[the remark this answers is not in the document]
bdice: We have to bump our tqdm version requirement regardless, >=4.42.0 is needed for both `thread_map` and `process_map`.

flow/util/misc.py
```diff
@@ -303,3 +309,77 @@ def __iter__(self):
...
+    if parallelization == "thread":
+        parallel_executor = thread_map
+    elif parallelization == "process":
+
+        def parallel_executor(func, iterable, chunksize=1, **kwargs):
+            # tqdm progress bar requires a total
+            if "total" not in kwargs:
+                kwargs["total"] = len(iterable)
+
+            with multiprocessing.Pool() as pool:
```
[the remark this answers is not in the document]
bdice: I switched to tqdm's `process_map` which uses concurrent futures internally.

tests/generate_template_reference_data.py
bdice: Very nice refactor. Good find.

flow/project.py
bdice: This is resolved now.

flow/util/misc.py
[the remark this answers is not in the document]
vyasr: Yes, I'm happy to try and test this somewhere. Since the choice of default is independent of this PR I'll make a separate issue for it.

tests/generate_template_reference_data.py
vyasr: Got it, you're right that this won't work in that case. How about the following context manager instead:

```
@contextmanager
def wrap_id_generation(fp):
    """Create a context where id generation will use a mocked root directory."""
    orig_generate_id = fp._generate_id 
    orig_root_directory = fp.root_directory
    mocked_root_directory = lambda: PROJECT_DIRECTORY

    def wrapped_generate_id(self, aggregate, operation_name=None):
        """Mock the root directory used for id generation to ensure consistent ids."""
        # This mocking has to happen within this method to avoid affecting other methods called during the test.
        self.root_directory = mocked_root_directory
        _generate_id(aggregate, operation_name)
        self.root_directory = orig_root_directory
    
    fp._generate_id = wrapped_generate_id 
    yield
    fp._generate_id = orig_generate_id
```
We don't execute anything in parallel here, so I think this should be safe. It's a few more lines of code than the current solution, but I would really like to avoid injected this code into flow/project.py if at all possible.

bdice: You're right, this would be much better. I'll adapt this.

flow/project.py
vyasr: I'll make an issue from this as well, then.

vyasr:
I'm happy with this PR once the mocking of the root directory has been handled as requested.

vyasr (approved):
LGTM!

PR #421, opened 2021-01-08
Optimization: don't deepcopy directives

### LINKED ISSUE #420: Optimization: don't deepcopy _Directives
https://github.com/glotzerlab/signac-flow/blob/ece9c2495a625b8c7d4b2f3efdba60827c7af062/flow/project.py#L1197-L1203

While I was profiling `submit`, I saw that the function `_create_run_job_operations` performs a deep copy of a `_Directives` instance. In the profiles, I see that this `deepcopy` is taking nearly 50% of the run time (far more than the actual generation of aggregates/groups).

image

I believe this can be significantly improved by only copying the necessary parts. At this point in the code, the directives have already been evaluated. The constructor for `_JobOperation` only _needs_ two parts of the `directives` object:
1. The names of the keys set by the user, `set(directives._user_directives)`
2. The keys/values of the evaluated directives, which are used as `TrackGetItemDict(directives)`

The deep copy is doing a lot of extra work because the `_Directives` object has many complex parts that are not pure data. For example, every `_Directive` contained in the `_Directives` mapping holds a variety of function objects. These are only useful up to the point of evaluation, and thus do not need to be re-created in the deep copy.

I recommend that we change the constructor of `_JobOperation` to accept `directives` as a _plain dict_ and `user_directive_keys` as a set of keys.

Tagging @b-butler @kidrahahjo since you previously worked on this.

b-butler:
## Description

This changes the call to deepcopy `_Directives` instances to `dict` conversions. We also manually pass in the user specified directives for `_JobOperation`. I also implemented a small improvement to the evaluation logic for `_Directives` instances by only evaluating (changing callable directives to concrete values) when unevaluated. I can revert this last change if necessary.

## Motivation and Context

Resolves #420. This provides an opportunity for major performance improvements.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -1294,7 +1294,7 @@ class FlowGroup:
             self.name,
             jobs,
             cmd=unevaluated_cmd,
-            directives={**submission_directives},
+            directives=dict(submission_directives),
             user_directives=set(submission_directives.user_keys),
             eligible_operations=eligible_operations,
             operations_with_unmet_preconditions=operations_with_unmet_preconditions,
@@ -1362,7 +1362,7 @@ class FlowGroup:
                     operation_name,
                     jobs,
                     cmd=unevaluated_cmd,
-                    directives={**directives},
+                    directives=dict(directives),
                     user_directives=set(directives.user_keys),
                 )
                 # Get the prefix, and if it's non-empty, set the fork directive
--- a/flow/project.py
+++ b/flow/project.py
@@ -352,8 +352,14 @@ class _JobOperation:
 
         # We use a special dictionary that tracks all keys that have been
         # evaluated by the template engine and compare them to those explicitly
-        # set by the user. See also comment above.
+        # set by the user. See also comment below.
         self.directives = TrackGetItemDict(directives)
+
+        # Keys which were explicitly set by the user, but are not evaluated by
+        # the template engine are cause for concern and might hint at a bug in
+        # the template script or ill-defined directives. We are therefore
+        # keeping track of all keys set by the user and check whether they have
+        # been evaluated by the template script engine later.
         self.directives._keys_set_by_user = user_directives
 
     def __str__(self):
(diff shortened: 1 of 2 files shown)


flow/project.py
```diff
@@ -334,33 +334,27 @@ class _JobOperation:
...
-        A :class:`flow.directives._Directives` object of additional parameters
-        that provide instructions on how to execute this operation, e.g.,
-        specifically required resources.
-
+    directives : dict
+        A `dict` object of additional parameters that provide instructions on
+        how to execute this operation, e.g., specifically required resources.
+    user_directives : set
+        User specified directives that were not part of the environment's
+        standard directives.
```
bdice, line 342:
```suggestion
        Keys in ``directives`` that correspond to user-specified directives
        that are not part of the environment's standard directives.
```

flow/project.py
```diff
@@ -1294,7 +1288,8 @@ def _get_run_ops(ignore_ops, additional_ignores_flag):
             self.name,
             jobs,
             cmd=unevaluated_cmd,
-            directives=submission_directives,
+            directives={**submission_directives},
```
bdice, line 1291: I would prefer a constructor call rather than a comprehension here, since there's no "comprehending" to be done.
```suggestion
            directives=dict(submission_directives),
```

flow/project.py
```diff
@@ -1361,7 +1356,8 @@ def _create_run_job_operations(
                     operation_name,
                     jobs,
                     cmd=unevaluated_cmd,
-                    directives=deepcopy(directives),
+                    directives={**directives},
```
bdice, line 1359:
```suggestion
                    directives=dict(directives),
```

bdice (approved):
This fixes the performance problems noted in #420. I pushed one commit to fix a comment (it would have been difficult to make an inline suggestion) and a few minor edits, otherwise LGTM. I will go ahead and commit the minor edits and merge, I would like to incorporate this into my current PR.

image

PR #422, opened 2021-01-12
Refactor aggregation

bdice:
## Description
- Adds base class `_BaseAggregateStore` that sets up the abstractions that are used by both `_AggregateStore` and `_DefaultAggregateStore`
- Fixes a number of docstrings that were no longer accurate
- Renames some classes/methods for consistency in naming
- Moves `_get_unique_function_id` to module scope instead of making it a method of the `aggregator` class
- Improves some method implementations in the `aggregator` class
- Improves some aggregator tests, renames tests to clarify their purpose
- Adds minor feature to allow `sort_by` with a callable function of `job` (passed through to `sorted`) with new tests
- Moved filtering/validation logic directly into the aggregator functions. It is no longer allowed for an aggregator function to return iterables containing `None` with the expectation that those `None` values will be automatically removed.
- Added more tests for ensuring that all aggregates are tuples of jobs.

## Motivation and Context
Cleaning up aggregation while preparing #417.

## Types of Changes
- [x] Documentation update
- [x] Bug fix
- [x] New feature
- [x] Breaking change

## Checklist:

If necessary:


flow/aggregates.py
```diff
@@ -332,18 +351,42 @@ def __call__(self, func=None):
             )
 
 
-class _AggregatesStore(Mapping):
+class _BaseAggregateStore(Mapping):
```
kidrahahjo, line 354: I am not sure about whether we should be creating a base class for aggregate stores because of the way we use it.

flow/aggregates.py
```diff
@@ -8,10 +8,40 @@
 default aggregator produces individual jobs.
 """
 import itertools
+from abc import abstractmethod
 from collections.abc import Iterable, Mapping
 from hashlib import md5
 
 
+def _get_unique_function_id(func):
```
kidrahahjo, line 16: Since this is a utility function which currently only serves the `aggregator` class, I think this can be a static method.

tests/test_aggregates.py
```diff
@@ -129,131 +129,144 @@ def test_groupby_with_invalid_length_default_key_for_Iterable(self):
...
         # This test ensures this feature.
-        # list_of_aggregates contains 14 distinct storing objects (because an
+        # list_of_aggregators contains 14 distinct store objects (because an
         # aggregator object is differentiated on the basis of the `_is_aggregate` attribute).
         # When this list is converted to set, then these objects are hashed first
         # and then compared. Since sets don't carry duplicate values, we test
         # whether the length of the set obtained from the list is equal to 14 or not.
-        assert len(set(list_of_aggregates)) == 14
+        assert len(set(list_of_aggregators)) == 14
+        # Ensure that equality implies hash equality.
```
kidrahahjo, line 142: @bdice I am unable to understand why do we need this test as python sets use hash values and equality checks (right?)

tests/test_aggregates.py
```diff
@@ -263,50 +276,63 @@ def keyfunction(job):
...
+        # whether the length of the set obtained from the list is equal to 14
+        # or not.
+        assert len(set(list_of_stores)) == 14
+
+    def test_aggregates_are_tuples(self, setUp, project, list_of_aggregators):
+        # This test ensures that all aggregators return tuples.
+        for aggregator_instance in list_of_aggregators:
+            aggregate_store = aggregator_instance._create_AggregateStore(project)
+            if not isinstance(aggregate_store, _DefaultAggregateStore):
+                for aggregate in aggregate_store._generate_aggregates():
```
kidrahahjo, line 316:
```suggestion
                for aggregate in aggregate_store.values():
```
When we create `aggregate_store`, all the aggregates are automatically calculated hence I think we should go with `aggregate_store.values()` here.

Hence, this would also mean that we don't need an `if-else` block here

kidrahahjo (commented):
@bdice I like these changes but I am not really sure if we should introduce the class `_BaseAggregateStore`. Can you please help me understand why we introduced this class?

flow/aggregates.py
bdice: Although this _could_ be a static method, we generally prefer to define helper methods over static methods in our code base. In this particular case, the code doesn't even have anything to do with the `aggregator` class, so calling it as `aggregator._get_unique_function_id(func)` would indicate an association between the class and function even though they're not related.

flow/aggregates.py
```diff
@@ -226,90 +270,65 @@ def keyfunction(job):
...
         return cls(aggregator_function, sort_by, sort_ascending, select)
 
     def __eq__(self, other):
         """Test equality with another aggregator."""
-        return (
-            type(self) == type(other)
-            and not self._is_aggregate
-            and not other._is_aggregate
-        )
+        if not isinstance(other, type(self)):
```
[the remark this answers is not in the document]
bdice: This isn't an error, it defers to the other object's `__eq__` method (if implemented). `NotImplemented` is a special value (it's different than `NotImplementedError`). I described what this does and linked to references in this recent PR to signac: https://github.com/glotzerlab/signac/pull/455

vyasr:
@bdice @kidrahahjo @b-butler I have not reviewed this PR at all, but it's worth noting that I asked about exactly the same combination of the aggregate storage classes [here](https://github.com/glotzerlab/signac-flow/pull/348/files/3af7e0e7454f1a4a4d46faf5f62a7bfb6b131414#r478688950). The argument against doing it was that they ultimately share no logic, so any parent class would just be an empty shell (perhaps indicating an interface, but a) that interface would be purely internal rather than user-facing, and b) the bulk of the interface is dunder methods anyway). For something that (at least for now) doesn't appear to require any further generalization, the value of creating a common base class just to have them in the same inheritance tree (especially if they're not exposing any common APIs) is questionable.

flow/aggregates.py
bdice: There are a few reasons why this is helpful:
- Creates a standard _interface_ for both aggregate store classes. It's an abstract class because it doesn't implement the requirements of its parent class `Mapping`, and thus `_BaseAggregateStore` cannot be instantiated.
- If we want to have a method with an implementation that is shared by both aggregate store classes, this would allow us to share that logic. For example, I can (and should) refactor `__iter__` to be a base class method that performs `yield from self.keys()`.
- This class allows us to check `isinstance(obj, _BaseAggregateStore)` if we just want _some_ aggregate store without worrying about whether it's a default or non-default.

flow/aggregates.py
kidrahahjo: Okay! But since this is not closely related to aggregation, should we put this private function inside the utils directory?

tests/test_aggregates.py
bdice: We implement both `__hash__` and `__eq__` for the aggregator class. It is a requirement of the Python data model that `a == b` implies `hash(a) == hash(b)`. Since our implementation of the methods doesn't necessarily guarantee that (if they were implemented incorrectly, for example), I am verifying that manually with this test.

Python sets can contain objects with the same hash as long as they are not equal. Example:

```python
class A:
    def __hash__(self):
        # All instances have the same hash.
        return 42
    def __eq__(self, other):
        # Instances are never equal.
        return False

a1 = A()
a2 = A()
assert len({a1, a2}) == 2
```

tests/test_aggregates.py
bdice: My intention here was to test that the internal aggregator functions called by `_generate_aggregates` all return tuples. This behavior is intentional. I'll add a clarifying comment.

tests/test_aggregates.py
kidrahahjo: I now understand this, thanks!

tests/test_aggregates.py
kidrahahjo: Awesome, thanks.

bdice:
> 
> 
> @bdice @kidrahahjo @b-butler I have not reviewed this PR at all, but it's worth noting that I asked about exactly the same combination of the aggregate storage classes [here](https://github.com/glotzerlab/signac-flow/pull/348/files/3af7e0e7454f1a4a4d46faf5f62a7bfb6b131414#r478688950). The argument against doing it was that they ultimately share no logic, so any parent class would just be an empty shell (perhaps indicating an interface, but a) that interface would be purely internal rather than user-facing, and b) the bulk of the interface is dunder methods anyway). For something that (at least for now) doesn't appear to require any further generalization, the value of creating a common base class just to have them in the same inheritance tree (especially if they're not exposing any common APIs) is debatable. Since I haven't reviewed this yet I don't want to take any position yet, but unless the underlying implementation has changed dramatically the old arguments might still apply.

Having a common internal interface was my primary motivation for introducing the base class. I think it's an appropriate and conceptually helpful abstraction. I agree that not much of the implementation is shared, but I think indicates an even _stronger_ motivation to provide developers with a clear understanding that these classes share a common set of concepts even though the implementation differs.

flow/aggregates.py
bdice: Possibly. But keeping it in the same module seems best until we use it elsewhere.

vyasr:
> > @bdice @kidrahahjo @b-butler I have not reviewed this PR at all, but it's worth noting that I asked about exactly the same combination of the aggregate storage classes [here](https://github.com/glotzerlab/signac-flow/pull/348/files/3af7e0e7454f1a4a4d46faf5f62a7bfb6b131414#r478688950). The argument against doing it was that they ultimately share no logic, so any parent class would just be an empty shell (perhaps indicating an interface, but a) that interface would be purely internal rather than user-facing, and b) the bulk of the interface is dunder methods anyway). For something that (at least for now) doesn't appear to require any further generalization, the value of creating a common base class just to have them in the same inheritance tree (especially if they're not exposing any common APIs) is debatable. Since I haven't reviewed this yet I don't want to take any position yet, but unless the underlying implementation has changed dramatically the old arguments might still apply.
> 
> Having a common internal interface was my primary motivation for introducing the base class. I think it's an appropriate and conceptually helpful abstraction. I agree that not much of the implementation is shared, but I think indicates an even _stronger_ motivation to provide developers with a clear understanding that these classes share a common set of concepts even though the implementation differs.

I agree that can be valuable, and it's what I was originally requesting as well. If that is what you intend, however, the class needs to cleanly delineate the interface, which provides a concrete description of what concepts the class encodes. That means adding abstract methods corresponding to all the dunder methods that must exist for the object to be used anywhere an `AggregateStore`-like object is valid, e.g. `__getitem__`, `__len__`, etc. Since it's for developers, it would also be helpful (although not necessary) to include information in the docstring indicating which methods are performance critical since you've found this to be a bottleneck, although I'd be less of a stickler for that since performance concerns can change based on internal implementation changes that otherwise have no bearing on these classes.

kidrahahjo (approved):
The changes looks good to me.

flow/aggregates.py
```diff
@@ -497,7 +529,7 @@ def __getitem__(self, id):
         try:
             return (self._project.open_job(id=id),)
         except KeyError:
-            raise LookupError(f"Did not find aggregate with id {id}.")
+            raise KeyError(f"Aggregate id {id} is not in this aggregate store.")
```
kidrahahjo, line 532: Since the store classes are private, I think we should not mention about them anywhere in the "user-space".

kidrahahjo (approved):
Apologies for approving without any comment earlier, I lost this review and forgot about it before submitting. Here is one final comment, and if you're okay with it, I approve.

flow/aggregates.py
kidrahahjo: There is one more place in aggregates.py where this error message is thrown.

bdice:
The signac 1.0 tests will fail on this branch and can be ignored. PR #417 fixes this.

flow/aggregates.py
bdice: Seems reasonable to me. I'll revert the message but I'll keep it as a `KeyError`.

tests/test_aggregates.py
```diff
@@ -263,50 +282,66 @@ def keyfunction(job):
...
+        ]
+        assert len(list_of_stores) == len(list_of_aggregators)
+        # The above list contains 14 distinct store objects (because a
+        # store object is differentiated on the basis of the
+        # ``_is_default_aggregate`` attribute of the aggregator). When this
+        # list is converted to a set, then these objects are hashed first and
+        # then compared. Since sets don't carry duplicate values, we test
+        # whether the length of the set obtained from the list is equal to 14
+        # or not.
+        assert len(set(list_of_stores)) == 14
```
vyasr, line 315: I'm confused why we need both this test and the one below. The one above the comment seems better since it's robust to the fixture length changing.

vyasr (changes requested):
Almost there.

flow/aggregates.py
```diff
@@ -36,15 +66,18 @@ def foo(*jobs):
...
         jobs and can return or yield subsets of jobs as an iterable. The
         default behavior is creating a single aggregate of all jobs.
-    sort_by : str or None
-        Before aggregating, sort the jobs by a given statepoint parameter.
-        The default behavior is no sorting.
+    sort_by : str, callable, or None
+        Before aggregating, sort the jobs by a given statepoint parameter. If
+        the argument is callable, this will be passed as the callable
+        argument to :func:`sorted`. If None, no sorted is performed. (Default
+        value = None)
```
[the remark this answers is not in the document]
bdice: I don't think we're inconsistent - I think we use the period-before in most/all cases. That style was favored at some point by a formatting tool, perhaps when converting from Sphinx-style to NumPy-style docstrings. I just stuck with it.

flow/aggregates.py
```diff
@@ -332,18 +351,45 @@ def __call__(self, func=None):
...
+    an aggregate is defined as a tuple of instances of
+    :class:`signac.contrib.job.Job`.
+    """
+
+    def __init__(self, project):
+        self._project = project
+        self._register_aggregates()
+
+    @abstractmethod
+    def _register_aggregates(self):
```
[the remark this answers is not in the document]
bdice: Good idea. Fixed.

tests/test_aggregates.py
bdice: These are testing different things: all default aggregators are equivalent, so `len(set(list_of_stores)) != len(list_of_stores)`. The `set` removes some aggregators that have different parameters provided to the constructor but that are all equivalent (default aggregators). We can't easily make this test robust to the fixture length changing (only something like `len(fixtures) - num_known_duplicates`, which I don't think is helpful).

flow/aggregates.py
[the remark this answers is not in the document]
vyasr: I checked through signac, signac-flow, freud, and coxeter as a set of representative examples (just a simple ripgrep for "Default value"). While we're not 100% consistent anywhere, signac, freud, and coxeter all have a majority (I'd say ~80%) of the period after. The only exception is signac-flow, and since you did a lot of work on the numpydoc conversion as well as your recent dev pushes I suspect that you have (consistently) chosen the opposite convention we've preferred in all our other code and probably changed preexisting period-after docstrings. Let's revert those changes incrementally as we find them.

tests/test_aggregates.py
vyasr: I missed the `set` call, thanks for the clarification.

PR #424, opened 2021-01-13
Remove the use of arguments scheduled to be removed in 0.12

kidrahahjo:
## Description
Remove the arguments which were scheduled to be removed in `flow` version 0.12.

@glotzerlab/signac-maintainers  I am not exactly sure why the tests fail. I think it's because while testing, the project uses `test_project.MockEnvironment` for creating `JobOperation` instances and while "submitting" it uses the user-defined environment. I think we should update the template reference data but I will wait for your comments on this.

## Motivation and Context
The `env` and `no_parallelize` arguments were deprecated in flow version 0.10 and are scheduled to be removed in 0.12.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:


bdice:
@kidrahahjo The current template reference data is correct. We need to make sure that the templates are being submitted with the correct environments. I'll look into it.

bdice (approved):
@kidrahahjo I looked into this and I don't think the problem was introduced by your changes. It looked like somewhere along the line, the template reference data was using "mpiexec" where it should have used the cluster-specific MPI commands like "mpirun" and "ibrun" and "jsrun". It looks like this PR was behaving correctly and the template reference data was wrong. I'm not sure how this occurred. I spent some time trying to figure out what went wrong but the issue is solved now (whatever its cause) and I don't have time to go any deeper.

I pushed a couple of commits that fixed the documentation (you removed the deprecated features but not their docs). Otherwise this is good to go. Thanks!

Issue #414, opened 2021-01-02
Remove unbuffered mode


bdice:
### Feature description

Currently we support an "unbuffered" mode in signac-flow that disables signac's buffering capabilities, by disabling the config option `use_buffered_mode` (it's on by default). I don't know if it still makes sense to support an unbuffered mode because we require `signac >= 1.0`. I believe buffering was initially an optional feature (rather than always-on) because signac didn't support buffering in some earlier versions. Our CI takes a long time to run and testing both options (on/off) seems unnecessary at this point.

### Proposed solution

The option defaults to `True`, and I think this feature could be immediately removed with no deprecation cycle -- in my understanding, it's basically a flag that disables an important performance optimization. However, I don't want to immediately remove it from the config validator (that's more of a breaking change than just requiring buffered mode). I suggest the following.

Deprecation:
- Change the behavior of `FlowProject._potentially_buffered` to warn if `use_buffered_mode` is specified and _ignore the option_ (always buffer).
- Remove the tests for unbuffered projects.

Removal (after 2 minor versions):
- Remove `use_buffered_mode` option from the config validator (config files with the option specified are no longer valid).
- The `FlowProject._potentially_buffered` context manager could be removed and replaced with: `with signac.buffered():` or we could just change the name to `FlowProject._buffered`.

Before anyone begins work on this issue, I would like to get input from @glotzerlab/signac-committers.

atravitz:
Doing a short deprecation cycle for this seems reasonable to me 👍

vyasr:
The main hesitation I would have with changing this immediately is relating to glotzerlab/signac#397. It should be relatively easy to modify the new code for glotzerlab/signac#249 to handle thread parallelism transparently (and I plan to do this), but I don't know of any way to make it intrinsically safe for process parallelism. If a signac-flow user sets the `status_parallelization` option to `'process'`, I think signac-flow will have to be responsible for doing the update in a parallel-safe manner. It's possible that future refactorings in flow will make this easier to achieve (or a moot point), but so long as that limitation exists the options are:

1. Always disable buffering when using process-parallel status updates. This will always be safe, but may end up making things slow enough that process parallelism becomes useless. Since switching to process-parallel status updates indicates that the user has identified processing rather than I/O as the bottleneck, there's no way to know for sure without testing on a particular workflow.
[...]

PR #425, opened 2021-01-13
Deprecate unbuffered mode

bdice:
## Description
Deprecates `use_buffered_mode` config option and always enables buffering. The "unbuffered" mode is probably not helpful and the tests of unbuffered mode are quite long (which extends our CI time considerably).

**With these changes, the CI tests take ~4 minutes instead of ~26 minutes.** 🐎

## Motivation and Context
Solves the first part of #414. A follow-up PR will need to be created for 0.14 that removes the config option and (possibly) replaces the `FlowProject._buffered` context manager with `signac.buffered`.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


tests/test_project.py
```diff
@@ -144,12 +144,13 @@ def recursive_update(d, u):
...
             project = project_class(config=config)
-        for a in range(3):
+        for a in range(2):
             if heterogeneous:
                 # Add jobs with only the `a` key without `b`.
                 project.open_job(dict(a=a)).init()
                 project.open_job(dict(a=dict(a=a))).init()
-            for b in range(3):
+            # Tests assume that there are even and odd values of b
+            for b in range(2):
```
bdice, line 153: I discussed this change with @vyasr. It should be fine for us to reduce the number of jobs from 9 to 4, by reducing the number of values for `a` and `b`. I scanned the tests for any place where this might reduce coverage but I don't see any issues.

flow/project.py
```diff
@@ -3614,15 +3614,19 @@ def _convert_jobs_to_aggregates(self, jobs):
...
-        if self.config["flow"].as_bool("use_buffered_mode"):
-            logger.debug("Entering buffered mode...")
-            with signac.buffered():
-                yield
-            logger.debug("Exiting buffered mode.")
-        else:
+        if not self.config["flow"].as_bool("use_buffered_mode"):
+            warnings.warn(
+                "The use_buffered_mode config option is deprecated as of 0.12 "
+                "and will be removed in 0.14. Buffered mode is always enabled "
```
kidrahahjo, line 3622: If we keep the buffering mode always enabled then I think we should remove this in 0.13 instead.

kidrahahjo (approved):
One minor suggestion, else I approve of these changes.

flow/project.py
bdice: I read our deprecation policy again and that makes sense. https://docs.signac.io/projects/flow/en/latest/support.html#deprecation-policy
```suggestion
                "and will be removed in 0.13. Buffered mode is always enabled "
```

atravitz (approved):
looks good!

bdice:
There is a failing test for signac 1.0, but I fixed this on another PR. I will merge with the failing test and handle it in #417.

PR #434, opened 2021-01-21
Fix jinja templates for status output.

bdice:
## Description
Fixes a regression in status templates noted by @atravitz. Job ids and group names were not being handled correctly in the Jinja templates. This bug was introduced in #417.

We should probably add tests for status printing (possibly regression tests like we have for template submission). Tests would have caught this problem. I created issue #435.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:


atravitz (approved):
I tested this locally and it looks good! We should probably add tests for this though, since it's kind of a critical feature.

PR #436, opened 2021-01-22
Add oldest & minimal CI configurations.

kidrahahjo:
## Description

- Adds "oldest" CI configuration. Python 3.6 may no longer be supported by the latest versions of our testing dependencies.
- Re-organizes all the requirements files into a subdirectory called requirements.
- Add minimum version 1.1.1 for cloudpickle in the requirements file. (Tests fail on older versions)
- Drops support for `Jinja2` versions older than `2.10.0`.
- Drops support for `tqdm` versions older than `4.43.0`.

## Motivation and Context

Replicates https://github.com/glotzerlab/signac/pull/474 for `flow`.
## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


kidrahahjo:
@csadorf @bdice @tommy-waltmann The `contrib` submodule in `tqdm` was added `4.40.0` (Check [release notes](https://tqdm.github.io/releases/#v4400-2019-12-01))

I think we should pin the version in the `requirement` files and `setup.py` accordingly.

bdice:
@kidrahahjo On the topic of tqdm, we need 4.42.0 or newer to have the wrappers around concurrent futures. See this comment in #417:
> Uses a tqdm wrapper around concurrent futures. Note: if we choose to use this approach (which is a lot less total code), we will need to bump our tqdm version requirement to tqdm>=4.42.0.

kidrahahjo:
> @kidrahahjo On the topic of tqdm, we need 4.42.0 or newer to have the wrappers around concurrent futures. See this comment in #417:
> 
> > Uses a tqdm wrapper around concurrent futures. Note: if we choose to use this approach (which is a lot less total code), we will need to bump our tqdm version requirement to tqdm>=4.42.0.
> 
> https://github.com/tqdm/tqdm/releases/tag/v4.42.0

@bdice should I go ahead and change the version numbers in both requirements file and `setup.py` to `4.42.0` for `tqdm`?

kidrahahjo:
@csadorf  @bdice I have Python 3.8.5 running on my machine and it seems like we don't support `Jinja2` versions older than `2.10`.

bdice:
> @csadorf @bdice I have Python 3.8.5 running on my machine and it seems like we don't support `Jinja2` versions older than `2.10`.

That’s correct. There are template filters we require from 2.10.

kidrahahjo:
I found out that the minimum tqdm version we require is 4.43.0! The tests are failing for anything below it.

kidrahahjo:
The tests will pass now, I think this is ready! @csadorf @bdice @tommy-waltmann please review whenever you get time! Thanks.

.circleci/config.yml
```diff
@@ -54,14 +55,24 @@ jobs:
...
+            python -m pip install --progress-bar off -U pip>=20.3
             python -m pip install --progress-bar off -U codecov
+            python -m pip install --progress-bar off -U ${SIGNAC_VERSION}
+            # DEPENDENCIES can be "OLDEST", or "NEWEST"
+            if [ "${DEPENDENCIES}" == "OLDEST" ]; then
+              python -m pip install --progress-bar off \
+                jinja2==2.10 \
+                cloudpickle==1.1.1 \
+                deprecation==2 \
+                tqdm==4.43.0
```
tommy-waltmann, line 71: Would it be best to just put these dependencies in another file, similar to `requirements.txt`, and `requriements-test.txt`?

tommy-waltmann (commented):
I just have one idea that might make things look a little cleaner.

.circleci/config.yml
kidrahahjo: This is a nice idea, and I think making a file called `requirements-oldest.txt` should do the work. But we'd then have to configure `depandabot` accordingly because I think it modifies all the files having the requirements keyword in it. (Which would make the implementation more complicated). 

@tommy-waltmann let me know what you think?

tommy-waltmann: I'm not sure how re-configuring `dependabot` would work, I think @bdice or @csadorf would be better able to comment on that.

bdice: Right, as @kidrahahjo noted, the challenge is that we don't want dependabot to update the "oldest" dependency file. We might be able to add an exception but I couldn't figure out how to do that from the dependabot documentation. If someone else can identify a solution for that, I would definitely prefer that to hard-coding the "oldest" requirements in the CI config.

tommy-waltmann: If `dependabot` just looks for files that have the `requirements` substring in them, we could just have a file with a different name.

bdice:
> 
> 
> I found out that the minimum tqdm version we require is 4.43.0! The tests are failing for anything below it.

Good find! It's a fix that was in tqdm PR 893. https://tqdm.github.io/releases/#v4430-2020-02-19

.circleci/config.yml
bdice: I believe that dependabot looks in all files named `*requirements*.txt` and also `requirements/*.txt`, so we could name it `.circleci/ci-oldest-reqs.txt`?

kidrahahjo: This is now done!

kidrahahjo:
I have created a requirements file called `.circleci/ci-oldest-reqs.txt` for testing the oldest versions. In that file, I also pinned the `signac` version to `1.0.0`.

.circleci/ci-oldest-reqs.txt
```diff
@@ -0,0 +1,5 @@
+cloudpickle==1.1.1
+deprecation==2
```
bdice, line 2:
```suggestion
deprecation==2.0.0
```

.circleci/config.yml
```diff
@@ -7,10 +7,10 @@ version: 2.1
 references:
   restore_keys: &restore_keys
     keys:
-      - python-env-v1-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "requirements-test.txt" }}-{{ checksum "requirements-precommit.txt" }}-{{ checksum ".pre-commit-config.yaml" }}
+      - python-env-v1-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "requirements/requirements-test.txt" }}-{{ checksum "requirements/requirements-precommit.txt" }}-{{ checksum ".pre-commit-config.yaml" }}
```
bdice, line 10:
```suggestion
      - python-env-v1-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "requirements/requirements-test.txt" }}-{{ checksum "requirements/requirements-precommit.txt" }}-{{ checksum ".circleci/ci-oldest-reqs.txt" }}-{{ checksum ".pre-commit-config.yaml" }}
```

.circleci/config.yml
```diff
@@ -7,10 +7,10 @@ version: 2.1
...
 
   save_key: &save_key
-    key: python-env-v1-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "requirements-test.txt" }}-{{ checksum "requirements-precommit.txt" }}-{{ checksum ".pre-commit-config.yaml" }}
+    key: python-env-v1-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "requirements/requirements-test.txt" }}-{{ checksum "requirements/requirements-precommit.txt" }}-{{ checksum ".pre-commit-config.yaml" }}
```
bdice, line 13:
```suggestion
    key: python-env-v1-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "requirements/requirements-test.txt" }}-{{ checksum "requirements/requirements-precommit.txt" }}-{{ checksum ".circleci/ci-oldest-reqs.txt" }}-{{ checksum ".pre-commit-config.yaml" }}
```

.circleci/config.yml
```diff
@@ -54,14 +55,20 @@ jobs:
...
             mkdir -p ./venv
-            python -m virtualenv ./venv
+            python -m virtualenv ./venv --clear
             . venv/bin/activate
-            python -m pip install --progress-bar off ${SIGNAC_VERSION}
-            python -m pip install --progress-bar off -r requirements.txt
-            python -m pip install --progress-bar off -r requirements-test.txt
+            python -m pip install --progress-bar off -U pip>=20.3
             python -m pip install --progress-bar off -U codecov
+            # DEPENDENCIES can be "OLDEST", or "NEWEST"
```
bdice, line 64:
```suggestion
            # DEPENDENCIES can be "OLDEST" or "NEWEST"
```

.circleci/config.yml
```diff
@@ -98,8 +109,11 @@ jobs:
     <<: *test-template
     docker:
       - image: circleci/python:3.7
-  linux-python-36:
+  linux-python-36-oldest:
     <<: *test-template
+    environment:
+      DEPENDENCIES: "OLDEST"
+      SIGNAC_VERSION: "signac"
```
bdice, line 116: I'd like to delete the pinning for `signac` from the "oldest" requirements and let that be handled by the `${SIGNAC_VERSION}` variable.
```suggestion
      SIGNAC_VERSION: "signac==1.0.0"
```

.circleci/config.yml
```diff
@@ -54,14 +55,20 @@ jobs:
...
-            python -m pip install --progress-bar off ${SIGNAC_VERSION}
-            python -m pip install --progress-bar off -r requirements.txt
-            python -m pip install --progress-bar off -r requirements-test.txt
+            python -m pip install --progress-bar off -U pip>=20.3
             python -m pip install --progress-bar off -U codecov
+            # DEPENDENCIES can be "OLDEST", or "NEWEST"
+            if [ "${DEPENDENCIES}" == "OLDEST" ]; then
+              python -m pip install --progress-bar off -r .circleci/ci-oldest-reqs.txt
+            else
+              python -m pip install --progress-bar off ${SIGNAC_VERSION}
```
bdice, line 68: I'd like to delete the pinning for `signac` from the "oldest" requirements and let that be handled by the `${SIGNAC_VERSION}` variable.

Move this line outside the "if", install `signac` before or after the rest of the dependencies (this should run for both OLDEST and NEWEST).

.circleci/ci-oldest-reqs.txt
```diff
@@ -0,0 +1,5 @@
...
+jinja2==2.10
+signac==1.0.0
```
bdice, line 4: I'd like to delete the pinning for `signac` from the "oldest" requirements and let that be handled by the `${SIGNAC_VERSION}` variable.
```suggestion
```

.circleci/config.yml
```diff
@@ -77,18 +84,22 @@ jobs:
   linux-python-39-signac-10:
     <<: *test-template
     environment:
+      DEPENDENCIES: "NEWEST"
       SIGNAC_VERSION: "signac~=1.0.0"
```
bdice, line 88: This configuration can be dropped, in favor of the "oldest" configuration.
```suggestion
```

.circleci/config.yml
```diff
@@ -77,18 +84,22 @@ jobs:
...
   linux-python-39-signac-13:
     <<: *test-template
     environment:
+      DEPENDENCIES: "NEWEST"
       SIGNAC_VERSION: "signac~=1.3.0"
   linux-python-39-signac-14:
     <<: *test-template
     environment:
+      DEPENDENCIES: "NEWEST"
       SIGNAC_VERSION: "signac~=1.4.0"
```
bdice, line 98: We aim to test the two most recent minor versions. Since we're updating this file anyway, we can change that now.
```suggestion
  linux-python-39-signac-15:
    <<: *test-template
    environment:
      DEPENDENCIES: "NEWEST"
      SIGNAC_VERSION: "signac~=1.5.0"
  linux-python-39-signac-14:
    <<: *test-template
    environment:
      DEPENDENCIES: "NEWEST"
      SIGNAC_VERSION: "signac~=1.4.0"
```

requirements/requirements.txt
```diff
@@ -0,0 +1,5 @@
+signac>=1.0.0
+jinja2>=2.10
+cloudpickle>=1.1.1
+deprecation>=2
```
bdice, line 4:
```suggestion
deprecation>=2.0.0
```

setup.py
```diff
@@ -9,13 +9,13 @@
     # The core package.
     "signac>=1.0.0",
     # For the templated generation of (submission) scripts.
-    "jinja2>=2.8",
+    "jinja2>=2.10",
     # To enable the parallelized execution of operations across processes.
-    "cloudpickle",
+    "cloudpickle>=1.1.1",
     # Deprecation management
     "deprecation>=2",
```
bdice, line 16:
```suggestion
    "deprecation>=2.0.0",
```

bdice (changes requested):
Nice work, @kidrahahjo! I have a few suggested changes. I'm going to work on applying those changes now.

bdice (approved):

PR #439, opened 2021-01-25
Remove `--test` option from `submit` CLI

### LINKED ISSUE #419: Remove --test option from "submit" CLI
Discussed on Slack with @vyasr @atravitz.
> I am not convinced that the behavior of `submit --test` is implemented in a way that is logically consistent. The option is only in the CLI, not the Python API. Its documentation reads "Do not interact with the scheduler, implies `--pretend`."

Summary of the conversation:
- It is unclear whether "not interacting with the scheduler" should rely on the project document's cached status from previous scheduler states.
- "the expected use case was for testing during development, possibly in environments where the scheduler was unknown/unconfigured/problematic. also at some point when schedulers were interacted with more directly to generate submission scripts (pre-jinja) i don’t remember what kind of output you’d get from the [FakeScheduler], i.e. even if you didn’t get errors you might get something that cluttered the output. given that context, it probably assumes that no cached status would exist, or just doesn’t consider that possibility at all" -- @vyasr 
- If users wish to skip interacting with the scheduler, it is possible to use the `TestEnvironment` (which uses the `FakeScheduler`).
- We do not expose any mechanism to perform "real" (not pretend) submission unless the scheduler state is checked first. This is important for safety/correctness.
- This feature is no longer useful and should be removed.

kidrahahjo:
## Description

Remove the `test` option from `submit` CLI. Users should now use `submit --pretend` instead.

## Motivation and Context

Fix #419 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -56,6 +56,7 @@ Removed
 - Removed the deprecated ``env`` argument from submission methods (#424).
 - ``flow.render_status.Renderer`` class has been removed. ``FlowProject.print_status`` no longer returns the renderer (#426).
 - Removed deprecated ``status.py`` module (#426).
+- Removed the ``--test`` argument from ``FlowProject.submit``. (#439).
 
 Version 0.11
 ============
--- a/changelog.txt
+++ b/changelog.txt
@@ -56,7 +56,7 @@ Removed
 - Removed the deprecated ``env`` argument from submission methods (#424).
 - ``flow.render_status.Renderer`` class has been removed. ``FlowProject.print_status`` no longer returns the renderer (#426).
 - Removed deprecated ``status.py`` module (#426).
-- Removed the ``--test`` argument from ``FlowProject.submit``. (#439).
+- Removed the ``--test`` argument from ``FlowProject.submit`` (#439).
 
 Version 0.11
 ============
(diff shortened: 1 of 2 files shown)


atravitz (approved):
lgtm!

bdice:
I would like to clarify the description/purpose of this PR: `--test` implies `--pretend` but they're not the same. The `--test` feature would determine eligibility without respecting the current scheduler status. This is discussed in more detail in #419.

changelog.txt
```diff
@@ -56,6 +56,7 @@ Removed
 - Removed the deprecated ``env`` argument from submission methods (#424).
 - ``flow.render_status.Renderer`` class has been removed. ``FlowProject.print_status`` no longer returns the renderer (#426).
 - Removed deprecated ``status.py`` module (#426).
+- Removed the ``--test`` argument from ``FlowProject.submit``. (#439).
```
bdice, line 59:
```suggestion
- Removed the ``--test`` argument from ``FlowProject.submit`` (#439).
```

bdice (approved):

PR #440, opened 2021-01-25
Require tqdm 4.48.1.

bdice:
## Description
While working with @tcmoore3, we found a bug in the `master` branch caused by an insufficently recent version of tqdm. We need tqdm>=4.48.1 to fix this, which uses a shallow copy instead of a deep copy of the tqdm keyword arguments.

This PR also ensures that the root-level `requirements.txt` is correct and up to date.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/.circleci/ci-oldest-reqs.txt
+++ b/.circleci/ci-oldest-reqs.txt
@@ -1,4 +1,4 @@
 cloudpickle==1.1.1
 deprecation==2.0.0
 jinja2==2.10
-tqdm==4.43.0
+tqdm==4.48.1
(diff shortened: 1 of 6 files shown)


kidrahahjo:
If we decide to keep package level requirements (`requirements.txt`) to root level, we should probably rename the `requirements` directory to `requirements-testing`.

bdice:
> If we decide to keep package level requirements (`requirements.txt`) to root level, we should probably rename the `requirements` directory to `requirements-testing`.

We could move them to the `.circleci` directory, too, but I don’t know if it really matters. The goals are:
1. Allow Dependabot to find and update all requirements files except the “oldest” CI configuration. (I’m not sure if dependabot will search the hidden `.circleci` folder by default.)
2. Use names that make sense to users and especially new developers.
3. Keep the root of the repo clean (fewer top-level items).

mikemhenry (approved):
LGTM!

PR #441, opened 2021-01-25
Feature/bridges2 environment

tcmoore3:
Add Bridges-2 environment + template.

## Description
Add an environment class and template for the XSEDE-funded Bridges-2 machine at PSC.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


tcmoore3:
Making this a draft for now while I:

1. Update the docs
2. Make sure my jobs are running properly with this template

jennyfothergill:
@tcmoore3 Thank you for working on this! I wanted to let you know I successfully used this PR (merged with master) with my workflow. :)

bdice:
I am starting some runs on Bridges2 and will finish up / review this PR.

bdice:
@tcmoore3 @jennyfothergill Have you used RM-shared with large numbers of jobs? I tried to submit 81 single-CPU jobs on RM-shared. I expected that it would work with up to 128 tasks on a node (2 CPUs, 64 cores per CPU). With `--partition=RM-shared, -N 1, --ntasks=81` I get a failure:
```
sbatch: error: QOSMaxCpuPerJobLimit
sbatch: error: Batch job submission failed: Job violates accounting/QOS policy (job submit limit, user's size and/or time limits)
```

However, when I limited it to 64 tasks, it ran fine. I was curious if you'd seen any related issues and what configurations you're actively using.

tcmoore3:
@bdice That's curious. I've had different issues on Bridges 2, but nothing like that.

flow/scheduling/base.py
```diff
@@ -159,10 +159,12 @@ def _call_submit(submit_cmd, script, pretend):
...
             submit_cmd.append(tmp_submit_script.name)
             try:
-                subprocess.check_output(submit_cmd, universal_newlines=True)
+                subprocess.check_output(
+                    submit_cmd, stderr=subprocess.STDOUT, universal_newlines=True
+                )
             except subprocess.CalledProcessError as error:
                 raise SubmitError(
-                    f"Error when calling submission command {submit_cmd_string}: {error.output()}"
+                    f"Error when calling submission command {submit_cmd_string}:\n{error.output}"
```
bdice, line 167: The changes in this file fix a bug that is unrelated to Bridges-2. The error output was incorrectly handled, and this had not been tested. (The bug has been in signac-flow for a long time, as far as I could tell.)

bdice (approved):
This template appears to be working for me, except for my one question about RM-shared with more than 64 tasks. I have emailed the PSC help desk and will follow up with whatever they tell me. In the meantime, I approve this PR as-is and will fix any issues in a follow-up PR.

bdice:
> RM-shared is intended for jobs that use <= half of the 128 RM node cores. If you need more than 64, please request the whole node.

PSC’s support staff got back to me. We could add a check that RM-shared jobs request 64 cores or fewer, but this is effectively already the case because SLURM rejects the submission. @glotzerlab/signac-committers How do we typically handle this kind of situation?

vyasr:
The one that immediately comes to mind is that Stampede2 (I think it was Stampede2) limits the total number of submitted jobs to 20 at any given time, and we just let that fail on submission rather than catching ahead of time. I'd be fine relying on that here as well. On the other hand we could preempt future questions by catching this error and returning a more readable one, saving us work in the long run, so I'd recommend at least documenting this within the environment.

tcmoore3:
You're right, it is Stampede2 that limits the number of jobs you can submit. And even though these are similar issues, this one in particular is something that we can easily catch pre-submission, whereas it's more involved (currently impossible?) to check a user's queue load before submitting the job. We should catch this and return a clear error messsage as @vyasr suggested, because this issue definitely will come up.

jennyfothergill (approved):
LGTM! I was able to submit my jobs to the GPU-shared partition. As I mentioned, I got an error when I run `submit` after setting the account: (`signac config --global set 'flow.Bridges2Environment.account' myaccountname`). Has anyone else encountered this? I'm not sure if it's a signac or Bridges issue.

bdice:
@jennyfothergill That’s a warning/suggestion, not an error. Some schedulers require an account name but Bridges-2 will automatically select it for you (all users have a default account affiliated with their project).

jennyfothergill:
@bdice Ah, sorry--I don't mean the suggestion. I mean that after I set my account name, I get an error and I can't submit jobs.

flow/environments/xsede.py
```diff
@@ -206,11 +206,9 @@ def add_args(cls, parser):
...
                 "RM-shared",
                 "RM-small",
-                "LM",
+                "EM",
                 "GPU",
                 "GPU-shared",
-                "GPU-small",
-                "GPU-AI",
             ],
             default="RM-shared",
```
bdice, line 213: Should the default partition be `RM` or `RM-shared`? @jennyfothergill @tcmoore3

bdice:
I'd like to merge this today after figuring out the issue @jennyfothergill mentioned (I messaged her separately to discuss the problem, since it involves account information).

flow/environments/xsede.py
jennyfothergill: I only have allocations on GPU partitions... 😅

jennyfothergill:
@bdice yes, thank you! (I was confused between my account name and the the grant account name. oops!) 😅 

`signac config --global set 'flow.Bridges2Environment.account' myaccountname` In case anyone else has this confusion, I wonder if there could be a way to make the distinction more clear in this message?

flow/environments/xsede.py
bdice: @jennyfothergill Good to know, I hadn’t considered that possibility. I suppose I’ll just leave RM-shared as the default since it is less likely that someone would accidentally use a large number of hours on a shared queue. (RM reserves full nodes.)

bdice:
@jennyfothergill I see the confusion -- I updated the message text and hopefully this helps future users.

tcmoore3:
I’m a little late to the party on this one, but I agree RM-shared is the better default partition. Those 128-core nodes burn up hours pretty quickly, so better safe than sorry here.

Issue #291, opened 2020-06-02
Warning for requesting to run non-existent flow operation


rsdefever:
### Feature description

IMO it would be beneficial to throw a warning if a user requests to run a non-existent flow operation. Example, the flow operation is named `run_simulation` and the user types: `python project.py run -o run_simulations`. This would not execute the flow operation (it has a typo; `run_simulations` does not exist), and it would also not raise any warning or provide any other indication that nothing happened. 

### Proposed solution

A simple warning such as `The requested flow operation does not exist`.

csadorf:
@rsdefever Thank you for the suggestion!

bdice:
Resolved by #442.

PR #442, opened 2021-01-26
Display message for non existent operations/groups while execution

kidrahahjo:
## Description

Users will now be shown a message _**The requested flow operation(s) does not exist: names**_ if they pass a non-existent operation.
@glotzerlab/signac-committers should we add tests for this? :thinking: 

## Motivation and Context

Thanks to @rsdefever for suggesting this feature.
Addresses #291 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


flow/project.py
```diff
@@ -3199,6 +3199,12 @@ def run(
             )
         if names is None:
             names = list(self.operations)
+        else:
+            absent_ops = (set(self._groups.keys()) ^ set(names)) & set(names)
+            if absent_ops:
+                print(
+                    f"The requested flow operation(s) does not exist: {', '.join(absent_ops)}"
+                )
```
bdice, line 3207: We can't perform this check yet. At this point, the `names` are regular expressions that can match groups. There's a line below here (3287) where these regular expressions are converted to a list of matching operations.

```python
run_groups = set(self._gather_flow_groups(names))
```

What we probably need to do is check below and show a warning if the provided expressions match zero group names. Try this with operations named `op1`, `op2` and an expression like `project.py run -o op`.

bdice: Oh wait, sorry. I see that `_gather_flow_groups` was also edited. If that's correct, then why do we need the check here?

bdice: Could we just edit `_gather_flow_groups` instead?

bdice:
This should be merged after releasing v0.12 (we're in feature freeze for that release ❄️).

flow/project.py
kidrahahjo: @bdice We do a `re.fullmatch` for the names. This means performing either the way I did or moving the code to `_gather_flow_groups` would do the same thing. The reason I was not willing to move it to `_gather_flow_groups` was that since we iterate over line [`3287`](https://github.com/glotzerlab/signac-flow/blob/feature/warnings-for-non-existent-op/flow/project.py#L3287) multiple times in `run`, we'd then end up printing the message multiple times.

changelog.txt
```diff
@@ -68,6 +68,7 @@ Fixed
 - Ensure that directives are always evaluated before running or submitting (#408, #409).
 - Cache the fully qualified domain name during environment detection to fix a performance issue on macOS (#339, #394).
 - Ensure that ``next`` CLI command displays eligible jobs for the exact operation name provided (#443).
+- Users will now know if they passed a non-existing operation/group to ``run``, ``submit``, or ``next`` (#291, #442).
```
csadorf, line 71:
```suggestion
- Display warning when a non-existing operation/group is passed to ``run``, ``submit``, or ``next`` commands. (#291, #442).
```

flow/project.py
```diff
@@ -3045,6 +3045,13 @@ def run(
             )
         if names is None:
             names = list(self.operations)
+        else:
+            absent_ops = (set(self._groups.keys()) ^ set(names)) & set(names)
+            if absent_ops:
+                print(
+                    f"The requested flow operation(s) does not exist: {', '.join(absent_ops)}",
```
csadorf, line 3052:
```suggestion
                print(
                    f"Unrecognized flow operation(s): {', '.join(absent_ops)}",
```

flow/project.py
```diff
@@ -3576,6 +3583,13 @@ def submit(
             Additional keyword arguments forwarded to :meth:`~.ComputeEnvironment.submit`.
 
         """
+        if names is not None:
+            absent_ops = (set(self._groups.keys()) ^ set(names)) & set(names)
+            if absent_ops:
+                print(
+                    f"The requested flow operation(s) does not exist: {', '.join(absent_ops)}"
```
csadorf, line 3590:
```suggestion
                    f"Unrecognized flow operation(s): {', '.join(absent_ops)}"
```

csadorf (changes requested):
I'm suggesting a slight change in wording.

flow/project.py
```suggestion
                    f"Unrecognized flow operation(s): {', '.join(absent_ops)}"
```

tests/test_project.py
```diff
@@ -1010,6 +1010,23 @@ def test_run_fork(self):
...
+        def op1(job):
+            pass
+
+        project = self.mock_project(A)
+        output = StringIO()
+        with redirect_stderr(output):
+            project.run(names=["op1", "op2", "op3"])
+        output.seek(0)
+        message = output.read()
+        fail_msg = "The requested flow operation(s) does not exist:"
```
bdice, line 1027:
```suggestion
        fail_msg = "Unrecognized flow operation(s):"
```

tests/test_project.py
```diff
@@ -1247,6 +1262,18 @@ def test_main_run(self):
...
 
+    def test_main_run_invalid_op(self):
+        assert len(self.project)
+        run_output = " ".join(
+            self.call_subcmd("run -o invalid_op_run", subprocess.STDOUT)
+            .decode("utf-8")
+            .split()
+        )
+        assert (
+            "The requested flow operation(s) does not exist: invalid_op_run"
```
bdice, line 1273:
```suggestion
            "Unrecognized flow operation(s): invalid_op_run"
```

changelog.txt
```diff
@@ -68,6 +68,7 @@ Fixed
 - Ensure that directives are always evaluated before running or submitting (#408, #409).
 - Cache the fully qualified domain name during environment detection to fix a performance issue on macOS (#339, #394).
 - Ensure that ``next`` CLI command displays eligible jobs for the exact operation name provided (#443).
+- Display warning when a non-existing operation/group is passed to ``run``, ``submit``, or ``next`` commands. (#291, #442).
```
bdice, line 71:
```suggestion
- Display warning when a non-existing operation/group is passed to ``run``, ``submit``, or ``next`` commands (#291, #442).
```

bdice (approved):
Nice work @kidrahahjo, I applied @csadorf's suggested changes and I approve if tests pass.

flow/project.py
```diff
@@ -3576,6 +3583,13 @@ def submit(
             Additional keyword arguments forwarded to :meth:`~.ComputeEnvironment.submit`.
 
         """
+        if names is not None:
+            absent_ops = (set(self._groups.keys()) ^ set(names)) & set(names)
+            if absent_ops:
+                print(
+                    f"Unrecognized flow operation(s): {', '.join(absent_ops)}"
+                )
```
bdice, line 3591:
```suggestion
                print(f"Unrecognized flow operation(s): {', '.join(absent_ops)}")
```

flow/project.py
```diff
@@ -3576,6 +3583,13 @@ def submit(
             Additional keyword arguments forwarded to :meth:`~.ComputeEnvironment.submit`.
 
         """
+        if names is not None:
+            absent_ops = (set(self._groups.keys()) ^ set(names)) & set(names)
+            if absent_ops:
+                print(
```
bdice: Should this go to stderr?

tests/test_project.py
```diff
@@ -1247,6 +1262,18 @@ def test_main_run(self):
...
+        assert len(self.project)
+        run_output = " ".join(
+            self.call_subcmd("run -o invalid_op_run", subprocess.STDOUT)
+            .decode("utf-8")
+            .split()
+        )
+        assert (
+            "Unrecognized flow operation(s): invalid_op_run"
+            in run_output
+        )
```
bdice, line 1275:
```suggestion
        assert "Unrecognized flow operation(s): invalid_op_run" in run_output
```

bdice (changes requested):
I realized I have one question to address regarding stdout/stderr.

flow/project.py
bdice: I only noticed this inconsistency because the error message proposed by @csadorf is shorter, and thus triggers a reformat on this line which caused pre-checks to fail. Without specifying `file=sys.stderr`, the auto-formatting looks different for this line than the other messages. 😅

kidrahahjo: The reason for choosing `sys.stderr` was partly because some of the warning statements (when printing them) were redirected to the error stream.

flow/project.py
```diff
@@ -3576,6 +3583,11 @@ def submit(
             Additional keyword arguments forwarded to :meth:`~.ComputeEnvironment.submit`.
 
         """
+        if names is not None:
+            absent_ops = (set(self._groups.keys()) ^ set(names)) & set(names)
+            if absent_ops:
+                print(f"Unrecognized flow operation(s): {', '.join(absent_ops)}")
```
bdice, line 3589:
```suggestion
                print(
                    f"Unrecognized flow operation(s): {', '.join(absent_ops)}",
                    file=sys.stderr,
                )
```

bdice (approved):

flow/project.py
bdice: @kidrahahjo I'm not sure if we were talking about the same thing. I fixed the thing I meant in ecf0d2b.

PR #443, opened 2021-01-26
Fix next command-line-argument comparison method

kidrahahjo:
## Description

In the commit https://github.com/glotzerlab/signac-flow/pull/324/commits/9a7ca7e18bfb338f96298cc1f6f78c08232be79f, the operation comparison behaviour was changed which resulted in the display of eligible jobs for every operation having user-specified operation name as its substring.

For example, if an operation `op` is present in a `FlowProject`, then executing `python project.py next o` would result in displaying of eligible jobs for operation `op`.

## Motivation and Context

Link to slack thread: https://signac.slack.com/archives/CPCMN12EL/p1611687106012300

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -17,7 +17,7 @@ Added
 - Code is formatted with ``black`` and ``isort`` pre-commit hooks (#365).
 - Add official support for Python version 3.9 (#365).
 - Documentation has been added for all public classes and methods (#387, #389).
-- Added internal support for aggregates of jobs (#334, #348, #351, #364, #383, #390, #415, #422, #430).
+- Added internal support for aggregates of jobs (#334, #348, #351, #364, #383, #390, #415, #422, #430, #443).
 - Added code coverage to continuous integration (#405).
 
 Changed
@@ -47,7 +47,6 @@ Fixed
 
 - Ensure that directives are always evaluated before running or submitting (#408, #409).
 - Cache the fully qualified domain name during environment detection to fix a performance issue on macOS (#339, #394).
-- Ensure that ``next`` command line argument does not display eligible jobs for unequal operations (#443).
 
 Removed
 +++++++
--- a/changelog.txt
+++ b/changelog.txt
@@ -47,7 +47,7 @@ Fixed
 
 - Ensure that directives are always evaluated before running or submitting (#408, #409).
 - Cache the fully qualified domain name during environment detection to fix a performance issue on macOS (#339, #394).
-- Ensure that ``next`` command line argument does not display eligible jobs for unequal operations (#443).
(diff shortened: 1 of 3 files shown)


changelog.txt
```diff
@@ -47,6 +47,7 @@ Fixed
 
 - Ensure that directives are always evaluated before running or submitting (#408, #409).
 - Cache the fully qualified domain name during environment detection to fix a performance issue on macOS (#339, #394).
+- Ensure that ``next`` command line argument does not display eligible jobs for unequal operations (#443).
```
bdice, line 50: In my understanding, this bug was never released, so this does not need to be marked as a "fix." I would just put this number in the list above with all the other aggregation-related PRs, which is how I've handled some similar PRs:
```
- Added internal support for aggregates of jobs (#334, #348, #351, #364, #383, #390, #415, #422, #430, #443).
```

bdice (approved):
Good fix. I have one suggestion for the changelog.

bdice:
This should be merged **before** releasing v0.12 (it fixes a known bug that would otherwise get released).

changelog.txt
kidrahahjo: I think this bug was released in `0.11.0`.

kidrahahjo: Here is the link to the code in the branch `release/0.11.0`: https://github.com/glotzerlab/signac-flow/blob/release/0.11.0/flow/project.py#L3659

tests/test_project.py
```diff
@@ -1250,8 +1250,12 @@ def test_main_run(self):
     def test_main_next(self):
         assert len(self.project)
         job_ids = set(self.call_subcmd("next op1").decode().split())
+        assert len(job_ids) > 0
         even_jobs = [job.get_id() for job in self.project if job.sp.b % 2 == 0]
         assert job_ids == set(even_jobs)
+        # Use only exact operation matches
+        job_ids = set(self.call_subcmd("next op").decode().split())
+        assert len(job_ids) == 0
```
kidrahahjo, line 1258: @bdice I am in favour of this test. I wrote it myself but chose not to push it because it is kind-of misleading (It may be possible that `op` is a valid operation and its just the case that there are no eligible jobs). I think we might want to add a test which would match the exact output thrown by the `next` command-line flag after #442 is merged.

bdice: > (It may be possible that op is a valid operation and its just the case that there are no eligible jobs).
This is only going to run on the test project that we define -- if someone adds an operation named `op` then this test might fail, but otherwise this is a good test.

PR #445, opened 2021-01-31
Remove deprecated features for release 0.13

bdice:
## Description
Removes deprecated code marked for removal in 0.13.0.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


flow/project.py
```diff
@@ -687,6 +610,9 @@ def __call__(self, *jobs, **kwargs):
             raise ValueError(f"Invalid keyword arguments: {', '.join(kwargs)}")
 
         if job is not None:
+            # TODO 0.13: How to remove this keyword argument? How should
+            # FlowCmdOperation act if multiple jobs are provided? Need to
+            # update both documentation and behavior.
```
bdice, line 615: I'd like input on this, @glotzerlab/signac-committers. I don't know how to proceed.

bdice: Maybe we change the end of the function to:
```python
cmd = self._cmd(*jobs) if callable(self._cmd) else self._cmd
format_arguments = {"jobs": jobs}
if len(jobs) == 1:
    format_arguments["job"] = jobs[0]
return cmd.format(**format_arguments)
```

kidrahahjo: Can we advise users to not use the "job" variable in the command? Using only curly braces would make things simple, right?

bdice: We should not break existing cmd operations. Typically a user provides a string like `my_executable --input {job.ws}/file.txt`. In that case we need to keep `job` as a format argument.

klywang: What would be an example use for this?

> Maybe we change the end of the function to:
> 
> ```python
> cmd = self._cmd(*jobs) if callable(self._cmd) else self._cmd
> format_arguments = {"jobs": jobs}
> if len(jobs) == 1:
>     format_arguments["job"] = jobs[0]
> return cmd.format(**format_arguments)
> ```

Is this assuming that, if someone provides multiple jobs, they would expect it to execute as something like `my_executable --input job1 job2 job3 --another_flag`? 

And would the alternative expected behavior is to have `my_executable --input {} --another_flag` (or whatever the provided shell command was) be executed for each job?

vyasr: @klywang A shell command with multiple jobs would be used in the same way that a Python operation with multiple operations would be used. For example, if you want to make a combined plot of all jobs where the statepoint key `a = 1`, the new aggregation feature would let you write a Python operation `plot(jobs)` that takes all the desired jobs and makes a plot. Now if for some reason you need to use a non-Python tool to make a specific kind of plot, you could use a shell command operation that runs `OTHER_TOOL --files job1_fn job2_fn ...`.

@bdice I agree, we should not break existing code. I think your proposal makes sense, we should handle this in as close to the same manner that we handle Python operations as possible. Python operations are called as `op(*jobs)`, which works in both the single-job and many-job contexts. The only difference between what you proposed and an exact adaptation to the Python code is that the format arguments still contains the tuple, not an unpacked set of jobs, but in a format string context that kind of unpacking doesn't make much sense, so passing the tuple in addition to aliasing the single-job case seems right to me.

bdice: Great. I updated this code.

kidrahahjo (approved):
This looks good to me. Also, It's worth writing tests for the cmd operations when used with aggregation. I will write those tests in the branch that I created for aggregation and push the code.

tcmoore3 (approved):
LGTM

Issue #457, opened 2021-02-08
Add `before_operations` block to `base_script.sh`


berceanu:
I propose adding an empty block, `before_operations`, right after the `for` loop. 

https://github.com/glotzerlab/signac-flow/blob/dbe89263b002df8389e5dfaab02b81d5fec253e8/flow/templates/base_script.sh#L18-L20

Use case: `export CUDA_VISIBLE_DEVICES`, see https://github.com/glotzerlab/signac-flow/issues/455.

vyasr:
Resolved in #458.

Issue #437, opened 2021-01-22
Stampede 2  bundeled submissions


DomFijan:
### Description

Submission of bundles on stampede2 produces wrong submission script. Might be a template issue. 

### To reproduce

` python3 project.py submit -n 3 -b 3 -w 0.2  --parallel --force --pretend `

### Error output

On stampede2 bundeling is done thorugh the `iburn` and if bundles are submitted an offest must be provided with -o. The old  version of signac-flow (0.8) produced the correct command :
```
# equilibrate(662e116bc54baa4fbc7a00991e45c6fa)
ibrun -n 16 -o 0 task_affinity singularity exec software.simg python3 project.py exec equilibrate 662e116bc54baa4fbc7a00991e45c6fa &
# equilibrate(92dec1dd2f231126cd7aa1c57d7b324c)
ibrun -n 16 -o 16 task_affinity singularity exec software.simg python3 project.py exec equilibrate 92dec1dd2f231126cd7aa1c57d7b324c &
# equilibrate(159dee0e06c1defdd968cb473a786585)
ibrun -n 16 -o 32 task_affinity singularity exec software.simg python3 project.py exec equilibrate 159dee0e06c1defdd968cb473a786585 &
```
while the new version produces (pulled from master on 01/22/2021):
```
# equilibrate(52d99c004c293079cbc490ce1857f271)
_FLOW_STAMPEDE_OFFSET_=0 /opt/apps/intel18/python3/3.7.0/bin/python3 /scratch/07885/dfijan/TRIANGLES0_25/project.py run -o equilibrate -j 52d99c004c293079cbc490ce1857f271 &
# Eligible to run:
# ibrun -n 16 -o 0 task_affinity  singularity exec software.simg python3 /scratch/07885/dfijan/TRIANGLES0_25/project.py exec equilibrate 52d99c004c293079cbc490ce1857f271
# equilibrate(2a5a7766066be122f7c47d02df18981f)
_FLOW_STAMPEDE_OFFSET_=16 /opt/apps/intel18/python3/3.7.0/bin/python3 /scratch/07885/dfijan/TRIANGLES0_25/project.py run -o equilibrate -j 2a5a7766066be122f7c47d02df18981f &
# Eligible to run:
# ibrun -n 16 -o 0 task_affinity  singularity exec software.simg python3 /scratch/07885/dfijan/TRIANGLES0_25/project.py exec equilibrate 2a5a7766066be122f7c47d02df18981f
[...]

PR #458, opened 2021-02-08
Add empty jinja block

berceanu:
Address https://github.com/glotzerlab/signac-flow/issues/457.


tommy-waltmann (approved):

vyasr (changes requested):
@berceanu can you also modify `stampede2.sh`? You'll notice that it completely overwrites the body block, and the new body block has two separate branches based on a variable `use_launcher`. The branch corresponding to `use_launcher == False` should look basically like the body block from `base_script.sh`, and you can copy-paste this solution in there as well.

csadorf (commented):
This might be the academic within me, but can we maybe name the block `pre_operation`?

flow/templates/base_script.sh
```diff
@@ -18,6 +18,8 @@ cd {{ project.config.project_dir }}
 {% for operation in operations %}
 
 # {{ "%s"|format(operation) }}
+{% block pre_operation %}
```
bdice, line 21: This is missing the `operation` variable from the upper scope when it is overridden in `stampede2.sh`. Defining this as a `scoped` block should help.
https://jinja.palletsprojects.com/en/2.11.x/templates/#block-nesting-and-scope
```suggestion
{% block pre_operation scoped %}
```

flow/templates/stampede2.sh
```diff
@@ -55,33 +54,11 @@ export LAUNCHER_JOB_FILE={{ launcher_file }}
...
-{% endfor %}
-{% endif %}
-{% if operation.operations_with_met_postconditions|length > 0 %}
-# Operations with all postconditions met:
-{% for run_op in operation.operations_with_met_postconditions %}
-# {{ run_op.cmd }}
-{% endfor %}
-{% endif %}
-{% endfor %}
+{% block pre_operation %}
```
bdice, line 57:
```suggestion
{# Only the pre_operation block is overridden, all other behavior is inherited from base_script.sh #}
{% block pre_operation %}
```

bdice: But I'm not even sure this will work -- the code you're replacing has the variable defined on the same line as the operation command.

Before:
```bash
_FLOW_STAMPEDE_OFFSET_=8 ibrun ... stuff
```

After:
```bash
_FLOW_STAMPEDE_OFFSET_=8
ibrun ... stuff
```

@vyasr can you confirm this and offer insight on how we should proceed, since you suggested the change to Stampede2? Do we add an export like `export _FLOW_STAMPEDE_OFFSET_=8` or do we just abandon this idea entirely and copy-paste the whole template as before?

vyasr: @bdice you are correct that the variable scope will not work with the intervening newline, and that adding an export should fix it. That is probably the best solution since it also has the benefit of significantly reducing the duplication here.

flow/templates/stampede2.sh
```diff
@@ -55,33 +54,12 @@ export LAUNCHER_JOB_FILE={{ launcher_file }}
...
-{% if operation.operations_with_met_postconditions|length > 0 %}
-# Operations with all postconditions met:
-{% for run_op in operation.operations_with_met_postconditions %}
-# {{ run_op.cmd }}
-{% endfor %}
-{% endif %}
-{% endfor %}
+{# Only the pre_operation block is overridden, all other behavior is inherited from base_script.sh #}
+{% block pre_operation %}
+{{ "export _FLOW_STAMPEDE_OFFSET_=%d "|format(operation.directives['nranks']|return_and_increment) }}
```
bdice, line 59:
```suggestion
export _FLOW_STAMPEDE_OFFSET_={{ "%d"|format(operation.directives['nranks']|return_and_increment) }}
```

flow/templates/stampede2.sh
```diff
@@ -35,7 +35,6 @@
 {% endif %}
 {% endblock %}
 
-{% block body %}
```
bdice, line 38: I _think_ we still need the `body` block?

bdice (commented):
I'm going to attempt to test this PR (generating a sample script for Stampede2 and checking to see if it makes sense -- probably not going to run it).

flow/templates/stampede2.sh
vyasr: Yes, you should still have that. Right now all of the logic for launcher is outside all blocks, which I think means that it won't render at all.

vyasr:
@bdice thanks for testing on Stampede2, that was going to be my main outstanding request. Once you are convinced that tests look ok (perhaps @b-butler can also take a look since he has looked at Stampede2 scripts before) @berceanu will need to regenerate the template testing scripts. See [here](https://docs.signac.io/projects/flow/en/latest/support.html#contributing-environments-to-the-package) for how to do that, but please hold off committing new versions of the testing output until @bdice confirms that it looks correct.

bdice:
I tried a few different things here but did not succeed and don't have time to do any more at the moment. In the variations I tried, either the offsets were wrong or the template failed to render entirely. We might need to just revert the changes to the Stampede2 template so that it retains the previous behavior. I really don't like that machine, it presents so many problems for signac-flow...

vyasr:
@bdice I think I've resolved the problem, aside from the known issue of pretend outputs not incrementing the counter correctly (see [here](https://github.com/glotzerlab/signac-flow/issues/437#issuecomment-767002432) for a refresher on that). Can you test on stampede again when you get a chance, and see if you're still getting output that you don't think makes sense? There's no way to validate the offset itself without actually submitting, but if you run the template tests and you see that the offset is always 0 in both cases (what I'd expect) then the output seems consistent.

If that seems right to you, then the reference output needs to be regenerated and this PR should be good to go.

bdice (approved):
@vyasr your changes appear to work for me. Thanks! I did not test on the actual Stampede2 machine but I verified that the generated templates all say `export _FLOW_STAMPEDE_OFFSET_=0`.

bdice:
Ugh, I committed with a custom setting in my `$HOME/.signac.rc`. Let me fix that.

Issue #456, opened 2021-02-08
Create new block in `slurm.sh`


berceanu:
I propose grouping these lines into a new block, nested inside the `header` block.

https://github.com/glotzerlab/signac-flow/blob/dbe89263b002df8389e5dfaab02b81d5fec253e8/flow/templates/slurm.sh#L3-L17

Use-case: allows inheriting these lines and changing just the `tasks` block.

vyasr:
Resolved in #459.

PR #459, opened 2021-02-08
Add nested block preamble

berceanu:
Addresses https://github.com/glotzerlab/signac-flow/issues/456


bdice:
@berceanu Can you comment with examples of what the user template looks like before and after this change? I'm not 100% sure if I understand what piece of the user template this is simplifying. I thought you could just call `super()` but maybe there's boilerplate I'm not seeing.

berceanu:
Well, this
```
{% block header %}
#!/bin/bash
#SBATCH --job-name="{{ id }}"
{% if partition %}
#SBATCH --partition={{ partition }}
{% endif %}
{% if memory %}
#SBATCH --mem={{ memory }}
{% endif %}
{% if walltime %}
#SBATCH -t {{ walltime|format_timedelta }}
{% endif %}
{% if job_output %}
#SBATCH --output={{ job_output }}
#SBATCH --error={{ job_output }}
{% endif %}

{% block tasks %}
{% set threshold = 0 if force else 0.9 %}
{% set cpu_tasks = operations|calc_tasks('np', parallel, force) %}
{% set gpu_tasks = operations|calc_tasks('ngpu', parallel, force) %}
{% if gpu_tasks and 'gpu' not in partition and not force %}
{% raise "Requesting GPUs requires a gpu partition!" %}
{% endif %}
{% set nn_cpu = cpu_tasks|calc_num_nodes(96) if 'gpu' not in partition else cpu_tasks|calc_num_nodes(96) %}
{% set nn_gpu = gpu_tasks|calc_num_nodes(16) if 'gpu' in partition else 0 %}
{% set nn = nn|default((nn_cpu, nn_gpu)|max, true) %}
{% if 'gpu' in partition %}
#SBATCH --nodes={{ nn|default(1, true) }}
#SBATCH --ntasks-per-node={{ (gpu_tasks, cpu_tasks)|max }}
#SBATCH --gres=gpu:{{ gpu_tasks }}
{% else %}
#SBATCH --nodes={{ nn }}
#SBATCH --ntasks-per-node={{ (96, cpu_tasks)|min }}
{% endif %}
{% endblock tasks %}
{% endblock header %}

{% block project_header %}
```

would simplify to 
```
{% block header %}
{% block preamble %}
{{ super() }}
{% endblock %}

{% block tasks %}
{% set threshold = 0 if force else 0.9 %}
{% set cpu_tasks = operations|calc_tasks('np', parallel, force) %}
{% set gpu_tasks = operations|calc_tasks('ngpu', parallel, force) %}
{% if gpu_tasks and 'gpu' not in partition and not force %}
{% raise "Requesting GPUs requires a gpu partition!" %}
{% endif %}
{% set nn_cpu = cpu_tasks|calc_num_nodes(96) if 'gpu' not in partition else cpu_tasks|calc_num_nodes(96) %}
{% set nn_gpu = gpu_tasks|calc_num_nodes(16) if 'gpu' in partition else 0 %}
{% set nn = nn|default((nn_cpu, nn_gpu)|max, true) %}
{% if 'gpu' in partition %}
#SBATCH --nodes={{ nn|default(1, true) }}
#SBATCH --ntasks-per-node={{ (gpu_tasks, cpu_tasks)|max }}
#SBATCH --gres=gpu:{{ gpu_tasks }}
{% else %}
#SBATCH --nodes={{ nn }}
#SBATCH --ntasks-per-node={{ (96, cpu_tasks)|min }}
{% endif %}
{% endblock tasks %}
{% endblock header %}

{% block project_header %}
```

bdice:
@berceanu Are you able to overwrite *only* the tasks block using the existing structure?

See:
- https://jinja.palletsprojects.com/en/2.11.x/templates.html
- https://stackoverflow.com/a/5438968

berceanu:
You mean without using `super()` at all?

bdice:
> You mean without using `super()` at all?

Right. I wanted to see if you could override only the inner block. I’m not sure but it may be possible.

berceanu:
Nope, the lines from above are not added if I use

```
{% block header %}
{% block tasks %}
{% set threshold = 0 if force else 0.9 %}
{% set cpu_tasks = operations|calc_tasks('np', parallel, force) %}
...
```

bdice:
@berceanu What if you only define the inner scope? Don’t do anything to the header block. Just define “tasks” in the top level scope and see what happens.

```
{% block tasks %}
{% set threshold = 0 if force else 0.9 %}
{% set cpu_tasks = operations|calc_tasks('np', parallel, force) %}
...
{% endblock %}
```

berceanu:
Bingo, that seems to work!

vyasr:
If the only goal is to override tasks, then as you've found, you can do it (this is why the nested tasks block was added). I still think it's sensible to make a change like this to support the opposite use case, which is leaving the tasks block as is while modifying only the new preamble block. I may have thought through all of this without articulating it at all in our Slack thread when I said I supported adding an extra block here...

berceanu:
Yes, very good point. So how do we proceed?

vyasr:
I'm not a reviewer on this PR. Assuming @bdice and @Tobias-Dwyer are OK with this change, I think we can move forward with this PR as is.

bdice (changes requested):
A few comments:
- The original motivation for this PR can be solved another way (only overriding the `tasks` block as I suggested). However, there may be some small benefits for having more blocks instead of fewer.
- Adding a preamble block to _only_ the SLURM template is probably not the right approach. We should put this in `base_script.sh`. Put it... `{% block header %}HERE{% endblock %}`, so it will be an empty block inside an empty block.
- I like that the `endblock` is named with `preamble` but that doesn't match our other template blocks. I would like to add names to _all_ `endblock` tags for readability but that can be done in a separate issue: I created #461.

bdice (approved):
LGTM. Merging.

Issue #435, opened 2021-01-21
Add tests for print_status output


bdice:
### Feature description

The package currently lacks tests for the output of `FlowProject.print_status`.

### Proposed solution

It would be good to add regression tests that compare to a saved output, with various options like `--detailed`, `--pretty`, `--stack`, etc.

kidrahahjo:
We'd also have to write a script which automatically generates status output from a workflow and saves it. Functionality would be similar to `generate_template_reference_data.py`

vyasr:
Here's a partial listing copied from [this comment](https://github.com/glotzerlab/signac-flow/pull/460#issuecomment-778822640).

All the options like `--detailed`, `--stack`, `-1`, `--pretty`, `--full`, `--no-overview`, `--parameters`, etc. must be tested (regression tests like submission scripts).
Here is a list of tests: 
- [ ] Test status output when only labels are present
- [ ] Test status output when using **only** aggregation in flow project.
- [ ] Test status output when using aggregation with non-aggregate operations in flow project.
- [ ] Test status with parameters when aggregation is used.
- [ ] Once labels for aggregation is implemented, test labels for aggregates.

bdice:
We discussed this in office hours today with @Charlottez112 @atravitz @b-butler. Here's a rough sketch of some code that can help. We are going to have two projects: one that contains state points like `{"overview": True, "detailed": False}` to control options to `FlowProject.print_status`, and another project that contains the jobs that are actually in the status output.

Here are some notes for the proposed code:

1. Make a temporary directory for the "status options project."
2. Initialize status options project with statepoints like:

```
    overview: True/False
    detailed: True/False
    ---------------------
    EXTRA CREDIT
    parameters: None/['i']
    expand: True/False
    all_ops: True/False
    only_incomplete: True/False
    dump_json: True/False
    unroll: True/False
    compact: True/False
```

3. Create a "define_status_test_project.py" (see below) with simple operations, conditions, and labels (groups, aggregates, ...). The conditions and labels should be a function of the job state points. The conditions should cover common cases (no eligible operations, one eligible operation, multiple eligible operations). Similar for labels.


Here's a snippet that might belong in a file named `generate_status_reference_data.py`:

```python
from define_status_test_project import StatusTestProject

def init_status_project(project):
    """Generate state points and return a FlowProject used for status tests."""
    for i in range(10):
        project.open_job({"i": i}).init()

    return StatusTestProject.get_project(project.root_directory())

[...]

PR #460, opened 2021-02-09
Fix labels and job count for projects with zero operations.

### LINKED ISSUE #454: Flow status returns no jobs when no operations
### Description

For a project with initialized jobs but no defined operations, `flow.print_status()` returns that there are no jobs.

### To reproduce

Create an empty flow project and initialize jobs, then run `print_status()`.

### Error output

```

Overview:

Total # of jobs: 0

label
-------


operation
-----------

```

### System configuration

 - Version of signac-flow: 0.12

bdice:
## Description
The new status fetching code did not correctly handle FlowProjects with zero operations. This resolves #454.

I don't really want to write tests for status checks in this PR but it should be done at some point.

I tested this manually by creating a project with no operations and with/without labels. I tested `python project.py status` with and without a bunch of flags, and it seems to be fine now.

## Types of Changes
- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -2630,6 +2630,7 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
         # Add labels to the status information
         for job_label_data in job_labels:
             job_id = job_label_data["job_id"]
+            # If no status information existed, we need to set default values
             statuses.setdefault(job_id, {})
             statuses[job_id].setdefault("aggregate_id", job_id)
             statuses[job_id]["labels"] = job_label_data["labels"]
--- a/flow/project.py
+++ b/flow/project.py
@@ -2336,20 +2336,24 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
                     selected_groups=single_operation_groups,
                 )
             )
-            # aggregate_groups is a list of tuples containing scheduler,
-            # aggregate, and group information. To compute labels, we fetch the
-            # unique jobs from the aggregates containing only one job.
-            individual_jobs = {
-                aggregate_group[3][0]
-                for aggregate_group in aggregate_groups
-                if len(aggregate_group[3]) == 1
-            }
             status_results = parallel_executor(
                 compute_status,
                 aggregate_groups,
                 desc="Fetching status",
                 file=err,
             )
+            # aggregate_groups is a list of tuples containing scheduler,
+            # aggregate, and group information. To compute labels, we fetch the
+            # unique jobs from the aggregates containing only one job.
+            if len(single_operation_groups) > 0:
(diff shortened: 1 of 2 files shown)


kidrahahjo (approved):
This is an interesting find. I think we might need to do regression tests for status checks. I did not write the status tests for aggregation in #464 because we'll have to rewrite most of the status checks if we go for regression testing. For now, we can merge this and handle tests as part of #435 in a separate PR.

flow/project.py
```diff
@@ -2336,20 +2336,24 @@ def compute_status(data):
...
+            # unique jobs from the aggregates containing only one job.
+            if len(single_operation_groups) > 0:
+                individual_jobs = {
+                    aggregate_group[3][0]
+                    for aggregate_group in aggregate_groups
+                    if len(aggregate_group[3]) == 1
+                }
+            else:
+                # If no operations exist, use all jobs in the project.
+                individual_jobs = set(self)
```
vyasr, line 2356: Why did the location of this block move? This change doesn't seem necessary.

vyasr (changes requested):
This PR looks fine to me (just one question, but it's not holding anything back). I didn't approve yet, because I would like us to have a record somewhere of what tests are known to be missing and maybe start making a plan for them. In the push to get the last release ready we made a lot of changes, and while we actually improved coverage in some cases, we almost certainly degraded it in others. Can you open an issue we can use as a checklist for these kinds of things that we find as aggregation gets merged and we continue refactoring? And @kidrahahjo can you populate it with other cases that you're aware need to be tested? That would be very helpful, and I'll approve this PR once that's done.

kidrahahjo:
All the options like `--detailed`, `--stack`, `-1`, `--pretty`, `--full`, `--no-overview`, `--parameters`, etc. must be tested (regression tests like submission scripts).
Here is a list of tests: 
- [ ] Test status output when only labels are present
- [ ] Test status output when using **only** aggregation in flow project.
- [ ] Test status output when using aggregation with non-aggregate operations in flow project.
- [ ] Test status with parameters when aggregation is used.
- [ ] Once labels for aggregation is implemented, test labels for aggregates.

flow/project.py
vyasr: @bdice any comments?

vyasr:
@kidrahahjo can you move that checklist to a separate issue? Then we can merge this PR.

flow/project.py
bdice: I just wanted to move the construction of `individual_jobs` closer to the place where it's actually used.

vyasr (approved):

Issue #362, opened 2020-11-08
To-Do list for aggregation


kidrahahjo:
### Description
Even though #335 introduces most of the changes required for aggregation, a few more pull requests are needed to get everything in place. This issue lists all the To-Do items for aggregation which are to be addressed in

### Related to #335 
- [x] Descriptive Variable Names i.e. variable names like `_aggregator_per_group`, `_stored_aggregates` doesn't seem to be immediately obvious for developers so a proper description of these variables is necessary.
- [x] Refactor tests as per this [comment](https://github.com/glotzerlab/signac-flow/pull/335#discussion_r506731035)
- [x] The private methods like `_fetch_status`, `_get_group_status`, etc are fairly complicated. We have to reduce the complexity and try to keep as minimum complexity as possible. (Refer these comments https://github.com/glotzerlab/signac-flow/pull/335#discussion_r504689525, https://github.com/glotzerlab/signac-flow/pull/335#discussion_r504698327, https://github.com/glotzerlab/signac-flow/pull/335#discussion_r504698752)
- [x] Normalize the parallelization process of operations, status fetching. Refer to this [comment](https://github.com/glotzerlab/signac-flow/pull/335#discussion_r504698752)
- [x] [Check for singleton groups](https://github.com/glotzerlab/signac-flow/pull/335/files/a4f0fe6f2769e83b882d4e7a1e5ca6f0ea342330#r492354699)
- [x] Remove all parallelization logic from status checks, inline the calls to methods like `_get_job_status` etc, and see how much redundant processing is being done that can be optimized away(https://github.com/glotzerlab/signac-flow/pull/335/files/a4f0fe6f2769e83b882d4e7a1e5ca6f0ea342330#r504698752)
- [ ] Update the API for [`run_operations`](https://github.com/glotzerlab/signac-flow/pull/335/files/a4f0fe6f2769e83b882d4e7a1e5ca6f0ea342330#r519890035)
- [x] Remove duplicated aggregate validation [here](https://github.com/glotzerlab/signac-flow/pull/335/files/a4f0fe6f2769e83b882d4e7a1e5ca6f0ea342330#r520249420).
- [ ] Add unit tests that properly tests unit operations, rather than the complex integration tests like [this one](https://github.com/glotzerlab/signac-flow/pull/335/files/a4f0fe6f2769e83b882d4e7a1e5ca6f0ea342330#r507204478)
[...]

PR #464, opened 2021-02-11
Enable Aggregate Operation

kidrahahjo:
## Description
Enable the use of aggregate operation in `flow`. 

#### How will jobs get aggregated?
- For operations, jobs will be aggregated based on aggregators associated with operations.
- For group, jobs will be aggregated based on aggregators associated with groups. 

#### Aggregate IDs
Similar to the concept of job ids, an aggregate id is a unique id assigned to every aggregate which is created in a project. An aggregate id will always have a prefix `agg-` so that users can distinguish between aggregate id and a job id.
Users can calculate the aggregate id of a job using flow's Python API.
``` python
from flow import get_aggregate_id
project = # Code to get a signac project
list_of_jobs = list(project)
id = get_aggregate_id(project)
print(id)  # Prints aggregate id for an aggregate of all jobs in a project.
```
#### Ways of Aggregation
- Aggregation of all jobs in a project.
- Aggregation of jobs grouped by keys, where keys can be state points, callables, or a sequence of state points. 
- Aggregation of jobs in "groups of" fashion.
- Sorting, or selecting jobs in an aggregate based on state points or a condition.

#### Code Example
``` python
# project.py

from flow import FlowProject, aggregator

group = FlowProject.make_group('agg_group', aggregator_obj=aggregator())

@FlowProject.operation
@aggregator()
@group
def op_1(*jobs):
    pass

@FlowProject.operation
@group
def op_2(*jobs): 
    pass
```

``` bash
# Run operations op_1 and op_2 individually. 
# op_1 acts as an aggregate operation, op_2 acts as a non-aggregate operation
$ python project.py run -o op_1 op_2

# Run operations op_1 and op_2 as a group.
# Both op_1 and op_2 acts as aggregate operations
$ python project.py run -o agg_group

# Run operation op_1 for an aggregate whose aggregate id is `agg-asdg3425jgnoet`
# op_1 will only get executed if the aggregator associated will op_1 generated an aggregate whose aggregate id is `agg-asdg3425jgnoet`
$ python project.py run -o agg_group -j agg-asdg3425jgnoet
```

## Motivation and Context

Explained in detail above!

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


vyasr:
@kidrahahjo I recommend that you take the opportunity now to open a new PR in the [signac-docs repo](https://github.com/glotzerlab/signac-docs/) that documents the aggregation feature. This is probably the best time for you to do it since all the logic will be fresh and you'll already be writing relevant explanations as part of this thread.

Is this PR waiting on anything else at this point (other issues being resolved, etc) or do you think that once this PR is approved we'll just be ready?

kidrahahjo:
@vyasr That's a good suggestion, I will start working on adding documentation for aggregation. For the second point of yours, I did not write regression tests for submission in order to restrict the length of this PR. Once this gets ready to merge, I will write submission tests and we can merge that pull request into this one. What do you think?

As far as documentation is concerned, @vyasr @bdice do you want me to write the documentation in a glossary like fashion? For more context, in yesterday's office hours, we came to a point that our concept page should be glossary like (inspired by what Airflow has on their [concepts page](https://airflow.apache.org/docs/apache-airflow/stable/concepts.html)). For topic documentation (like aggregation, groups, etc), we can choose to document things in a glossary like fashion as well.

flow/project.py
```diff
@@ -3549,9 +3563,9 @@ def submit(
         ----------
         bundle_size : int
             Specify the number of operations to be bundled into one submission, defaults to 1.
-        jobs : iterable of :class:`~signac.contrib.job.Job`
-            Only submit operations for the given jobs, or all if the argument
-            is None. (Default value = None)
+        jobs : iterable of :class:`~signac.contrib.job.Job` or aggregates of jobs
```
kidrahahjo, line 3566:
```suggestion
        jobs : iterable of :class:`~signac.contrib.job.Job` or iterable of aggregate of :class:`~signac.contrib.job.Job`
```

flow/project.py
```diff
@@ -3549,9 +3563,9 @@ def submit(
...
+            Only submit operations for the given jobs or aggregates of jobs,
+            or all if the argument is None. (Default value = None)
```
kidrahahjo, line 3568:
```suggestion
            or all the associated aggregates/jobs for an operation if the argument is None. (Default value = None)
```

This document really raises the question that do we want our users to know that internally, a normal operation always acts on aggregate of single job?

flow/project.py
```diff
@@ -4207,18 +4210,22 @@ def foo(job):
         options : str
             A string to append to submissions. Can be any valid
             :meth:`FlowOperation.run` option. (Default value = "")
+        aggregator_obj : :class:`~.aggregator`
+            aggregator object associated with the :class:`FlowGroup`. If None, the
```
kidrahahjo, line 4214:
```suggestion
            An instance of :class:`~flow.aggregator` to associate with the:class:`FlowGroup`. If None, the
```

aggregator object seems like a vague term, this feels more clear.

flow/project.py
bdice: I think the suggestion is okay but not necessary. Either way is fine with me. Maybe get another opinion on this.

flow/project.py
bdice: Seems fine to me.

flow/project.py
```diff
@@ -4242,6 +4249,16 @@ def _register_groups(self):
         for entry in group_entries:
             group = FlowGroup(entry.name, options=entry.options)
             self._groups[entry.name] = group
+            # Since we store aggregator parameter for operations as None,
+            # we fetch the aggregators associated with operations using
+            # the registered operations.
```
bdice, line 4254: I'm confused, is this comment related to the comment above? (quoted here)

```
        # We register aggregators associated with operation functions
        # in `_register_groups` and hence the aggregator parameter is passed as None.
        # This is because we do not restrict the decorator placement in terms of
        # `FlowGroupEntry`, `aggregator`, or `operation`.
```

flow/project.py
```diff
@@ -4271,6 +4288,14 @@ def _register_groups(self):
                 operation_name
             ] = directives
 
+    def reregister_aggregates(self):
```
bdice, line 4291: Is it possible to combine this logic with the original registration, so that there is only one implementation used by "registration" and "re-registration"?

flow/project.py
```diff
@@ -4518,12 +4543,13 @@ def _select_jobs_from_args(self, args):
             # aggregates must be a set to prevent duplicate entries
             aggregates = set()
             for id in args.job_id:
-                # TODO: We need to add support for aggregation id parameter
-                # for the -j flag ('agg-...')
-                try:
-                    aggregates.add((self.open_job(id=id),))
-                except KeyError as error:
-                    raise LookupError(f"Did not find job with id {error}.")
+                if id[0:4] == "agg-":
```
bdice, line 4546:
```suggestion
                if id.startswith("agg-"):
```

flow/project.py
```diff
@@ -4518,12 +4543,13 @@ def _select_jobs_from_args(self, args):
             # aggregates must be a set to prevent duplicate entries
             aggregates = set()
             for id in args.job_id:
```
bdice, line 4545: Change this variable name (perhaps `job_id`) so that it doesn't shadow the built-in function `id`.

flow/project.py
```diff
@@ -4751,8 +4777,8 @@ class MyProject(FlowProject):
             "job_id",
             type=str,
             nargs="*",
-            help="The job ids, as registered in the signac project. "
-            "Defaults to all jobs.",
+            help="The job ids present or aggregate ids registered in the FlowProject. "
```
bdice, line 4780:
```suggestion
            help="The job ids or aggregate ids in the FlowProject. "
```

flow/templates/base_status.jinja
```diff
@@ -27,8 +28,8 @@
 {% endblock%}
 
 {% block operation_summary %}
-| operation | number of eligible jobs |
-| --------- | ----------------------- |
+| operation | number of eligible scheduler jobs |
```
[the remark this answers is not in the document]
bdice: This is related to previous discussions where we've considered names like `ExecutionUnit` or `SubmissionTask`, we need to create a separate issue for this because we do need a name for that unit of work.

bdice (changes requested):
@kidrahahjo Thanks for your work on this! I have some comments, but I think I will need to come back to it with fresh eyes for a better review.

flow/templates/base_status.jinja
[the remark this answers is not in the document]
kidrahahjo: Then I believe that we should not make this change in this PR. I'll revert the related changes.

flow/project.py
```diff
@@ -175,7 +175,7 @@ def _isfile(*jobs):
 
     @classmethod
     def true(cls, key):
-        """Evaluate if a document key is True for the job(s).
+        """Evaluate if a document key evaluates to True for the job(s).
```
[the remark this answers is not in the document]
kidrahahjo: I agree, let me add this to #362 as a task for aggregation.

flow/project.py
kidrahahjo: Yes, this is related to the quoted comment. I think of it as a hack to disable to placement requirements of the `aggregator` decorator.

flow/project.py
```diff
@@ -685,24 +698,14 @@ class FlowGroupEntry:
         The :meth:`FlowProject.run` options to pass when submitting the group.
         These will be included in all submissions. Submissions use run
         commands to execute.
-
+    aggregator : :class:`~.aggregator`
+        aggregator object associated with the :class:`FlowGroup`.
```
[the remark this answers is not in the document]
kidrahahjo: This is not quite true, `None` doesn't necessarily mean that the default value is used. When used for a `FlowGroup` (created using `FlowProject.make_group`), only then `None` would mean the default aggregation is used. See this comment for reference

```
# We register aggregators associated with operation functions
# in `_register_groups` and hence the aggregator parameter is passed as None.
# This is because we do not restrict the decorator placement in terms of
# `FlowGroupEntry`, `aggregator`, or `operation`.
```

tests/test_project.py
```diff
@@ -1775,6 +1776,178 @@ class TestGroupDynamicProjectMainInterface(TestProjectMainInterface):
...
+                "use of @with_job decorator with aggregation" in runinfo.value.message
+            )
+
+    def test_main_run_cmd(self):
+        project = self.mock_project()
+        assert len(project)
+
+        run_output = self.call_subcmd("run -o agg_op4").decode().splitlines()
+
+        assert "1 and 2" in "\n".join(run_output)
```
bdice, line 1898: I think you can do this instead of splitting/joining. I recently made this change in some other tests. Match this change (remove `splittlines` and `join`) in the other tests below, please.
```suggestion
        run_output = self.call_subcmd("run -o agg_op4").decode("utf-8")

        assert "1 and 2" in run_output
```

bdice (changes requested):
Some additional comments.

flow/project.py
vyasr: Can you remind me why exactly this function is necessary to make part of the API again? Is it because the user needs to tell flow when jobs have been added to the data space? The implementation is different enough from the way that aggregators are determined inside `_register_groups` to give me a little pause.

tests/define_test_aggregate_project.py
```diff
@@ -0,0 +1,77 @@
+from flow import FlowProject, aggregator, cmd, with_job
+
+
+class _TestAggregateProject(FlowProject):
+    pass
+
+
+def custom_aggregator(jobs):
```
vyasr, line 8: This name is too generic and will make it harder for readers of the tests to know what the custom aggregation is based on. Rename it to something like `sp_i_parity_aggregator`.

tests/define_test_aggregate_project.py
```diff
@@ -0,0 +1,77 @@
+from flow import FlowProject, aggregator, cmd, with_job
+
+
+class _TestAggregateProject(FlowProject):
```
vyasr, line 4: There's no need to make this internal. This is part of our testing suite, not the package.
```suggestion
class TestAggregateProject(FlowProject):
```

tests/define_test_aggregate_project.py
vyasr: On second thought, I guess `parity` may not be a word everyone's familiar with, I'm not sure. If the next reviewer who sees this thinks that might be the case, then maybe we use an even simpler name like `sp_i_even_odd_aggregator`. Either name is fine with me, although my first suggestion is a bit less verbose :)

tests/define_test_aggregate_project.py
```diff
@@ -0,0 +1,77 @@
...
+def op1(job):
+    pass
+
+
+@_TestAggregateProject.operation
+@aggregator.groupby("even")
+def agg_op1(*jobs):
+    total = sum([job.sp.i for job in jobs])
+    for job in jobs:
+        job.document.sum = total
```
vyasr, line 27: All your tests are doing this. I'd define a small function
```
def set_all_job_docs(jobs, key, val):
    for job in jobs:
        job[key] = val
```
and use that. Then half of these operations become trivial (setting `True`) and the others do one calculation and call this.

tests/define_test_aggregate_project.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@_TestAggregateProject.operation
+@cmd
+@aggregator(sort_by="i", select=lambda job: job.sp.i <= 2)
+def agg_op4(*jobs):
+    return "echo '{jobs[0].sp.i} and {jobs[1].sp.i}'"
+
+
+@_TestAggregateProject.operation
```
vyasr, line 69:
```suggestion
# This operation should raise an exception and only exists to test that.
@_TestAggregateProject.operation
```

tests/test_project.py
```diff
@@ -1775,6 +1776,178 @@ class TestGroupDynamicProjectMainInterface(TestProjectMainInterface):
...
+class TestProjectUtilities(TestAggregatesProjectBase):
+    def test_AggregatesCursor(self):
+        project = self.mock_project()
+        agg_cursor = _AggregatesCursor(project=project)
+        assert agg_cursor._project == project
+        assert agg_cursor._filter is None
+        assert agg_cursor._doc_filter is None
+        # Since default aggregate is not present in project, hence no aggregator
+        # will be evaulated as equal and hence all the aggregates present in this project
+        # will be considered unique.
```
vyasr, line 1830: I don't think I understand what you're trying to say here. The "default aggregate" is just an individual job, right?

vyasr (changes requested):

bdice:
I'm going to spend a little time fixing up open review comments.

tests/define_test_aggregate_project.py
bdice: This should not be changed, it matched the style of the other projects in our testing suite. However, it should be named like:
```suggestion
class _AggregateTestProject(FlowProject):
```

bdice: Resolved in 7a8a806d52a8c6a01a8f32db90d087aa2e32b1fb.

tests/define_test_aggregate_project.py
bdice: Resolved in 1f3efb6b4ff57c8e9094347e3bc9a5a71ca22251.

flow/project.py
```diff
@@ -4172,18 +4175,22 @@ def foo(job):
         options : str
             A string to append to submissions. Can be any valid
             :meth:`FlowOperation.run` option. (Default value = "")
+        aggregator_obj : :class:`~.aggregator`
```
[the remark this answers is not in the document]
bdice: Renamed to `group_aggregator` in 76c06048da93724d819de21e4c406241a668b38f, changed elsewhere for consistency and to avoid shadowing the imported name `aggregator`.

flow/project.py
bdice: I don't know what to do about this immediately, but I think that it might be best to merge this as-is and simplify this logic later once we remove the legacy `add_operation` feature and consolidate group registration. Related discussion on PRs: #382, #487.

flow/render_status.py
```diff
@@ -158,10 +158,47 @@ def highlight(string, eligible, pretty):
...
+
+        Parameters
+        ----------
+        jobs : list
+            Status of signac jobs or aggregates of jobs.
+        kind : str
+            Nature of the job passed.
+
+            * 'jobs' (Unique signac jobs of :class:`~signac.contrib.job.Job`)
+            * 'aggregates' (Unique aggregates in :class:`~flow.project.FlowProject`)
```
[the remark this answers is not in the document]
bdice: Bulleted lists require a blank line before and after. Also if the string must be one of two options, you can specify it like the argument "engine" to `pandas.read_csv`:
https://github.com/pandas-dev/pandas/blob/f2c8480af2f25efdbd803218b9d87980f416563e/pandas/io/parsers.py#L176
https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html

I'll change that above, edit the docstring, and resolve this comment.

flow/render_status.py
```diff
@@ -158,10 +158,47 @@ def highlight(string, eligible, pretty):
...
 
+    def project_length(jobs, kind):
+        """Get the number of unique signac jobs or aggregates present in the FlowProject.
+
+        Parameters
+        ----------
+        jobs : list
+            Status of signac jobs or aggregates of jobs.
+        kind : str
+            Nature of the job passed.
```
bdice, line 169:
```suggestion
        kind : {{"jobs", "aggregates"}}
            Whether to interpret the ``jobs`` argument as individual jobs or aggregates of jobs.
```

flow/aggregates.py
```diff
@@ -441,6 +441,8 @@ def _register_aggregates(self):
 
         Every aggregate is required to be a tuple of jobs.
         """
+        # Clear the internal mapping from id to aggregate
+        self._aggregates_by_id = {}
```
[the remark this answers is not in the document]
bdice: I removed this variable initialization from the constructor in 1d40e8309c7152f625d24105468a65bd673468c1.

tests/test_project.py
bdice: What this means is that all of the operations define a custom aggregator, and thus the aggregates cursor will have some aggregates with the same id. For example, `@aggregator()` and `@aggregator.groupsof(30)` will return the same aggregates (a single aggregate containing all 30 jobs in the project) but the `_AggregatesCursor` is not capable of distinguishing which aggregators are equivalent. Thus, there are 41 aggregates in the cursor, even though there are only 34 unique aggregate ids. I updated the comments and test to clarify this behavior in e8f3701e217df0615c8fc56e9540ca533fc0fbeb.

flow/render_status.py
```diff
@@ -158,10 +158,47 @@ def highlight(string, eligible, pretty):
         else:
             return string
 
+    def project_length(jobs, kind):
```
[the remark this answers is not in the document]
bdice: The purpose of this template filter is to show an output line in the status overview like:

```
Total # of jobs: 81
Total # of unique aggregates: 27
```

where the project has 81 jobs and 27 "proper" aggregates with more than one job. I am not a huge fan of this additional complexity, but I'm not immediately sure what would be better -- I improved this code in a few ways already and am alright with it for now. A better solution would be to get both values (jobs and aggregates) in a single iteration.

bdice:
I'm finished with the review pass and small updates that I wanted to make for now. There are a handful of significant design questions left to answer in `flow/project.py`, mostly raised by @vyasr. The next step on this PR is to tackle those.

flow/project.py
```diff
@@ -273,36 +273,49 @@ class _AggregatesCursor:
...
         self._filter = filter
         self._doc_filter = doc_filter
-        self._jobs_cursor = project.find_jobs(filter, doc_filter)
+        # If no filter or doc_filter is provided by the user, then select every
+        # aggregate present in the FlowProject. If filter or doc_filter are provided,
+        # then use the jobs returned via JobsCursor instance.
+        if filter is None and doc_filter is None:
+            self._cursor = project._group_to_aggregate_store.inverse.keys()
+        else:
+            self._cursor = project.find_jobs(filter, doc_filter)
```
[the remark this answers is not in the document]
kidrahahjo: 1. Yes, any filter provided will prevent finding aggregates.
2. As far as I remember, I think we decided that these filters should not be limited to jobs and hence, we should also be able to apply aggregate filters in future.

tests/test_project.py
```diff
@@ -1787,6 +1788,154 @@ class TestGroupDynamicProjectMainInterface(TestProjectMainInterface):
...
+
+    def test_invalid_with_job(self):
+        project = self.mock_project()
+        assert len(project)
+
+        with pytest.raises(Exception) as runinfo:
+            project.run(names=["agg_op5"])
+            assert (
+                "use of @with_job decorator with aggregation" in runinfo.value.message
+            )
```
b-butler, line 1888: Can we not detect this earlier than trying to run the operation?

b-butler (commented):
I looked over the code before our meeting today, and left just a few comments along the way.

flow/project.py
```diff
@@ -2979,9 +2997,9 @@ def run(
 
         Parameters
         ----------
-        jobs : iterable of :class:`~signac.contrib.job.Job`
-            Only execute operations for the given jobs, or all if the
-            argument is None. (Default value = None)
+        jobs : iterable of :class:`~signac.contrib.job.Job` or aggregates of jobs
```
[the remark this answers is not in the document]
bdice: Consensus of @bdice @atravitz @b-butler @vyasr : This should be left as-is:
```
iterable of :class:`~signac.contrib.job.Job` or iterable of aggregate of :class:`~signac.contrib.job.Job`
```

flow/project.py
bdice: Resolved by consensus on a new wording.

flow/project.py
```diff
@@ -277,36 +277,49 @@ class _AggregatesCursor:
...
-            # sufficient performance with earlier versions of signac.
-            return aggregate[0] in self._project
-        # Slow path: requires O(N) iteration over the JobsCursor
-        return aggregate[0] in self._jobs_cursor
+            return any(
+                get_aggregate_id(aggregate) in aggregate_store
+                for aggregate_store in self._cursor
+            )
+        else:
+            return aggregate[0] in self._cursor
```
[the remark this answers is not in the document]
bdice: @vyasr has an idea for improvement.

flow/project.py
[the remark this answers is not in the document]
bdice: @b-butler Can you take a look at this? `FlowGroupEntry` allows `group_aggregator=None` but `make_group` replaces `group_aggregator=None` with `group_aggregator=aggregator.groupsof(1)`.

flow/project.py
bdice: @b-butler Also this comment is related to the above discussion of group/operation initialization.

flow/render_status.py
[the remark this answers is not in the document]
bdice: I thought some more about this and I'm pretty sure that we should pre-compute this information in `print_status` if it is needed. Also there were merge conflicts with PR #501 so the existing design of this PR had to be changed. To punt on further debate about the purpose/scope of this function, I removed it in commit 4cc3cba902986f67008f5ec80b8908f21d54ca6a and we can re-visit this at a later time if we wish to change the status output format to distinguish jobs and aggregates. I'm going to resolve this comment thread for now since it doesn't seem like any of us have strong opinions (except that we don't want this to be a blocker for all of aggregation).

Current output: `Overview: 11 jobs/aggregates, 11 jobs/aggregates with eligible operations.`

tests/test_project.py
```diff
@@ -1715,3 +1716,206 @@ def test_main_submit(self):
...
+                    return subprocess.check_output(
+                        _cmd.split(), stderr=subprocess.DEVNULL
+                    )
+        except subprocess.CalledProcessError as error:
+            print(error, file=sys.stderr)
+            print(error.output, file=sys.stderr)
+            raise
+
+
+class TestProjectUtilities(TestAggregatesProjectBase):
```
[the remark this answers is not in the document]
bdice: Status output tests will be handled in another PR that @Charlottez112 has started. Tests for status output are currently outside the scope of this PR so I'll resolve this.

flow/project.py
```diff
@@ -273,36 +273,49 @@ class _AggregatesCursor:
...
+        # aggregate present in the FlowProject. If filter or doc_filter are provided,
+        # then use the jobs returned via JobsCursor instance.
+        if filter is None and doc_filter is None:
+            self._cursor = project._group_to_aggregate_store.inverse.keys()
+        else:
+            self._cursor = project.find_jobs(filter, doc_filter)
 
     def __eq__(self, other):
-        return self._jobs_cursor == other._jobs_cursor
+        if not (
```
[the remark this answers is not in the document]
bdice: Done: fdf7343d25a1a9cd606721a98fbd35e1dce3f307

tests/define_test_aggregate_project.py
bdice: Done in 6266b3728188381fc4af60dadd9fc9f074f9dab1.

flow/project.py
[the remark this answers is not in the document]
vyasr: Here's my suggestion. Currently most methods of `AggregatesCursor` behave differently based on the exact same criterion, namely whether or not there are filters, but semantically the presence of a filter indicates something more because if we have a filter we only iterate over a `JobsCursor`, except that we wrap the returned jobs in a tuple. That to me indicates that we should have (very small) subclasses to distinguish the two cases, with one case being a true `AggregatesCursor`, whereas the other is very thin wrapper around a `JobsCursor` that returns tuples instead.

tests/test_project.py
bdice: Fixed in 865277b2a7b65c31bf2fe9b8ac0a183b073cd5b1.

flow/operations.py
```diff
@@ -51,6 +51,8 @@ def hello(job):
 def with_job(func):
     """Use ``arg`` as a context manager for ``func(arg)`` with this decorator.
 
+    This decorator can only be used for operations that accept a single job as a parameter.
+
```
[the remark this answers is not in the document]
bdice: Done.

flow/project.py
```diff
@@ -681,24 +694,14 @@ class FlowGroupEntry:
...
-        # def __init__(self, name, options="", aggregator=None):
+    def __init__(self, name, options="", aggregator=None):
         self.name = name
         self.options = options
-        self.aggregator = _aggregator.groupsof(1)
-        # TODO: Enable aggregator argument.
-        # if aggregator is None:
-        #     aggregator = _aggregator.groupsof(1)
-        # self.aggregator = aggregator
+        self.aggregator = aggregator
```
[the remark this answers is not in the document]
bdice: Discussed with @kidrahahjo: This has to do with delayed initialization. We cannot know at this point whether there is an aggregator affiliated with the operation. We added this comment at another point in the code, and I copied it here in 1680f46f2f5755e053efe0ab243c1f9b045575a7.
```
        # We register aggregators associated with operation functions in
        # `_register_groups` and we do not set the aggregator explicitly.
        # We delay setting the aggregator because we do not restrict the
        # decorator placement in terms of `@FlowGroupEntry`, `@aggregator`, or
        # `@operation`.
```

flow/project.py
bdice: I agree this is worth further investigation. For now, we made the method private in e52aa43. I am going to document this in a separate TODO issue.

bdice:
**Reviewers:** I have worked through this PR with @kidrahahjo and addressed a number of issues. All of the remaining comments pertain to the implementation of `_AggregatesCursor`. I would recommend that we perform a final round of review on this PR, ignoring that class, and make a separate pull request (into this PR or into `master` depending on the timing of merge) that addresses the suggestions from @vyasr and others around `_AggregatesCursor`. I will re-request review from all.

bdice (approved):
I made a few last rounds of changes to documentation and expanded the tests / code coverage. I think this is ready to merge aside from the outstanding issues with `_AggregatesCursor`. I would vote in favor of merging this PR in its current state and leaving that class refactor as a "TODO" to get done before the next release.

tests/test_project.py
```diff
@@ -1836,6 +1868,171 @@ class TestGroupDynamicProjectMainInterface(TestProjectMainInterface):
...
+        self.project = self.mock_project()
+        os.chdir(self._tmp_dir.name)
+        request.addfinalizer(self.switch_to_cwd)
+
+
+class TestAggregatesProjectUtilities(TestAggregatesProjectBase):
+    def test_AggregatesCursor(self):
+        project = self.mock_project()
+        agg_cursor = _AggregatesCursor(project=project)
+        assert agg_cursor._project == project
```
b-butler, line 1898:
```suggestion
        assert agg_cursor._project is project
```

This is really an identity check isn't it?

b-butler (approved):
I have a tiny suggestion that can just be committed or resolved, but I generally agree with @bdice and am approving.

tests/test_project.py
bdice: I agree. Committing.

bdice:
I think this PR is ready to merge! We're punting on the `_AggregatesCursor` redesign for a follow-up PR. @vyasr You have an outstanding review request but we already know your primary set of change requests on the cursor design. Please feel free to follow up if you have additional comments and we can address those in subsequent PRs. https://github.com/glotzerlab/signac-flow/pull/464

PR #466, opened 2021-02-14
Handle memory requests via directives

### LINKED ISSUE #258: Handle job memory requests via directives
### Feature description

While working with @jdice on #256 and #257, I considered whether our memory usage requests should be handled with directives. I think this would be a good use case for directives.

### Proposed solution

Add a directive `@directives(memory='4g')` or similar. It may need to parse strings like `1024m` or `4g` and do its math in terms of gigabytes, or we could always assume units of gigabytes and only accept numeric inputs.

### Additional context

Currently memory requests are handled inconsistently across clusters/schedulers - SLURM's handling is improved by #256 and #257 but TORQUE/LSF needs to be made consistent at some point.

kidrahahjo:
## Description
Use the memory directive for handling memory requests in all the schedulers. The default memory reserved for each bundle is `4 gigabytes`. Users can still use the `--memory` flag with CLI. In case they specify the `--memory` flag (where `memory` is only valid when the suffix is either "g", "G",  "m", or "M") this flag will override the memory directive specified by the user. In the `memory` directive, users can only enter a numeric value in gigabytes.
## Motivation and Context
Resolves #258 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


kidrahahjo:
@bdice @vyasr I tried covering all the environments. Please let me know how to proceed on this. I think once this gets merged, we need to update signac docs.

vyasr:
@kidrahahjo thanks for this! I think this is the right idea. As far as deprecation, I would add a deprecation in `_main_submit` that checks whether `memory` is one of the passed kwargs. I think that's the only place where it would be necessary.

flow/util/template_filters.py
```diff
@@ -109,6 +109,61 @@ def calc_tasks(operations, name, parallel=False, allow_mixed=False):
...
+    except ValueError:
+        raise ValueError(
+            'Invalid memory passed. For gigabytes use suffix "g", '
+            'for megabytes use suffix "m".'
+        )
+    except TypeError:
+        raise
+
+
+def _calc_memory(operations, memory=None):
```
bdice, line 145: This needs to be public API.
```suggestion
def calc_memory(operations, memory=None):
```

flow/util/template_filters.py
```diff
@@ -109,6 +109,61 @@ def calc_tasks(operations, name, parallel=False, allow_mixed=False):
...
+        The parsed memory (numeric value) in gigabytes.
+    """
+    try:
+        size_type = memory[-1]
+        if size_type.lower() == "m":
+            return float(memory[0:-1]) / 1024
+        elif size_type.lower() == "g":
+            return float(memory[0:-1])
+        else:
+            raise ValueError
```
bdice, line 135: Should this default to gigabytes? If not, we need to add a message for the error.
```suggestion
            raise ValueError("Memory amounts must be specified with a suffix "g" for gigabytes or "m" for megabytes.")
```

flow/templates/lsf.sh
```diff
@@ -2,6 +2,7 @@
 {% block header %}
 #!/bin/bash
 #BSUB -J {{ id }}
+#BSUB -M {{ operations | _calc_memory(memory) }}G
```
bdice, line 5: Can you verify if this is the correct syntax? I found two references online for BSUB memory requests, and both of them state that this should be given as an integer value in _kilobytes_.

https://scicomp.ethz.ch/public/manual/LSF/9.1.3/lsf_admin/index.htm?chap_interactive_jobs.html~main
http://www.glue.umd.edu/lsf-docs/man/bsub.html

bdice: Submissions that do not define a memory directive should not add this line, so this needs to be set in a conditional block. See the section below with `{% if walltime %}` for an example.

flow/templates/slurm.sh
```diff
@@ -2,12 +2,10 @@
 {% block header %}
 #!/bin/bash
 #SBATCH --job-name="{{ id }}"
+#SBATCH --mem={{ operations | _calc_memory(memory) }}G
```
bdice, line 5: Submissions that do not define a memory directive should not add this line, so this needs to be set in a conditional block. See the section below with `{% if walltime %}` for an example.

flow/templates/torque.sh
```diff
@@ -1,15 +1,13 @@
 {% extends "base_script.sh" %}
 {% block header %}
 #PBS -N {{ id }}
+#PBS -l pmem={{ operations | _calc_memory(memory) }}G
```
bdice, line 4: Submissions that do not define a memory directive should not add this line, so this needs to be set in a conditional block. See the section below with `{% if walltime %}` for an example.

bdice: The references I found below use `mem` instead of `pmem` and expect the suffix to be `gb` or `GB`. I'm not sure what is best to use here.

https://www.osc.edu/supercomputing/batch-processing-at-osc/pbs-directives-summary
https://docs.pace.gatech.edu/faqs/pbs/

flow/util/template_filters.py
```diff
@@ -109,6 +109,61 @@ def calc_tasks(operations, name, parallel=False, allow_mixed=False):
         )
 
 
+def _parse_memory(memory):
```
bdice, line 112: Rename this to make it clearer that it's parsing a string to a numerical value.
```suggestion
def _parse_memory_flag(memory):
```

bdice: If this should be removed in 0.14 (when `--memory` is removed) then please add a deprecation notice or comment that contains the string "0.14" (that's what I'll search for when checking that deprecated code was removed).

flow/util/template_filters.py
```diff
@@ -109,6 +109,61 @@ def calc_tasks(operations, name, parallel=False, allow_mixed=False):
...
+        size_type = memory[-1]
+        if size_type.lower() == "m":
+            return float(memory[0:-1]) / 1024
+        elif size_type.lower() == "g":
+            return float(memory[0:-1])
+        else:
+            raise ValueError
+    except ValueError:
+        raise ValueError(
+            'Invalid memory passed. For gigabytes use suffix "g", '
```
bdice, line 138:
```suggestion
            'Invalid memory flag. For gigabytes use suffix "g", '
```

bdice (changes requested):
Changes requested in comments. Overall this is a good start, we just need to iron out some scheduler-specific stuff.

flow/util/template_filters.py
```diff
@@ -109,6 +109,61 @@ def calc_tasks(operations, name, parallel=False, allow_mixed=False):
...
+
+    Returns
+    -------
+    float
+        The parsed memory (numeric value) in gigabytes.
+    """
+    try:
+        size_type = memory[-1]
+        if size_type.lower() == "m":
+            return float(memory[0:-1]) / 1024
```
bdice, line 131:
```suggestion
            return float(memory[:-1]) / 1024
```

flow/util/template_filters.py
```diff
@@ -109,6 +109,61 @@ def calc_tasks(operations, name, parallel=False, allow_mixed=False):
...
+        elif size_type.lower() == "g":
+            return float(memory[0:-1])
```
bdice, line 133:
```suggestion
            return float(memory[:-1])
```

flow/templates/lsf.sh
kidrahahjo: I saw this documentation which says that we can use `G` or `GB` for `gigabytes` but by default, the memory is specified in kilobytes.
https://www.ibm.com/support/knowledgecenter/SSWRJV_10.1.0/lsf_command_ref/bsub.__m.1.html

flow/util/template_filters.py
kidrahahjo: The command line flag currently requires users to enter memory with units in gigabytes (`g`) or megabytes(`m`). In order to default it to `gigabytes`, we need to deprecate the current behaviour. (Right?)

flow/util/template_filters.py
kidrahahjo: Sounds good!

flow/templates/torque.sh
kidrahahjo: For the command line flag, we ask users to enter memory to request per node. 
`pmem` -> Memory to request per process
`mem` -> Memory to request overall.

I am not sure what's the context of "per node" in our case. (I think it means total memory so we should use `mem`)

flow/templates/slurm.sh
kidrahahjo: I agree that we should add it only when users ask us to but we define a default value for the memory directive [here](https://github.com/glotzerlab/signac-flow/blob/master/flow/directives.py#L380), I thought this was done in order to actually default the memory requests to 4 gigabytes. Can you please confirm this?

bdice: Good observation. signac-flow should not define a default memory amount because (a) it breaks backwards compatibility and (b) doesn't match how schedulers would normally act. We should set the directive's default to `None`. I'm not sure off the top of my head how walltimes work with directives when some operations define a walltime and others do not, but that would be a good reference. I propose that accumulation would be done like this:
- In serial, we take the max. The max of a defined amount (e.g. 4 GB) and an undefined amount (`None`) should be the defined amount (4 GB).
- In parallel, we take the sum. The sum of a defined amount (e.g. 4 GB) and an undefined amount (`None`) should be the defined amount (4 GB).

flow/templates/lsf.sh
bdice: Excellent! That's the right reference, I just couldn't find it when I was googling. Let's use `GB` here.
```suggestion
#BSUB -M {{ operations | _calc_memory(memory) }}GB
```

flow/util/template_filters.py
bdice: Currently values without a suffix `g` or `m` are not supported. I was proposing adding a _new_ behavior (not breaking a current behavior) that would allow a numeric value with no suffix to be interpreted as gigabytes.

If you think that providing a raw number without a suffix (and interpreting it as gigabytes) _should_ be allowed, then accept the first suggestion `return float(memory)`. If you think that providing a raw number without a suffix _should not_ be allowed, then accept the second suggestion `raise ValueError(...)` because it makes the error message clear.

flow/templates/torque.sh
bdice: I agree with your assessment. Let's use `mem`, we intend to request a total amount of memory (not an amount per core requested).

vyasr:
@bdice @kidrahahjo let me know if you would like me to review this, I am happy to provide another set of eyes on this if you think it would be helpful or if there are any questions/issues you're not able to resolve between the two of you.

bdice:
@vyasr Thanks for the offer, I'd really appreciate your input on this PR when @kidrahahjo is ready for review!

kidrahahjo:
@bdice @vyasr I think this is ready now, please have a look.

flow/directives.py
```diff
@@ -255,32 +255,36 @@ def _evaluate(value, jobs):
...
-            return self.type(value)
-        except Exception:
-            raise TypeError(
-                f"Expected an object of type {self.type}. "
-                f"Received {value} of type {type(value)}."
-            )
-
-
-def _raise_below(threshold):
+        for type_ in self.types:
```
kidrahahjo, line 272: I am not really satisfied with the logic I used here. But since it's internal and the variables which use this are internal as well, I am not really sure if we should be too worried about this.

vyasr: I think this code is fine. It looks identical to the place in hoomd-blue where I assume @b-butler [originally advocated copying it from](https://github.com/glotzerlab/hoomd-blue/blob/master/hoomd/data/typeconverter.py#L153).

flow/directives.py
```diff
@@ -255,32 +255,36 @@ def _evaluate(value, jobs):
 
 
 class _OnlyType:
```
vyasr, line 257: Since this class now accept multiple types, `_OnlyTypes` would be a more appropriate name.

flow/directives.py
```diff
@@ -331,7 +335,7 @@ def _is_fraction(value):
 _natural_number = _OnlyType(int, postprocess=_raise_below(1))
 _nonnegative_int = _OnlyType(int, postprocess=_raise_below(0))
 _nonnegative_real = _OnlyType(float, postprocess=_raise_below(0))
-_positive_real = _OnlyType(float, postprocess=_raise_below(1e-12))
+_positive_real = _OnlyType(float, type(None), postprocess=_raise_below(1e-12, True))
```
vyasr, line 338: I would add a comment that 1e-12 is an arbitrarily chosen minimum threshold for what constitutes 0.

flow/directives.py
```diff
@@ -377,7 +381,7 @@ def _is_fraction(value):
 walltime. Defaults to 12 hours.
 """
 
-_MEMORY = _Directive("memory", validator=_positive_real, default=4)
+_MEMORY = _Directive("memory", validator=_positive_real, default=None)
```
vyasr, line 384: Is there a reason not to support units other than Gb? I see that the function for parsing of string memory flags is marked for deprecation because we'll be removing the `--memory` CLI arguments, but we could still support passing a string here. I have a slight preference for supporting this since it's a more natural mapping to the interface that users are accustomed to when manually writing cluster submission scripts, and I don't think there's much overhead to doing so (we can still default to Gb if no units are provided).

vyasr: This does make the code slightly more complex since the templates for different schedulers can't assume a suffix. I'm fine skipping this for now if you can't find a clean and simple solution, we can always add it later if people want that support.

flow/project.py
```diff
@@ -4498,6 +4500,15 @@ def _main_submit(self, args):
...
         names = args.operation_name if args.operation_name else None
+
+        if kwargs.get("memory", None):
+            print(
+                "WARNING: "
+                "The memory argument is deprecated as of 0.13 "
+                "and will be removed in 0.14. Use directives to handle the "
+                "reservation of memory per node.",
+                file=sys.stderr,
+            )
```
vyasr, line 4511: Use warnings.warn with a DeprecationWarning.

flow/util/template_filters.py
```diff
@@ -109,6 +109,96 @@ def calc_tasks(operations, name, parallel=False, allow_mixed=False):
...
+            operation.directives["memory"] if operation.directives["memory"] else 0
+            for operation in operations
+        ]
+        if parallel:
+            return sum(list_of_memories)
+        else:
+            return max(list_of_memories)
+    else:
+        parsed_memory = _parse_memory_flag(memory)
+        return parsed_memory if not parallel else parsed_memory * len(operations)
```
vyasr, line 199: I don't think this logic is quite right. The `--memory` command-line argument's help says that this is the amount of memory to request *per node*. That is also likely why the old torque.sh template uses `pmem` instead of `mem`. Unfortunately, this complicates our logic quite a bit because we don't usually know how many nodes we will have a priori, and it's a fundamentally different calculation than the per operation calculation. Since the `--memory` flag is command line, and I suspect that very few people are using it anyway, I think we may be better off skipping a deprecation cycle for it and just removing it rather than trying to support two incompatible measures of memory.

vyasr (changes requested):
You can look through my inline comments, but there is one significant change I'm suggesting that may make some of my comments inapplicable. The `--memory` command-line flag is actually incompatible with the memory directive in its current form because it claims to define "memory per node". We are not even consistent about supporting this correctly between different schedulers (lsf doesn't seem to have had any memory support, slurm treated the memory as a total rather than per-node, and torque did what we claimed it should), and this calculation can't be nicely mapped to the per-operation directive introduced by this PR. My suggestion would be to skip the deprecation cycle in this case and just remove support for the CLI memory entirely. It's a very minor break in the CLI that saves us significant problems here _and_ makes our execution model better defined going forward.

flow/project.py
kidrahahjo: This whole block is removed hence this change is no longer required

flow/directives.py
kidrahahjo: We now support

1. Positive numeric value with suffix "g" or "G" indicating memory requested in gigabytes.
2. Positive numeric value with suffix "m" or "M" indicating memory requested in megabytes.
3. Positive numeric value with no suffix indicating memory requested in gigabytes.

flow/util/template_filters.py
```diff
@@ -109,6 +109,49 @@ def calc_tasks(operations, name, parallel=False, allow_mixed=False):
...
+    """Calculate the maximum memory to reserve for submission of operations.
+
+    Parameters
+    ----------
+    operations : list
+        The operations of :class:`~._JobOperation` used to calculate the maximum memory required.
+    parallel : bool
+        If True, operations are assumed to be executed in parallel, which means
+        that the total memory requested will be the sum of all memories requested instead of the
+        maximum memory requested. (Default value = False)
```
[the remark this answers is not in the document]
vyasr: I thought that all instances of the period before the `(Default value = ...)` were created by Bradley during all his refactoring and performance optimization work, but perhaps some were yours as well? In any case, period after is much more commonly used in `signac` (and other projects in the `glotzerlab` organization that use Google docstrings, for instance `freud`, `rowan`, and `coxeter`). It would be nice to make this consistent everywhere... but it's probably the lowest priority task every so I'll just point it out here for future changes 😂 .

tests/test_util.py
```diff
@@ -48,3 +49,34 @@ def test_bidict(self):
...
+
+class TestTemplateFilters:
+    def test_calc_memory(self):
+        class MockOp:
+            def __init__(self, memory=None):
+                self.directives = {"memory": memory}
+
+        op1 = MockOp(1)
+        op2 = MockOp(8)
+        op3 = MockOp()
```
vyasr, line 62: You could define these as fixtures then reuse them in `test_check_memory` below.

vyasr (approved):
The code looks good to me now. Two main notes:
1. I'm not sure why the `template_reference_data.tar.gz` file shows up in the diff as empty. Can you double-check that the file has contents? I assume it's fine since tests are passing, but please just check again to be sure.
2. There are a couple of code paths missing coverage in `flow/directives.py`. It would be nice to add coverage for them.

The coverage issue isn't a big deal and IMO not worth blocking this PR for, @bdice feel free to comment when you review though. Before approving/merging, can you also make sure that the first issue (empty template reference data file) has been resolved one way or another. Aside from that I am approving this PR.

bdice:
@vyasr @kidrahahjo I think the template reference data appearing to have zero size is a bug in GitHub, and not a problem with this PR. I saw that on the last couple PRs that altered template reference data, but the file appeared just fine. You can check it just to be sure but don’t worry too much about it.

flow/util/template_filters.py
```diff
@@ -109,6 +109,49 @@ def calc_tasks(operations, name, parallel=False, allow_mixed=False):
         )
 
 
+def check_memory(operations):
```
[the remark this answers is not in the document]
bdice: Above, I suggested removing `check_memory` and simply checking the value of `calc_memory` in the template (`None` if no operations specify a memory directive). That would eliminate the need for Vyas's suggestion.

bdice (changes requested):
The PR looks good overall but can be simplified a bit by removing `check_memory` and only using `calc_memory`. I may try to make a pass on this PR on Monday/Tuesday to apply the suggested changes and help finish it up.

tests/test_util.py
kidrahahjo: Test check_memory is no longer needed but I'll still define them as fixtures.

flow/util/template_filters.py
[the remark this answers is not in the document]
kidrahahjo: The use of `check_memory` has now been removed.

PR #476, opened 2021-03-09
Handle walltime requests via directives

### LINKED ISSUE #240: Walltime should be a directive
### Feature description

The amount of time an operation should be submitted for should be part of the recognized directives dictionary, rather than an additional command line option.

### Proposed solution

Update the directives parsing in the `FlowProject` to handle walltime directives.

### LINKED ISSUE #295: Broken backwards-compatibility: walltime conversion
### Description

A concise description of what is causing an issue and what you expected to happen.

To prevent my jobs from running too long on the cluster and exiting cleanly, I usually set some environment variables with variables from the flow template. I get the total walltime (in seconds) via
`export WT={{ walltime*3600 }}`.

On `master`, (commit `7721966094419`), where support for `GreatLakesEnvironment` has been added, this functionality does not work. That expression gets evaluated to `3000 days, 0:00:00`, which is 3600x the length that I want to set (i.e., walltime is already in seconds in the template whereas it should be in hours).

I can get the old behavior back by using `export WT={{ walltime.seconds }}`.

### System configuration

This is on Great Lakes.

```
 $ python -c 'import platform; print(platform.platform()); import sys; print(sys.version); import signac; print(signac.__version__); import flow; print(flow.__version__)'
Linux-3.10.0-1062.9.1.el7.x86_64-x86_64-with-glibc2.10
3.8.3 | packaged by conda-forge | (default, Jun  1 2020, 17:43:00) 
[GCC 7.5.0]
1.4.
```

bdice:
## Description
Use the walltime directive for handling walltime requests in all the schedulers. The default walltime reserved for each bundle is 12 hours. Users can still use the `--walltime` flag with CLI. In case they specify the `--walltime` flag this flag will override the walltime directive specified by the user. In the walltime directive, users can only enter a numeric value in hours.

## Motivation and Context
Resolves #240, resolves #295.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


flow/project.py
```diff
@@ -3613,7 +3616,7 @@ def submit(
                 "The 'names' argument must be a sequence of strings, however "
                 f"a single string was provided: {names}."
             )
-        if walltime is not None:
+        if walltime is not None and not isinstance(walltime, datetime.timedelta):
```
kidrahahjo, line 3619: Since we're considering handling the walltime requested via directives, I propose deprecating the walltime argument from the public api.

flow/environment.py
```diff
@@ -434,9 +434,6 @@ def add_args(cls, parser):
 
         """
         super().add_args(parser)
-        parser.add_argument(
```
kidrahahjo, line 437: Via command line, this is the only gateway to the `walltime` argument in `submit`. I propose of deprecating it and not removing it.

kidrahahjo (commented):
This looks good to me, I suggested a few changes here. @bdice It'd be really great to add tests for `walltime` directive in `test/define_template_test_project.py`. If you agree, please react a :+1: to this review and I'll go ahead and make the changes.

flow/project.py
bdice: Yes, that is probably the right idea.

flow/environment.py
bdice: I agree, that makes sense.

bdice:
@kidrahahjo Good comments — it would be better to do a deprecation cycle as you mentioned. Also adding the tests you suggested would be good. Please go ahead and push to this PR if you have time, I won’t be able to work on it today.

b-butler (commented):
Beside adding the appropriate code for deprecation, this looks good.

kidrahahjo:
@bdice @b-butler This might require one more review pass.

bdice:
I will take a look! Thanks @kidrahahjo! I spent my signac dev time on https://github.com/glotzerlab/signac/issues/528 this week, I appreciate your help on this PR.

bdice (commented):
I have some comments and will address those comments and some additional changes (pushing commits after testing locally).

flow/directives.py
```diff
@@ -332,6 +333,32 @@ def _is_fraction(value):
...
+
+    A valid walltime argument is defined as:
+
+    1. Numeric value indicating walltime requested in hours.
+    2. :class:`datetime.timedelta` value.
+    3. None to indicate no specific walltime request.
+
+    Parameters
+    ----------
+    walltime : float or :class:`datetime.timedelta`
```
bdice, line 347:
```suggestion
    walltime : float or :class:`datetime.timedelta` or None
```

flow/directives.py
```diff
@@ -332,6 +333,32 @@ def _is_fraction(value):
...
+        Walltime to request.
```
bdice, line 348:
```suggestion
        Requested walltime.
```

flow/environment.py
```diff
@@ -225,7 +225,12 @@ def add_args(cls, parser):
             The argument parser where arguments will be added.
 
         """
-        return
+        parser.add_argument(
```
bdice, line 228: I moved all the "walltime" parser options to the base class `ComputeEnvironment`. We will remove this in 0.14, and this unifies the implementation in the meantime so that deleting it in 0.14 is easier. Until then, every environment has a `--walltime` option, and that also allows us to test the output behavior with the `StandardEnvironment`/`slurm.sh` in the walltime project tests.

tests/test_templates.py
```diff
@@ -115,4 +115,4 @@ def test_env(env, monkeypatch):
 
                         with open(job.fn(f"script_{op}.sh")) as file:
                             reference.extend([msg] + file.read().splitlines())
-            assert "\n".join(reference) == "\n".join(generated)
+            assert "\n".join(generated) == "\n".join(reference)
```
bdice, line 118: I swapped the direction of the template tests so that the "reference" data is treated as the base for the diff that shows up in the error output. This confuses me _every_ time I deal with these tests.

PR #483, opened 2021-03-23
Fix template converting mem to float

atravitz:
## Description
Fixes #482 

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change<sup>1</sup>

<sup>1</sup>The change breaks (or has the potential to break) existing functionality.

## Checklist:

- [x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac-flow/blob/master/ContributorAgreement.md).
- [x] My name is on the [list of contributors](https://github.com/glotzerlab/signac-flow/blob/master/contributors.yaml).
- [x] My code follows the [code style guideline](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md#code-style) of this project.
- [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.

If necessary:
- [x] I have updated the API documentation as part of the package doc-strings.
- [ ] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
- [ ] I have updated the [changelog](https://github.com/glotzerlab/signac-flow/blob/master/changelog.txt).

PR #484, opened 2021-03-23
Default memory requests to integer values

### LINKED ISSUE #482: flow converts memory directive to float
### Description

flow passes the `memory` directive to SLURM as a float, which at least for greatlakes is invalid.

### To reproduce

Passing `memory=8` results in a SLURM script containing `--mem=8.0G`

### Error output

`sbatch: error: Invalid --mem specification`

### System configuration

 - Operating System [e.g. macOS]:
 - Version of Python [e.g. 3.7]:
 - Version of signac [e.g. 1.0]:
 - Version of signac-flow: 0.13.0

Or copy & paste the output of: `python -c 'import platform; print(platform.platform()); import sys; print(sys.version); import signac; print(signac.__version__); import flow; print(flow.__version__)'`

atravitz:
## Description
Fixes #482 

## Motivation and Context
Float values are not considered valid by SLURM schedulers, so integers (of gigabytes or megabytes) are required.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:


b-butler (approved):
LGTM

bdice:
@atravitz @b-butler This would not correctly handle a request of `"512m"` -- it would convert to gigabytes and round down to zero.

bdice (changes requested):
@atravitz @b-butler This would not correctly handle a request of "512m" -- it would convert to 0.5 gigabytes and round down to zero.

Options, in order of my preference:
1. Convert non-integer values into megabytes.
2. Round up to the next integer (doesn't actually obey user request).
3. Disallow non-integer values (require input to be an integer number of gigabytes).

tommy-waltmann:
I like the first option a lot better than the other two.

b-butler:
@bdice I wasn't aware that we were allowing non-gigabyte specifications. In that case number 1 seems appropriate.

kidrahahjo:
I had the same concern as @bdice mentioned when I reviewed #483. Before we make any change in this regard, we should definitely test the code on the clusters we support and then decide to generalise this.

vyasr:
I've typically seen <1 Gb requests specified in Mb, not in fractional Gb. I think specification in Mb is much more familiar to users than fractional Gb, so we should definitely support that.

csadorf:
Let's make sure to use the correct prefix here ([GiB](https://en.wikipedia.org/wiki/Gigabyte#Base_2_(binary)) vs [GB](https://en.wikipedia.org/wiki/Gigabyte#Base_10_(decimal))! See also [_Understanding the difference between GB and GiB_](https://docs.ukcloud.com/articles/other/other-ref-gib.html).

vyasr:
> Let's make sure to use the correct prefix here ([GiB](https://en.wikipedia.org/wiki/Gigabyte#Base_2_(binary)) vs [GB](https://en.wikipedia.org/wiki/Gigabyte#Base_10_(decimal)))! See also: [_Understanding the difference between GB and GiB_](https://docs.ukcloud.com/articles/other/other-ref-gib.html).

Fair point. I think most official documents describing schedulers either spell out "gigabyte" or just use G, while most unofficial documents I've read (like the [Great Lakes user guide](https://arc.umich.edu/greatlakes/user-guide/)!) use the base 10 and base 2 equivalents freely without clarification. We may as well be precise in our discussions though.

I think your comment mostly applies to my previous comment 🙃. Everywhere in the code regarding directives I think it's just a single-letter prefix "G" or "g".

csadorf:
Yeah, I think using `G` as an ambiguous abbreviation is fine as long as it is clearly documents what it refers to and then make the appropriate conversion. Ideally we support the full unit symbol as well.

flow/util/template_filters.py
```diff
@@ -112,6 +112,18 @@ def calc_tasks(operations, name, parallel=False, allow_mixed=False):
         )
 
 
+def format_memory(memory):
+    """Format a float as either (int)G or (int)M
+    Returns
+    -------"""
```
bdice, line 118:
```suggestion
    """Format memory request as a string with an appropriate suffix.
    
    Parameters
    -----------
    memory : float
        Value of memory to format in gigabytes.

    Returns
    -------
    str
        String value with suffix "G" or "M".
    """
```

bdice (approved):
Good! Just needs an updated comment.

bdice (approved):
Thanks @atravitz! This looks good to me.

bdice:
@tommy-waltmann @b-butler @klywang This can be merged once we get a second review. (@b-butler I dismissed your first review approval because the PR has changed substantially since then.)

bdice:
> 
> 
> Yeah, I think using `G` as an ambiguous abbreviation is fine as long as it is clearly documented what it refers to and then make the appropriate conversion. Ideally we support the full unit symbol as well.

@csadorf Do you mean we should support "GB" as an input for the directive? This would be fine to support but I think it's out of scope for this PR. Feel free to create an issue!

tommy-waltmann (approved):

Issue #479, opened 2021-03-20
Old add_operation API is broken


vyasr:
### Description

The `add_operation` method of creating new operations no longer seems to work.

### To reproduce

Create the project file below.

```
# project.py
from flow import FlowProject

class Project(FlowProject):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.add_operation(
            'hello', cmd='echo hello', pre=[], post=[])

if __name__ == "__main__":
    Project().main()
```

Then execute `python project.py run` on the command line.

### Error output

The run above should print hello. Instead, it does nothing.

Looking through the code, the issue first appears during the run sequence in `Project._generate_selected_aggregate_groups` where the variable `aggregate_stores` is empty. The reason is that `self._group_to_aggregate_store` is empty. The problem is that `Project._register_groups` does not add items to `self._group_to_aggregate_store` from `self._operations` (i.e. the singleton operations. @bdice, @kidrahahjo, and @b-butler have all worked on this part of the code recently, so if any of you have a sufficient recollection to know offhand whether just adding instance-level operations to the bidict would be a safe and sufficient solution, let me know.

Alternatively, I traced the problem to #335 (unsurprisingly) using a git bisect. Just from looking at the diff, the problem is that the change to `_register_groups` in #335 only affects groups registered to the class, and that the PR likely missed the corresponding changes necessary to register aggregates for groups that are only added to the instance.

I have some other reviews that need to get done first, but if nobody gets around to this by then I can dig a little deeper.

csadorf:
A fix should probably contain a very basic unit tests that prevents a future introduction of this regression.

csadorf:
Since this is breaking a basic function, we should probably release a hotfix for this.

Issue #309, opened 2020-06-30
Move directives decorator to operations.directive


vyasr:
### Feature description

It would be nice if groups and operations were unified to a great degree. One area of cognitive dissonance is that directives are currently assigned as a distinct decorator for operations, but are part of the groups namespace, i.e.
```
@FlowProject.operation
@group.directives(...)
@directives(...)
def op(job):
    pass
```

### Proposed solution

We should consider moving this decorator into the operation namespace, enabling something more symmetric like:

```
@FlowProject.operation(...)
@group.directives(...)
def op(job):
    pass
```

b-butler:
Currently the APi for `FlowGroupEntry` is `group.with_directives`. I propose we use that instead of `operation.directive`

csadorf:
@vyasr I don't fully follow your example, is the idea to provide to directives as part of the `FlowOperation` constructor?

vyasr:
@csadorf yes exactly, in the same way that the API works for groups.

csadorf:
Ok, although I like @b-butler 's proposal.

bdice:
I agree that @b-butler's suggestion would make for a pretty and intuitive API. To make things more concrete for future readers / further discussion:

```python
@TestProject.operation
@flow.directives(nranks=4, omp_num_threads=2)
def hybrid_op(job):
    pass
```

would become one of the following:

```python
# Vyas's proposal: fewer words required.

@TestProject.operation(nranks=4, omp_num_threads=2)
def hybrid_op(job):
    pass

# Brandon's proposal: match the API of group.with_directives.

@TestProject.operation.with_directives(nranks=4, omp_num_threads=4)
def hybrid_op(job):
    pass
```

csadorf:
I'm in favor of Brandon's proposal! 👍

bdice:
Resolved by #502.

PR #487, opened 2021-03-30
Remove FlowProject.add_operation

b-butler:
## Description

<s>This PR makes `FlowProject.add_operation` a classmethod as per the development team's discussion on 2021-03-29. The method does essentially what it did before, but at the class level instead of the instance level storage. This means that using the method on an instance will not add the operation nor be capable of warning the user. A test of this method is added to `test_project.py`.</s>

I removed `FlowProject.add_operation` and added a warning in its place. This decision for direct removal comes from conversations shown in the PR and offline. We have decided to remove the registering of operation at both the instance and operation level, in favor of only registering at the class level.

In addition, I noticed some API requirements that we were not enforcing or were not enforcing early in terms of operation functions/names which I added checks for. I can add tests for these if others like the new checks.

## Motivation and Context

We decided to change `FlowProject.add_operation` in the development meeting on 2021-03-29 to address #479.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:


b-butler:
I am actually not sure if it makes sense now for `add_operation` to be a classmethod as it essentially just does what the decorators do a bit more concisely in some cases. I would be happy to field others' thoughts on this.

bdice (changes requested):
I chatted with @b-butler offline about his comment above (whether it makes sense to keep `add_operation`). The changes in this PR are currently *incompatible* with the past instance-level behavior of `add_operation`. That is, existing projects using `add_operation` would need to be updated in order to work (registering an operation with the class would require a re-instantiation of the class to get a new instance with those operations defined). The new behavior is also somewhat redundant with the class-level registrations that are performed by `@FlowProject.operation` decorator syntax.

I would like to get input from @glotzerlab/signac-committers on whether it is important to us to keep supporting instance-level operations and/or the `add_operation` style syntax in general. I do not think the benefits in functionality outweigh the costs in engineering time and future maintenance. In addition, the current implementations are a source of significant complexity in the codebase:
- #479 arose because this API was untested (a sign of low usage)
- Resolving #237, #309, and several other issues (can't locate the one I'm thinking of at the moment) would be easier if the code base didn't have to support so many methods of interaction/registration

I vote to remove this API entirely, with no deprecation period (since it is already broken).

csadorf:
> I would like to get input from @glotzerlab/signac-committers on whether it is important to us to keep supporting instance-level operations and/or the `add_operation` style syntax in general. I do not think the benefits in functionality outweigh the costs in engineering time and future maintenance. In addition, the current implementations are a source of significant complexity in the codebase:

While unfortunate, I agree that it appears that the effort for maintaining this API seems to be too high at this point and removal is warranted. It is especially unfortunate that the removal would go via the route *untested -> broken -> removed* without any kind of controlled deprecation. I am still a bit surprised that it is difficult to maintain it as it is essentially a wrapper at this point, but I admit that I did not have a chance yet to look at the technical details of the problem. As a mitigation measure, I suggest we provide an extremely clear error message when people use this function after all with an explanation as to 1. why the function was removed, 2. why so sudden, and 3. how to migrate.

As one alternative to immediate removal: Could we provide a minimal variant of this function that is not functionally equivalent to `@operation`, but provides some basic support? That would give us a chance to properly deprecate and remove it.

vyasr:
> I chatted with @b-butler offline about his comment above (whether it makes sense to keep `add_operation`). The changes in this PR are currently _incompatible_ with the past instance-level behavior of `add_operation`. That is, existing projects using `add_operation` would need to be updated in order to work (registering an operation with the class would require a re-instantiation of the class to get a new instance with those operations defined). The new behavior is also somewhat redundant with the class-level registrations that are performed by `@FlowProject.operation` decorator syntax.
> 
> I would like to get input from @glotzerlab/signac-committers on whether it is important to us to keep supporting instance-level operations and/or the `add_operation` style syntax in general. I do not think the benefits in functionality outweigh the costs in engineering time and future maintenance. In addition, the current implementations are a source of significant complexity in the codebase:
> 
>     * #479 arose because this API was untested (a sign of low usage)
> 
>     * Resolving #237, #309, and several other issues (can't locate the one I'm thinking of at the moment) would be easier if the code base didn't have to support so many methods of interaction/registration
> 
> 
> I vote to remove this API entirely, with no deprecation period (since it is already broken).

Clearly I didn't think through this suggestion when I made it enough to recognize that at the point `add_operation` is called the object is already instantiated and therefore has already collected the class-level operations. Sorry about making you do this work to figure that out @b-butler. I think fixing this problem is certainly possible, but quite a bit more work than it's worth and am also in favor of removing the old API.

> While unfortunate, I agree that it appears that the effort for maintaining this API seems to be too high at this point and removal is warranted. It is especially unfortunate that the removal would go via the route _untested -> broken -> removed_ without any kind of controlled deprecation. I am still a bit surprised that it is difficult to maintain it as it is essentially a wrapper at this point, but I admit that I did not have a chance yet to look at the technical details of the problem. As a mitigation measure, I suggest we provide an extremely clear error message when people use this function after all with an explanation as to 1. why the function was removed, 2. why so sudden, and 3. how to migrate.

@csadorf here's a summary of why maintaining `add_operation` is hard now. With the introduction of the decorator syntax back in 0.6 or so, we created an independent code path because some operations were registered to `FlowProject` subclasses (via the decorator) while other operations were added upon instance creation (using `add_operation`). That was largely handled using some registration logic that, upon instantiation of a `FlowProject` or a subclass, would 1) walk the MRO and pull all operations associated with all parent classes, then 2) add the operations created via `add_operation` (which in theory could happen any time after instantiation, but in practice was done in `__init__` because everyone copied that pattern from early examples of the project template etc). That was largely manageable until the introduction of groups, but now between groups and aggregation there are a lot of different entities that have to be managed independently along these code paths then merged on project instantiation, and that's what makes it nontrivial to maintain. There's much longer discussion of this on #382 if you'd like to look through that and see if it sparks any ideas on a better solution.

At this stage @b-butler and myself are probably the only ones currently involved with the project who are intimately familiar with the code paths taken using the old API and how they interact with the decorator API. You and @b-butler worked very closely together on groups, so you made sure that that groups would support the old API, but the lack of examples or test using the old API allowed @kidrahahjo (who's done a fantastic job of it) to implement all of aggregation essentially without considering the implications for the old API. I completely dropped the ball on considering this when reviewing aggregation work, but also considering how long aggregation has taken to get finished (and considering that groups took even longer!) I'm not sure that this is a bad thing in hindsight. The upshot is that it probably would have required investing significantly more dev effort for minimal payoff. So while I agree that the current situation is unfortunate and would have preferred a proper deprecation cycle, I think a loud error message is probably the best option at this stage.

bdice:
@vyasr I was hunting all over for #382, that's the issue I couldn't find! 😄

As for strategies to move forward, it sounds like "loud error message" is what we're agreeing on for now. I think I would implement this as:

```python
def add_operation(self, *args, **kwargs):
    raise AttributeError(
		"The add_operation method has been removed without deprecation. This "
		"message will be removed in version 0.16."
	)
```

Then we can start to clean up the messy registration logic.

vyasr:
Probably because you were looking in issues, but #382 was a PR 😄 (which you opened while you were trying to make what you thought was a minor fix 😂 if only). @csadorf in particular, what you would probably want to read are [@bdice's summary of what he found and his questions about the internals of operation/groups](https://github.com/glotzerlab/signac-flow/pull/382#issuecomment-739456466), and [my subsequent comment on that thread](https://github.com/glotzerlab/signac-flow/pull/382#issuecomment-739679375) explaining why we have these two paths. Note that the discussion on #382 is only part of the equation because it doesn't account for the ways in which aggregations are mapped to groups (which also would need to support both the old and new APIs.

csadorf:
Yes, we should completely remove any duplication of internal logic which also includes any registration of operations after instantiation, this is something we should simply no longer support. What I'm still wondering is whether we could at least provide a very simple wrapper classmethod as a mitigation measure.

I just implemented the following snippet:
```python
class Project(FlowProject):

    @classmethod
    def add_operation(cls, name, operation, directives=None, pre=None, post=None):

        if isinstance(operation, str):

            @flow.cmd
            def op_(job):
                return f"{operation} {job}"

            op_ = cls.operation(op_, name=name)

        else:
            op_ = cls.operation(operation, name=name)

        if directives is not None:
            op_ = flow.directives(**directives)(op_)

        if pre:
            for cond in pre:
                op_ = cls.pre(cond)(op_)
        if post:
            for cond in post:
                op_ = cls.post(cond)(op_)

        return op_
```

This actually works so I am bit confused as to what the issue is.

bdice:
@csadorf I think the `add_operation` API is currently usable like this:

```python
project = Project()
project.add_operation(...)
project.main()
```
In my understanding, the snippet you're proposing is similar to what's in this PR (unless I'm not seeing some crucial difference) -- but this won't work on instances in the same way that the old instance method could.

csadorf:
> @csadorf I think the `add_operation` instance method API is meant to be usable like this:
> 
> ```python
> project = Project()
> project.add_operation(...)
> project.main()
> ```
> 
> In my understanding, the snippet you're proposing is similar to what's in this PR (unless I'm not seeing some crucial difference) -- but this won't work on instances in the same way that the old instance method could. The operation will be registered with the class `Project` but the instance `project` already exists and won't find that operation unless re-instantiated. I'm not 100% sure if that is correct but that's my understanding of the current implementation.

You're right, I should have read the patch more diligently. I don't mean to imply that this would work without any changes, on the user side, but it might make the migration a bit easier. But either way, I am also ok with just removing it.

Maybe we can make the error message a bit nicer? How about this:

```python
def add_operation(self, *args, **kwargs):
    raise AttributeError(
        "The add_operation() method was removed in version 0.14. "
        "Please see https://docs.signac.io/en/latest/flow-project.html#defining-a-workflow "
        "for instructions on how to define operations using the current API. This message "
        "will be removed in version 0.16."
    )
```

I don't think we actually need to mention that it was removed without deprecation; it just adds insult to injury...

bdice:
@csadorf I like your message! Thanks for the suggestion. 😊

b-butler:
@bdice and @csadorf, I have removed the method and left the suggested error message.

flow/project.py
```diff
@@ -4065,11 +4008,19 @@ def hello(job):
...
+            raise ValueError("A condition function cannot be used as an operation")
+
         if name is None:
             name = func.__name__
 
-        if (name, func) in cls._OPERATION_FUNCTIONS:
+        if name in map(lambda x: x[0], cls._OPERATION_FUNCTIONS):
             raise ValueError(f"An operation with name '{name}' is already registered.")
+        if func in map(lambda x: x[1], cls._OPERATION_FUNCTIONS):
+            raise ValueError("An operation with this function is already registered.")
```
bdice, line 4023: The structure of `x` and meaning of `x[0]` are not clear from this pattern. We can avoid the map/lambda and use readable names. Here are two suggestions, pick whichever you prefer.
```suggestion
        if any(name == op_name for op_name, op_func in cls._OPERATION_FUNCTIONS):
            raise ValueError(f"An operation with name '{name}' is already registered.")
        if any(func == op_func for op_name, op_func in cls._OPERATION_FUNCTIONS):
            raise ValueError("An operation with this function is already registered.")
```
Or iterate just one time:
```suggestion
        for op_name, op_func in cls._OPERATION_FUNCTIONS:
            if name == op_name:
                raise ValueError(f"An operation with name '{name}' is already registered.")
            elif func == op_func:
                raise ValueError("An operation with this function is already registered.")
```

bdice (approved):
@b-butler I pushed a couple changes for cleanup and added one suggestion for improvement. I approve this after that suggestion is resolved.

bdice:
@b-butler Can you update this PR description to match its current implementation, for future reference? It can be brief.

flow/project.py
```diff
@@ -3927,75 +3929,18 @@ def labels(self, job):
...
-        if name in self._groups:
-            raise KeyError("A group with this identifier already exists.")
-        self._groups[name] = FlowGroup(
-            name, operations={name: op}, operation_directives=dict(name=kwargs)
+    @deprecated(
+        deprecated_in="0.14",
+        removed_in="0.16",
+        current_version=__version__,
+        details="Method has been removed.",
+    )
```
csadorf, line 3937: Does it make sense to use the `@deprecated` decorator here since the function was not deprecated, but just removed?

flow/project.py
```diff
@@ -4089,11 +4032,24 @@ def hello(job):
...
             name = func.__name__
 
-        if (name, func) in cls._OPERATION_FUNCTIONS:
-            raise ValueError(f"An operation with name '{name}' is already registered.")
+        for registered_name, registered_func in cls._OPERATION_FUNCTIONS:
+            if name == registered_name:
+                raise ValueError(
+                    f"An operation with name '{name}' is already registered."
+                )
+            if func == registered_func:
```
csadorf, line 4049:
```suggestion
            if func is registered_func:
```
This is an identity check, right?

flow/project.py
```diff
@@ -4089,11 +4032,24 @@ def hello(job):
         if isinstance(func, str):
             return lambda op: cls.operation(op, name=func)
 
+        if func in chain(
+            *cls._OPERATION_PRECONDITIONS.values(),
+            *cls._OPERATION_POSTCONDITIONS.values(),
+        ):
+            raise ValueError("A condition function cannot be used as an operation")
+
```
csadorf, line 4040: These changes seem somewhat unrelated and are not covered by the changelog entry. I assume this is in here due to historical reasons? I might suggest to split this off cleanly.

flow/project.py
b-butler: For function objects, I believe so.

flow/project.py
b-butler: We can do that, I didn't mention this in the changelog because we already make these API promises, and there is not an issue relating to this.

flow/project.py
bdice: @csadorf The function hasn't been removed -- its _implementation_ has been removed. This approach allows us to raise an `AttributeError` with a helpful message, until we remove the function entirely. The deprecation warning is possibly redundant because that user message contains the same info, but I thought it was better to include because developers might search the codebase for `deprecated` when finding code to remove around release time.

flow/project.py
bdice: Sure, let's just split these changes into a separate PR for cleanliness.

csadorf:
@b-butler Please, re-request the review from me when this is ready. Thank you.

csadorf (approved):
Great!

PR #488, opened 2021-03-31
Add submission summary feature.

### LINKED ISSUE #472: Submission summary in top section of status output
### Problem

I submitted an operation and then checked the status. I noticed that the status output indicated that the jobs were "eligible" but submitting again did nothing (because the submission logic knows not to re-submit jobs that are already running). However, the status code does not distinguish between eligible-by-conditions (preconditions met AND (zero postconditions OR at least one unmet postcondition)) and eligible-for-submission (`scheduler_status < JobStatus.submitted`).

### Proposed solution

In signac dev meeting on March 1, 2021, we discussed this and decided that we want to keep the definition of "eligible" that is used most consistently in the source: "a job-operation is eligible if and only if (preconditions met AND (zero postconditions OR at least one unmet postcondition))". To solve the lack of clarity in the status output, we want to add a new column that summarizes the scheduler interactions. It would look something like this:

```
operation          number of eligible jobs  submission status
---------------  -------------------------  -----------------------
compute_stuff                          147  99 [U], 15 [Q], 33 [A]

[U]:unknown [R]:registered [I]:inactive [S]:submitted [H]:held [Q]:queued [A]:active [E]:error
```

This solution is more informative than any other proposed, because it directly includes scheduler information while retaining the current definition of "eligibility" that is determined by only the workflow conditions.

Separately, we could rename the internal method `FlowProject._eligible_for_submission(group, aggregate, scheduler_status, scheduler_info)` so that it doesn't use the word "eligible," to avoid confusion for developers.

bdice:
## Description
Adds a column for submission summary at the top of the status overview. This makes it easier to see what is happening when interfacing with cluster scheduling systems, and to disambiguate "eligible" from "submittable".

## Motivation and Context
Resolves #472. See that issue for some more info.

image

In this example, I have two operations. I have 123 queued instances of `mpb_integrate_dos`. I have 5 queued instances of `mpb_plot_dos`, and the rest of the 27 (22 more instances, this is not shown) recently became eligible and are not yet submitted.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


atravitz:
I'd prefer to have a set order for the statuses (rather than being ordered by number of operations) so:

``` 
[U] : 40, [Q]: 10
[U] : 5,  [Q]: 42
```

bdice:
> 
> 
> I'd prefer to have a set order for the statuses (rather than being ordered by number of operations) so:
> 
> ```
> [U] : 40, [Q]: 10
> [U] : 5,  [Q]: 42
> ```

Done! Good idea, I was just excited about using the `Counter.most_common` feature. 😆

bdice:
@tcmoore3 @jennyfothergill This is ready for review. Thanks!

jennyfothergill (approved):
LGTM! I only have a simple project that I'm testing this on, but everything works and this is a helpful change. :)

tcmoore3 (approved):
LGTM

PR #496, opened 2021-04-10
Enforce Operation Requirements

b-butler:
## Description

This enforces two primary constraints we have in our work flow model:
1. Operations should not be condition functions and vice versa
2. A function and/or name should only be used once for all operations (not just the name-function pair)
This shouldn't be considered an API break in my opinion, but if someone is writing unorthodox scripts this could technically break their work-flow.

## Motivation and Context

These changes are in line with our current API design.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -26,8 +26,7 @@ Changed
 Fixed
 +++++
 
-- Enforce operations condition separation and single use of names and functions
-as operations (#496).
+- Enforce operations condition separation and single use of names and functions as operations (#496).
 - Changed default value of ``status_parallelization`` to none, to avoid bugs in user code caused by thread parallelism and overhead in process parallelism (#486).
 - Memory directives are converted to an integer number of gigabytes or megabytes in submission scripts (#482, #484).
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -26,7 +26,7 @@ Changed
 Fixed
 +++++
 
-- Enforce operations condition separation and single use of names and functions as operations (#496).
+- Strictly enforce that operation functions cannot be used as condition functions (and vice-versa) and prevent the registration of two operations with the same name (#496).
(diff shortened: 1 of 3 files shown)


bdice (approved):
For the historical record, this PR was split off from #487. LGTM.

bdice:
@b-butler It would be good to add a changelog line. Here are some cute dinosaurs who say thank you for the PR. 🦖 🦕

b-butler:
@bdice thanks dinos! Also I had the changelog entry, but I forgot to push it. :rofl:

changelog.txt
```diff
@@ -26,6 +26,8 @@ Changed
 Fixed
 +++++
 
+- Enforce operations condition separation and single use of names and functions
+as operations (#496).
```
csadorf, line 30: there should be no line break here

flow/project.py
```diff
@@ -4089,11 +4089,24 @@ def hello(job):
         if isinstance(func, str):
             return lambda op: cls.operation(op, name=func)
 
+        if func in chain(
+            *cls._OPERATION_PRECONDITIONS.values(),
+            *cls._OPERATION_POSTCONDITIONS.values(),
+        ):
+            raise ValueError("An condition function cannot be used as an operation")
```
csadorf, line 4096:
```suggestion
            raise ValueError("An condition function cannot be used as an operation.")
```

changelog.txt
```diff
@@ -26,6 +26,7 @@ Changed
 Fixed
 +++++
 
+- Enforce operations condition separation and single use of names and functions as operations (#496).
```
csadorf, line 29: I  find the changelog entry a little cryptic. How about this:
```suggestion
- Strictly enforce that operation functions cannot be used as condition functions (and vice-versa) and prevent the registration of two operations with the same name. (#496).
```

csadorf (approved):
Two suggestions, but overall good. Thank you very much @b-butler !

PR #498, opened 2021-04-13
Make LSF / SLURM scheduler detection consistent with PBS.

bdice:
## Description
See question in issue #53. Either we should accept this PR and make the scheduling behavior consistent with the PBS changes adopted in PR #54, or we should make `is_present` return `False` for all schedulers if the scheduler version cannot be determined.

Regardless of whether the result is `True` or `False`, this PR should ensure that all schedulers catch `subprocess.CalledProcessError`. If they don't, the resulting user message raised by `flow.get_environment()` is potentially confusing.

Pros/cons of each possible behavior:

If we make `is_present` return False when the exception is caught, then we break the behavior mentioned in https://github.com/glotzerlab/signac-flow/issues/53 where compute nodes cannot submit jobs (only login nodes), which would also likely break important parts of signac-flow.

If we make `is_present` return True when the exception is caught, then we run the risk of detecting the wrong kind of environment. However, this is already a risk we run if two scheduler executables happen to exit with code 0 when asked for their version, regardless of whether both actually work.

## Motivation and Context
Related to a question from @klywang on Andes, which incorrectly detects `bjobs` (LSF) even though it uses SLURM and not LSF. `bjobs -V` fails to execute and raises a confusing `subprocess.CalledProcessError` that the user does not expect, because it fails when querying the LSF scheduler (which isn't even the right scheduler for the Andes system).

## Types of Changes
- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -22,6 +22,7 @@ Changed
 +++++++
 
 - Renamed TorqueEnvironment and related classes to PBSEnvironment (#491).
+- LSF and SLURM schedulers will appear to be present if the respective commands ``bjobs -V`` or ``sbatch --version`` exit with a non-zero error code (#498).
 
 Fixed
 +++++
--- a/flow/scheduling/lsf.py
+++ b/flow/scheduling/lsf.py
@@ -160,6 +160,8 @@ class LSFScheduler(Scheduler):
         """Return True if an LSF scheduler is detected."""
         try:
             subprocess.check_output(["bjobs", "-V"], stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError:
+            return True
         except OSError:
             return False
         else:
--- a/flow/scheduling/slurm.py
+++ b/flow/scheduling/slurm.py
@@ -154,6 +154,8 @@ class SlurmScheduler(Scheduler):
         """Return True if a SLURM scheduler is detected."""
         try:
             subprocess.check_output(["sbatch", "--version"], stderr=subprocess.STDOUT)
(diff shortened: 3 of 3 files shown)


bdice:
@klywang Can you please check that this PR (with no Andes template) hides the `subprocess.CalledProcessError` for you? You might get the wrong environment (it may detect LSF instead of SLURM), but don't worry about that. See discussion in the top of this PR.

mikemhenry:
Detecting which scheduler is present is tough and when we detect it incorrectly, those failures tend to happen fast. So while we could end up with some mistakes with this PR, I think it is better that we are consistent.

mikemhenry (approved):

klywang:
This PR detects the environment as `DefaultLSFEnvironment`.

```bash
$ python flowprojects/project.py status
Using environment configuration: DefaultLSFEnvironment
Querying scheduler...
bjobs: error while loading shared libraries: libnsl.so.1: cannot open shared object file: No such file or directory
ERROR:flow.project:Error during status update: Command '['bjobs', '-json', '-u', 'klywang']' returned non-zero exit status 127.
Use '--ignore-errors' to complete the update anyways or '--show-traceback' to show the full traceback.
ERROR: Encountered error during program execution: ''NoneType' object has no attribute '__traceback__''

Execute with '--show-traceback' or '--debug' to get more information.
```

This message is much more clear! Thank you.

klywang (approved):

PR #503, opened 2021-04-20
Only write known JobStatus values.

bdice:
## Description
This PR skips writing job-operations with `unknown` status to disk, by performing an update of the current status in-memory and then filtering out all job-operations with `unknown` status before writing to the project document.

## Motivation and Context
For most projects, the vast majority of job-operations have status `JobStatus.unknown == 1`.
For example, my current project document is 78 MB and it's _all_ unknown statuses.
I am currently seeing some challenges with scalability caused by the size of the project document when the `_status` key just holds a huge dictionary with long keys and values equal to 1, like `'training_dat/ec1a26016d5d4c941273675e0cc28cd1/mpb_plot_dos/96dad330426b7160a70b44af0e4c99a2': 1`. There should be no effect on the correctness of submission or status checks, because `JobStatus.unknown` is already the default value when no value is present.

This PR sped up my status check from 46.4 seconds to 39.4 seconds. 7.0 seconds = 15% faster. 🏎️
The new project document is only 15 bytes, instead of 78 MB.

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:
- [x] The changes introduced by this pull request are covered by existing tests.

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -23,6 +23,7 @@ Changed
 
 - Renamed TorqueEnvironment and related classes to PBSEnvironment (#491).
 - LSF and SLURM schedulers will appear to be present if the respective commands ``bjobs -V`` or ``sbatch --version`` exit with a non-zero error code (#498).
+- Only known JobStatus values will be written to the project document, to save space and writing time (#503).
 
 Fixed
 +++++
--- a/flow/project.py
+++ b/flow/project.py
@@ -1896,14 +1896,23 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
         try:
             yield status_update
         finally:
-            if status_update:
-                status_update = {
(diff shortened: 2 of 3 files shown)


tommy-waltmann (approved):

tommy-waltmann:
I already approved this, but we should add some kind of test for the new behavior

bdice:
> 
> 
> I already approved this, but we should add some kind of test for the new behavior

@tommy-waltmann I reviewed the existing tests and I believe that this behavior change is essentially transparent to the tests -- only internal/private APIs could see the effect of this change, and existing tests appear to be sufficient for correctness. I made one slight modification to improve a test case.

b-butler (approved):
LGTM, I don't think this should cause any error and existing tests pass.

Issue #196, opened 2019-12-13
flow load_config doesn't work with separate project directory from project.py


tommy-waltmann:
### Description

Somehow, the monkey-patching of flow's config isn't working. It is supposed to ensure that a default value of 0.2 is applied for the key `status_performance_warn_threshold`. It's running into a problem on this line because the `warn_threshold` is `None`. https://github.com/glotzerlab/signac-flow/blob/893017073a9ce0203e03efbdcab75f094e2d6793/flow/project.py#L2666

### To reproduce

```python
>>> from flow.util import config
>>> config.get_config_value('status_performance_warn_threshold') is None
True
```

### Error output

```python
$ python project_training.py status
Using environment configuration: UnknownEnvironment
Collecting job status info: 100%|███████████████| 1/1 [00:00<00:00, 247.47it/s]

# Overview: 
Total # of jobs: 1 

label  ratio                                              
-----  -------------------------------------------------- 

operation     number of eligible jobs
------------  -----------------------
train_ml      1


ERROR: Encountered error during program execution: ''>' not supported between instances of 'float' and 'NoneType''
```

### System configuration

Please complete the following information:

 - Operating System [e.g. macOS]: Darwin-18.7.0-x86_64-i386-64bit
 - Version of Python [e.g. 3.7]: 3.6.7 | packaged by conda-forge | (default, Nov  6 2019, 16:03:31) [GCC Clang 9.0.0 (tags/RELEASE_900/final)]
 - Version of signac [e.g. 1.0]: 1.2.0
 - Version of signac-flow: 0.8.0 (latest master)

bdice:
@tommy-waltmann and I dug into this further today. The actual problem is that we're using a signac project that is in a different directory than `project.py`. This causes `flow.util.config.load_config()` to fail -- it doesn't find the project's config properly.

The monkey patch is working fine, so I'll rename this issue accordingly.

bdice:
## Context

The core `signac` package uses config values that correspond to the data space itself (e.g. project id, workspace directory). The `flow` package uses config values that don't really have anything to do with the data space itself (e.g. account names for submission, status output settings). When running a `FlowProject` from a different directory than the signac project, this causes some dissonance. I often keep my source code (e.g. project operations) in a git repository that is separate from the signac data space. I use `FlowProject.get_project('/path/to/project')` to create my `FlowProject` instance, where the project path is usually a symlink to network storage or a scratch directory.

[...]

PR #504, opened 2021-04-25
Feature/schema

vyasr:
## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [ ] New feature

## Checklist:

If necessary:


vyasr:
I'm explicitly requesting review from @bdice and @csadorf on this one since they did most of the work on glotzerlab/signac#253 and this should be easiest for them to review.

flow/errors.py
```diff
@@ -61,3 +61,9 @@ def parse(self, parser):
     def err(self, msg, caller):
         """Raise a template error."""
         raise jinja2.TemplateError(msg)
+
+
+class IncompatibleSchemaVersion(RuntimeError):
+    """The FlowProject's schema version is incompatible with this version of flow."""
```
bdice, line 67: Pick one of the following. The second suggestion might make more sense, just depends on how we view the connection between the FlowProject and the filesystem. There is a clear definition of a signac project on disk and in memory (`Project`) but the association between a `FlowProject` and its configuration is a little less clear since `project.py` can be in another directory.
```suggestion
    """The FlowProject's schema version is incompatible with this version of signac-flow."""
```
```suggestion
    """The signac-flow configuration schema version is incompatible with this version of signac-flow."""
```

flow/migration/__init__.py
```diff
@@ -0,0 +1,113 @@
+# Copyright (c) 2021 The Regents of the University of Michigan
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+"""Handle migrations of signac schema versions."""
```
bdice, line 4:
```suggestion
"""Handle migrations of signac-flow schema versions."""
```

flow/migration/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+    def get_config_schema_version():
+        # TODO: The means of getting schema versions will have to change for
+        # flow versions in schema versions > 1 that no longer rely on signac's
+        # configuration file and schema.
+        return version.parse(get_config_value("schema_version", config=project.config))
+
+    if get_config_schema_version() > schema_version:
+        # Project config schema version is newer and therefore not supported.
+        raise RuntimeError(
+            "The signac-flow schema version used by this project is {}, but flow {} "
```
bdice, line 72:
```suggestion
            "The signac-flow schema version used by this project is {}, but signac-flow {} "
```

flow/migration/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+            "only supports up to schema version {}. Try updating flow.".format(
```
bdice, line 73:
```suggestion
            "only supports up to schema version {}. Try updating signac-flow.".format(
```

flow/migration/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+        )
+
+    while get_config_schema_version() < schema_version:
+        for (origin, destination), migration in MIGRATIONS.items():
+            if version.parse(origin) == get_config_schema_version():
+                yield (origin, destination), migration
+                break
+        else:
+            raise RuntimeError(
+                "The signac-flow schema version used by this project is {}, but flow {} "
```
bdice, line 85:
```suggestion
                "The signac-flow configuration schema version used by this project is {}, but signac-flow {} "
```

flow/migration/v0_to_v1.py
```diff
@@ -0,0 +1,46 @@
...
+    # that must be accounted for:
+    # 1. User updates signac before flow. In this case, signac's config API may
+    #    already have changed, and we cannot rely on it. Therefore, we need add
+    #    configobj as an optional requirement for flow rather than relying on
+    #    signac's config API (it's available on PyPI). Alternatively, we could
+    #    bundle configobj like we do in signac, but I don't want to commit to
+    #    that long term so that would at best be a short term solution (say for
+    #    one or two flow releases) and then after that we tell users who
+    #    haven't migrated to just `pip install configobj`. Then, this function
+    #    can rely on the fact that in signac schema version one the config
```
bdice, line 29: Does "version one" mean "v1" here? Just unclear about why it's spelled out.

flow/migration/v0_to_v1.py
```diff
@@ -0,0 +1,46 @@
...
+    #    information is stored in signac.rc using the known schema and operate
+    #    accordingly to pull it.
+    # 2. User updates signac and attempts to migrate before migrating flow
+    #    schema. In this case, `signac migrate` should error and tell the user
+    #    to update flow and run `flow migrate` first. This can be accomplished
+    #    by having the v1->v2 migration in signac check for the presence of the
+    #    "flow" key in the config and error if it is present.  We will
+    #    introduce the flow v1->v2 migration before signac's to ensure that
+    #    this be possible.
```
bdice, line 38:
```suggestion
    #    this is possible.
```

setup.py
```diff
@@ -16,6 +16,10 @@
     "deprecation>=2.0.0",
     # Progress bars
     "tqdm>=4.48.1",
+    # Platform-independent file locking
+    "filelock~=3.0",
```
bdice, line 20: I wish we pinned to `>=3.0` like all our other dependencies (both here and in signac).

bdice (commented):
Overall looks good. I have some questions and _possibly inconsistent_ suggestions about how to write the error and log messages. We need more consistent names for things like "signac-flow configuration schema" vs. "flow schema" vs. "flow configuration." Also I prefer to use the name signac-flow instead of flow, unless specifically referring to the package you import.

flow/errors.py
vyasr: I think that there are a couple of different questions we need to address here:

1. How is flow supposed to work if you put `project.py` in a different directory, unless you explicitly construct the `FlowProject` providing it the directory? You dug into this a bit in #196, but it looks like you never resolved the underlying issue since the problem somehow went away. 
2. Once we migrate to flow schema version 2, flow will have its own independent configuration. Should we require that configuration to live in the same directory as the `project.py` file? If not, how will we associate them?

One idea I've been toying with is whether config information should be on a per-FlowProject basis, rather than associated with a particular directory. That seems more consistent with some of the issues above regarding different directories, as well as the fact that multiple FlowProjects can be associated with a single signac project. This wouldn't resolve issues like where bundles are stored though, only the config variables.

I just opened glotzerlab/signac#541, which I think is also tangentially related to this discussion.

flow/migration/v0_to_v1.py
vyasr: No reason, just stream of consciousness :) I'd appreciate commentary on whether my overall logic here makes sense, though.

setup.py
vyasr: Pros: We wouldn't have to update until something breaks
Cons: CI could break unexpectedly when a major version bump changes APIs. You could view this as a pro though since it means we'd be alerted sooner rather than getting stuck on an older version forever.

I'd be fine changing.

flow/migration/v0_to_v1.py
bdice: I think it made sense.

flow/errors.py
vyasr: I think enough of this is resolved in discussions on #506 that we can move forward with this PR now.

flow/errors.py
```diff
@@ -61,3 +61,9 @@ def parse(self, parser):
     def err(self, msg, caller):
         """Raise a template error."""
         raise jinja2.TemplateError(msg)
+
+
+class IncompatibleSchemaVersion(RuntimeError):
+    """The FlowProject's configuration schema version is incompatible with this version of signac-flow."""  # noqa: E501
```
bdice, line 67: Shorter docstring, to align with other error docstrings.
```suggestion
    """Indicates FlowProject schema is not supported by this signac-flow version."""
```

flow/migration/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def get_config_schema_version():
+        # TODO: The means of getting schema versions will have to change for
+        # flow versions in schema versions > 1 that no longer rely on signac's
+        # configuration file and schema.
+        return version.parse(get_config_value("schema_version", config=project.config))
+
+    if get_config_schema_version() > schema_version:
+        # Project config schema version is newer and therefore not supported.
+        raise RuntimeError(
```
bdice, line 71: Should this be `IncompatibleSchemaVersion`? Seems like the messages are very similar.

flow/migration/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+            )
+        )
+
+    while get_config_schema_version() < schema_version:
+        for (origin, destination), migration in MIGRATIONS.items():
+            if version.parse(origin) == get_config_schema_version():
+                yield (origin, destination), migration
+                break
+        else:
+            raise RuntimeError(
```
bdice, line 85: Should this be `IncompatibleSchemaVersion`? Seems like the messages are very similar.

flow/project.py
```diff
@@ -1597,6 +1610,39 @@ def __init__(self, config=None, environment=None, entrypoint=None):
...
+        """
+        schema_version = version.parse(SCHEMA_VERSION)
+        config_schema_version = version.parse(
+            flow_config.get_config_value("schema_version", config=self.config)
+        )
+        if config_schema_version > schema_version:
+            # Project config schema version is newer and therefore not supported.
+            raise IncompatibleSchemaVersion(
+                "The signac-flow schema version used by this project is '{}', but flow {} "
+                "only supports up to schema version '{}'. Try updating flow.".format(
```
bdice, line 1630: Make this error message consistent with the one in `flow/migration/__init__.py`. The terms `signac-flow` and `flow` are interchanged (and possibly other differences).

flow/project.py
```diff
@@ -1597,6 +1610,39 @@ def __init__(self, config=None, environment=None, entrypoint=None):
...
+                    config_schema_version, __version__, schema_version
+                )
+            )
+        elif config_schema_version < schema_version:
+            raise IncompatibleSchemaVersion(
+                "The flow schema version used by this project is '{}', but flow {} "
+                "requires schema version '{}'. Please use '$ flow migrate' to "
+                "irreversibly migrate this project's schema to the supported "
+                "version.".format(config_schema_version, __version__, schema_version)
```
bdice, line 1639:
```suggestion
                "The signac-flow schema version used by this project is '{}', but signac-flow {} "
                "requires schema version '{}'. Please use '$ flow migrate' to "
                "irreversibly migrate this project's schema to a supported "
                "version.".format(config_schema_version, __version__, schema_version)
```

bdice: Is it possible that the package will support multiple schema versions concurrently? That is, should this be able to say "requires 3 <= schema version <= 5" or similar?

requirements/requirements-test.txt
```diff
@@ -3,3 +3,5 @@ coverage==5.5
 pytest-cov==2.11.1
 pytest==6.2.3
 ruamel.yaml==0.17.4
+filelock>=3.0
+packaging>=15.0
```
bdice, line 7: These requirements are in addition to the core `requirements.txt`. Only testing requirements should be listed here.

bdice (approved):
I have a few minor comments to resolve. Otherwise this is fine. @vyasr I don't have a complete understanding of what next steps are needed, so please continue leading this effort through the sequence of pull requests you have in mind.

csadorf:
> I considered making use of signac's migration logic directly in flow, but I decided against this because our primary goal right now is to get to a point where flow is independent of signac and we're going to have enough trouble disentangling the two without introducing additional version dependencies between them. We can consider refactoring the code in the future to expose a common migration API for any package in the signac framework, but for now I think the duplication of logic is a preferable path forward.

I don't understand this. Why should flow be indepenent of signac? The package strongly and unidirectionally depends on signac which means that there is no cross-dependency and we also have full control over what signac API is available to us. A user cannot install a version of signac that is incompatible with flow unless they break their environment.

Do we have this much trouble with this one dependency? Maybe for 2.0, we should consider to use one of the [standard Python plugin](https://packaging.python.org/guides/creating-and-discovering-plugins/) approaches for a better integration of signac-core and other packages.

flow/migration/__init__.py
```diff
@@ -0,0 +1,115 @@
+# Copyright (c) 2021 The Regents of the University of Michigan
```
csadorf, line 1: I honestly think it is a terrible idea to duplicate all of this logic here. Is it really that hard to adjust the existing framework to perform the migration?

flow/migration/v0_to_v1.py
```diff
@@ -0,0 +1,46 @@
...
+    pass  # nothing to do here, serves purely as an example
+
+
+def migrate_v1_to_v2(project):
+    """Migrate from schema version 1 to version 2."""
+    # TODO: This migration is not yet implemented, but this comment documents
+    # what this function will need to do. There are a few different scenarios
+    # that must be accounted for:
+    # 1. User updates signac before flow. In this case, signac's config API may
+    #    already have changed, and we cannot rely on it. Therefore, we need add
```
csadorf, line 22: That makes no sense. flow depends on signac ergo we can rely on its API.

flow/migration/v0_to_v1.py
```diff
@@ -0,0 +1,46 @@
...
+    #    bundle configobj like we do in signac, but I don't want to commit to
+    #    that long term so that would at best be a short term solution (say for
+    #    one or two flow releases) and then after that we tell users who
+    #    haven't migrated to just `pip install configobj`. Then, this function
+    #    can rely on the fact that in signac schema v1 the config information
+    #    is stored in signac.rc using the known schema and operate accordingly
+    #    to pull it.
+    # 2. User updates signac and attempts to migrate before migrating flow
+    #    schema. In this case, `signac migrate` should error and tell the user
+    #    to update flow and run `flow migrate` first. This can be accomplished
```
csadorf, line 34: Why would flows schema be affected by a schema change of signac core? signac's API is how access the config, any schema change in the core package should be transparent to dependencies, including flow.

flow/migration/v0_to_v1.py
```diff
@@ -0,0 +1,46 @@
...
+    #    by having the v1->v2 migration in signac check for the presence of the
+    #    "flow" key in the config and error if it is present.  We will
+    #    introduce the flow v1->v2 migration before signac's to ensure that
+    #    this is possible.
+    # 3. Users update signac and create a new project, but still use an old
```
csadorf, line 39: That's not possible unless they break their environment.

csadorf (commented):
I have major reservations to this change set and hopefully have outlined my arguments clear enough. If not, I am happy to discuss them either on this PR or on a call.

vyasr:
Just chatted with @csadorf. We agree that shared migration logic and zero dependence of flow on signac's internals (including API but also extending to configuration) would be great. We mainly differ on whether we need to protect users with invalid Python environments (specifically, improper signac/flow dependency management) from corrupting their metadata. The duplication of logic in this PR, many of the inline comments, and the general move to decouple flow's config from signac's are targeted at making that safer. However, @csadorf feels strongly that sticking to a single configuration file will make our lives much easier in the long term, and that we should not be trying to protect against those invalid Python environments.

While we were chatting, we identified the major interdependencies between signac and flow that make flow's reliance on signac's internals problematic. Removing flow's monkey-patching of signac's configuration and moving validation of the flow components of the config into flow would allow flow to safely depend on signac's config API. Since moving in that direction may make it less important to consider separating out flow's config, I'm going to focus on trying to make that change in the shorter term and then work my way back to this PR later and see if I can't find a better compromise solution that address both Simon and my concerns.

bdice:
@vyasr and @csadorf I’m happy to let you come to whatever solution you mutually feel is workable. I have removed my request for review for now.

vyasr:
This PR is no longer really relevant. signac and flow's configs are now completely decoupled. We may eventually need to do some form of schema maintenance for flow beyond the simple validation introduced in https://github.com/glotzerlab/signac-flow/pull/573, but we can cross that bridge when we reach a point that flow tries to modify its schema, and we can solve that use post-load jsonschema-based validation. The extensive replication of signac's schema code is no longer necessary.

PR #508, opened 2021-04-30
Execution hooks (core features)

bdice:
## Description
Extension of #189 that includes only core features. This will make the PR easier to review and finalize. See #189 for details.

To-do:
- [x] Update test class structure, see comment.
- [x] Add tests that improve coverage.
- [ ] Discuss API for hooks. What should the signature be? Currently `hook(operation_name, error, *jobs)` (`error` only for `on_fail`).
- [ ] Update docs.

## Motivation and Context
Implement core features of execution hooks.

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


changelog.txt
```diff
@@ -39,7 +40,7 @@ Added
 - Add support for aggregation (operations acting on multiple jobs) via ``flow.aggregator`` (#464, #516, #542).
 - Add official support for Andes cluster (#500).
 - Decorator for setting directives while registering operation function ``FlowProject.operation.with_directives`` (#309, #502).
-- Feature to install execution hooks for the automated execution of functions with operations (#28, #189, #508).
+- Add new flow command ``flow template create`` for automatic creation of custom templates (#520, #534).
```
csadorf: This looks like it's not properly rebased.

klywang: Shouldn't this,

> - Feature to install execution hooks for the automated execution of functions with operations (#28, #189, #508).

be in the changelog entry for this branch be under the unreleased version 0.16.0 instead of the released version 0.15.0?

bdice: Yes. This PR’s development time has spanned multiple releases, so the changelog messages need to be moved to an unreleased version.

flow/hooks/__init__.py
```diff
@@ -0,0 +1,111 @@
...
+        set_hooks(self, "on_finish", on_finish)
+        set_hooks(self, "on_success", on_success)
+        set_hooks(self, "on_fail", on_fail)
+
+    def __setattr__(self, name, value):
+        """Convert to _HooksList when setting a hook trigger attribute."""
+        if name in self._hook_triggers:
+            super().__setattr__(name, _HooksList(value))
+        else:
+            super().__setattr__(name, value)
```
klywang, line 88: Does this code let you create hooks with nonstandard triggers without throwing and error?

bdice: The "else" block here preserves the standard behavior of `__setattr__` for all Python objects. It's not really targeted towards nonstandard triggers -- I don't think this code supports nonstandard triggers.

That is, the "else" simply allows setting arbitrary attributes on a Python object, like all Python objects support:
```python
class A:
    pass

a = A()
a.special_attribute = "...fancy"
print(a.special_attribute)  # ...fancy
```

klywang:
@cbkerr Let me know what you think

flow/project.py
```diff
@@ -267,6 +268,25 @@ def _make_bundles(operations, size=None):
...
+    def __getattr__(self, name):
+        class _InstallHook:
+            def __init__(install_self, hook_func):
+                install_self.hook_func = hook_func
+
+            def __call__(install_self, func):
+                self._operation_hooks[func][name].append(install_self.hook_func)
+                return func
+
+        return _InstallHook
```
b-butler, line 287: Do we want to allow any attribute? I feel we should restrict it to those that are proper hook triggers.
```
@FlowProject.hook.in_the_middel(lambda: operation_name, jobs)
```
should raise a `FlowProjectDefinition` error or `AttributeError`.

flow/project.py
```diff
@@ -267,6 +268,25 @@ def _make_bundles(operations, size=None):
             break
 
 
+class _HooksDecorator:
+    def __init__(self):
+        self._operation_hooks = defaultdict(lambda: defaultdict(list))
+
+    def __getitem__(self, func):
+        return self._operation_hooks[func]
+
```
b-butler, line 277: Should we define this allowing users to access the current hooks through `func_hooks = FlowProject.hook[func]`? We don't expose `FlowProject._operation_hooks` but do expose `_HookDecorator`'s hook list. This means that after instantiation `hooks = project.hook[op_func]` may not be the true hooks used, and does not point to the true store of that information at that point `project._operation_hooks`.

flow/project.py
```diff
@@ -267,6 +268,25 @@ def _make_bundles(operations, size=None):
             break
 
 
+class _HooksDecorator:
```
b-butler, line 271: I would prefer the name `_HooksRegistery` to indicate its storage capacity before instantiation or some name that indicates its storage role. Alternatively we could add that information to a short developer docstring.

tests/define_hooks_install.py
```diff
@@ -0,0 +1,27 @@
...
+    )
+
+    def install_hooks(self, project):
+        project.hooks.on_start.append(set_job_doc(self.keys[0]))
+        project.hooks.on_finish.append(set_job_doc(self.keys[1]))
+        project.hooks.on_success.append(set_job_doc(self.keys[2]))
+        project.hooks.on_fail.append(set_job_doc_w_error(self.keys[3]))
+        return project
+
+    __call__ = install_hooks
```
b-butler, line 23: Since you use `install_hooks` below do you need `__call__ = install_hooks?

tests/test_project.py
```diff
@@ -2317,6 +2319,184 @@ def test_main_submit(self, monkeypatch):
...
+        self.call_subcmd(f"run -o {operation_name} -j {job.id}")
+
+        if job.sp.raise_exception:
+            assert get_job_doc_value(self.keys[3])[0]
+            assert get_job_doc_value(self.keys[3])[1] == self.error_message
+        else:
+            assert get_job_doc_value(self.keys[3]) is None
+
+
+class TestHooksInstallBase(TestHooksBase):
```
b-butler, line 2402: I am confused by these tests. I am likely just missing it, but shouldn't the project level hooks ensure that the job document is edited for all operations not just those with hook decorators?

b-butler (commented):
This is a really cool features so thanks to everyone who worked on this PR (I know this is the first of a couple).

I did leave a few comments and questions I had regarding the code. I notice there are already two reviewers, so if my review is not needed, let me know and I will remove myself.

changelog.txt
```diff
@@ -8,6 +8,23 @@ The numbers in brackets denote the related GitHub issue and/or pull request.
 Version 0.16
 ============
 
+[0.17.0] -- 2021-0x-xx
```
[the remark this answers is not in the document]
klywang: Oh line 8 should say `Version 0.17`, not `Version 0.16`. I corrected this in the most recent commit.

flow/project.py
klywang: https://github.com/glotzerlab/signac-flow/blob/fcab466a238ecc6bff524393dbae3573b5da1f38/flow/hooks/__init__.py#L16-L28

The restriction to proper hook triggers appears here and raises an `Exception`. Maybe we should change the error type though? `AttributeError` might more intuitive than an `Exception`. 

Should we include more comments throughout the code? I also missed the portion of the code that caught invalid hook attributes.

tests/test_project.py
klywang: These test should check operations with and without decorators in the functions inherited from `TestHooksBase`, where `test_start_finish` will test job document keys "installed_start" and "install_finish", `test_success` will test "installed_success", and `test_fail` will test "installed_fail" for operations with and without decorators. 

Tests with the name `test_decorated_*` are only run on operations with decorators and check for changes to the job document from those decorators (i.e. job document keys "start", "finish" instead of "installed_start" or "installed_finish").

b-butler: I guess what I mean is why don't we test that `base_no_decorators` has the keys `install_finish`, etc. written?

b-butler (changes requested):
I left some other comments to address, but I was wondering everyone's thoughts on having both `hook` and `hooks` attributes for a `FlowProject` where the distinction is between project wide hooks defined at instantiation and definition time per operation hooks. I think that we should differentiate them somehow through the attribute names.

tests/test_project.py
klywang: Oh  the `install_*` for `base_no_decorators` is tested! The operations tested in `TestHooksInstallBase` are paramaterized like this:
https://github.com/glotzerlab/signac-flow/blob/7ea174f5505c3d55dfbe1cc6269920426e535012/tests/test_project.py#L2410-L2412
While in `TestHooksInstallCmd`, `operation_name` is paramaterized as `base_cmd` and `base_cmd_no_decorators`.

The test functions inherited from `TestHooksBase` all accept `operation_name` as a parameter and run operations as such:
https://github.com/glotzerlab/signac-flow/blob/7ea174f5505c3d55dfbe1cc6269920426e535012/tests/test_project.py#L2371

Does that make sense?

tests/test_project.py
```diff
@@ -2317,6 +2319,210 @@ def test_main_submit(self, monkeypatch):
...
+
+class TestHooksInvalidOption:
+    def test_invalid_hook(self):
+        class A(FlowProject):
+            pass
+
+        with pytest.raises(Exception):
+
+            @A.operation
+            @A.hooks.invalid_option(lambda: None)
```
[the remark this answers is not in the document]
klywang: Oh I see! Thanks for catching this

flow/project.py
b-butler: @klywang and I agreed this should be removed since it allows wacky and technically user facing (no underscored methods needed) behavior.

tests/test_project.py
b-butler: So, I understand now how the test work, but @klywang and I discussed ways to improve test clarity.

1. comment what class level variables/fixtures like `keys` and `operation_name` do.
2. Remove tests from `TestHooksInstallBase`, and move tests that change that `base_no_decorators` does not write `start`, `finish`, etc. job document keys in the `TestHooksBase` class. You may want to make the no decorator operation name a variable to allow you to reuse this for the cmd tests.
3. Make the aforementioned test only test that `base_no_decorators` does not have the decorator level hooks.
4. Rename `keys` to `job_doc_keys` and make it a dictionary with keys `"start", "finish", "success", "fail"`.

b-butler:
I suggest the names `project_hooks` for project level hooks, and `hook` or `add_hook` for decorating operation functions with hooks. @bdice, @csadorf, @klywang what do you think? I think this would prevent easy mistakes in typing `hook` versus `hooks` or vice versa.

klywang:
> I suggest the names `project_hooks` for project level hooks, and `hook` or `add_hook` for decorating operation functions with hooks. @bdice, @csadorf, @klywang what do you think? I think this would prevent easy mistakes in typing `hook` versus `hooks` or vice versa.

I like `project_hooks` -- the name is also more descriptive.

flow/project.py
klywang: @b-butler I just checked, and it turns out `hooks = project.hook[op_func]` returns an empty list rather than `_HookRegistry`'s hook list. In order to access the actual hooks, you need to access an underscored method (i.e.  `project.hook._operation_hooks[foo]['on_start']`). I guess that's a little safer since you can't do `project.hook[op_func].on_start.append(foo)` without error. However, it's still possible to do something like `project.hook[op_func]['on_start'].append(foo)`. 

Also, other attribute for `_FlowProjectClass`, like `pre`, `post`, or `operation`, are set using static methods  (lines 1248 - 1388, static methods with naming like `_setup_*`). For the sake of consistency, should we move `_HooksRegistry` to a new static method in `_FlowProjectClass`?

b-butler: This shouldn't be the case since `_HookRegistry.__getitem__` just returns the internal `_operation_func` value. Regardless, I don't think this magic method is necessary, and is better of removed.

changelog.txt
```diff
@@ -5,6 +5,23 @@ Changelog
...
++++++
+
+- Feature to install execution hooks for the automated execution of functions with operations (#28, #189, #508).
+
+Changed
++++++++
+
+Removed
++++++++
+
```
bdice, line 24: We add changelog sections once they have entries.
```suggestion
Changed
+++++++

Removed
+++++++

```

flow/hooks/__init__.py
```diff
@@ -0,0 +1,110 @@
+# Copyright (c) 2018 The Regents of the University of Michigan
```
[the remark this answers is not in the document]
bdice: I agree with the idea that `__init__.py` should be clean if possible. If we don't have the other hooks from #189, this could just live in `hooks.py` in the root of the package (without a subdirectory). However, it would be good to separate hook implementations into their own files in the `hooks/` subdirectory.

flow/project.py
```diff
@@ -267,6 +268,31 @@ def _make_bundles(operations, size=None):
...
+                def __init__(install_self, hook_func):
+                    install_self.hook_func = hook_func
+
+                def __call__(install_self, func):
+                    self._operation_hooks[func][name].append(install_self.hook_func)
+                    return func
+
+            return _InstallHook
+        else:
+            raise AttributeError
```
bdice, line 293: I think this needs to raise an instance of `AttributeError` with a message, not the class itself.

klywang:
@csadorf @bdice @b-butler 

This is read to review again. Some more major changes include:
1) Moved class definitions out of the `__init__.py` file
2) Moved `_HooksRegister` (formerly `_HooksDecorator`) to a class method in `_FlowProjectClass`
3) Renamed `hook` to `add_hook` to avoid confusion with `hooks`
4) Renamed `_HooksDecorator` to `_HooksRegister`

flow/hooks/__init__.py
```diff
@@ -2,109 +2,6 @@
...
-                for hook_trigger in self._hook_triggers
-            ),
-        )
-
-    def __bool__(self):
-        """Return True if hooks are defined."""
-        return any(
-            getattr(self, hook_trigger, None) for hook_trigger in self._hook_triggers
-        )
+Hooks  # Avoid flake8 error F401 '.hooks.Hooks' imported but unused
```
csadorf, line 7:
```suggestion
__all__ = [
  "Hooks"
]
```
This will place the `Hook` class explicitly into the module's namespace and avoid F401.

flow/project.py
```diff
@@ -1674,6 +1752,11 @@ def _show_template_help_and_exit(self, template_environment, context):
         )
         sys.exit(2)
 
+    @property
+    def hooks(self):
+        """:class:`.hooks.Hooks` is defined for this project's operations."""
```
b-butler, line 1757:
```suggestion
        """:class:`.hooks.Hooks` hooks defined for all project operations."""
```

tests/test_project.py
```diff
@@ -2317,6 +2318,236 @@ def test_main_submit(self, monkeypatch):
...
+        with pytest.raises(AttributeError):
+            InstallInvalidHook().install_hook(self.mock_project())
+
+    def test_raise_exception_in_hook(self):
+        job = self.mock_project().open_job(dict(raise_exception=False))
+
+        error_output = self.call_subcmd(f"run -o raise_exception_in_hook -j {job.id}")
+
+        assert "RuntimeError" in error_output
+        assert "Error occurred during execution of "
```
b-butler, line 2538: What is this supposed to assert?

tests/test_project.py
```diff
@@ -2317,6 +2318,236 @@ def test_main_submit(self, monkeypatch):
...
+
+    def test_raise_exception_in_hook_cmd(self):
+        job = self.mock_project().open_job(dict(raise_exception=False))
+
+        error_output = self.call_subcmd(
+            f"run -o raise_exception_in_hook_cmd -j {job.id}"
+        )
+
+        assert "RuntimeError" in error_output
+        assert "Error occurred during execution of "
```
b-butler, line 2548: Same here.

b-butler (commented):
I have a few comments, but I think this is just about ready.

tests/test_project.py
klywang: I removed it. I think the original intention was to make sure the error messages matched up and I didn't catch the wrong error.

klywang:
Oh also, I think this task would be for a separate PR, but should we test hooks compatibility with things like aggregates or groups?

bdice:
> 
> 
> Comments attached. This is the last round of review I can give to this PR, and I think it's ready to merge but cannot approve since I created this PR. I skimmed the tests (I've read them more thoroughly before, but may not be up to date with any changes).
> 
> One other question: what caused the template reference data to change?

@klywang @b-butler @csadorf Friendly ping since it's been a month since activity occurred on this PR. I previously "approved" this PR (but can't actually approve since I opened it).

flow/project.py
```diff
@@ -1507,6 +1512,75 @@ def add_operation_with_directives(function):
...
+                    The function that will be executed at a specified point.
+                trigger: string
+                    The point when a hook operation is executed.
+                """
+                self._hook_func = hook_func
+                self._hook_trigger = trigger
+
+            @classmethod
+            def on_start(cls, hook_func):
+                """Add hook function with trigger "on_start"."""
```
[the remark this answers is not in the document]
bdice: @klywang This specific instance was updated but not the following ones as suggested in the comment "and similarly below."

flow/project.py
```diff
@@ -1508,6 +1513,75 @@ def add_operation_with_directives(function):
...
+                self._hook_trigger = trigger
+
+            @classmethod
+            def on_start(cls, hook_func):
+                """Add a hook function triggered before an operation starts."""
+                return cls(hook_func, trigger="on_start")
+
+            @classmethod
+            def on_finish(cls, hook_func):
+                """Add hook function with trigger "on_finish"."""
```
bdice, line 1557:
```suggestion
                """Add hook function triggered after the operation finishes."""
```

flow/project.py
```diff
@@ -1508,6 +1513,75 @@ def add_operation_with_directives(function):
...
+                return cls(hook_func, trigger="on_finish")
+
+            @classmethod
+            def on_success(cls, hook_func):
+                """Add hook function with trigger "on_success"."""
```
bdice, line 1562:
```suggestion
                """Add a hook function triggered after the operation exits without error."""
```

flow/project.py
```diff
@@ -1508,6 +1513,75 @@ def add_operation_with_directives(function):
...
+                return cls(hook_func, trigger="on_success")
+
+            @classmethod
+            def on_fail(cls, hook_func):
+                """Add hook function with trigger "on_fail"."""
```
bdice, line 1567:
```suggestion
                """Add a hook function triggered after the operation exits with an error."""
```

flow/project.py
```diff
@@ -1508,6 +1513,75 @@ def add_operation_with_directives(function):
...
+                    @FlowProject.operation
+                    @FlowProject.operation_hook.on_start(lambda op_name, job: print(op_name, job))
+                    def foo(job):
+                        pass
+
+                A hook is a function that is called at at specific points during the execution
+                of a job operation. In this example, the anonymous hook function
+                is executed before the operation **foo** runs. Hooks can also run after an
+                operation finish, when an operation exits with error,
+                or when an operation exists without error.
```
bdice, line 1538: Fixed some typos, improved wrapping consistency.
```suggestion
                A hook is a function that is called at at specific points
                during the execution of a job operation. In this example,
                the anonymous hook function is executed before the operation
                **foo** runs. Hooks can also run after an operation finishes,
                when an operation exits with error, or when an operation
                exits without error.
```

bdice (commented):
Final touches, and I approve. @csadorf @b-butler we're seeking a second approval to merge.

csadorf:
@bdice I have acknowledged the request and will try to get to it asap.

tests/define_hooks_test_project.py
```diff
@@ -28,20 +28,20 @@ def raise_error(operation_name, job):
...
 
 @_HooksTestProject.operation
-@_HooksTestProject.add_hook.on_start(set_job_doc(HOOK_KEYS[0]))
-@_HooksTestProject.add_hook.on_finish(set_job_doc(HOOK_KEYS[1]))
-@_HooksTestProject.add_hook.on_success(set_job_doc(HOOK_KEYS[2]))
-@_HooksTestProject.add_hook.on_fail(set_job_doc_with_error())
+@_HooksTestProject.operation_hook.on_start(set_job_doc(HOOK_KEYS[0]))
+@_HooksTestProject.operation_hook.on_finish(set_job_doc(HOOK_KEYS[1]))
+@_HooksTestProject.operation_hook.on_success(set_job_doc(HOOK_KEYS[2]))
+@_HooksTestProject.operation_hook.on_fail(set_job_doc_with_error())
```
b-butler, line 34: Looking at this, I am reminded of our discussion to require an ordering to our decorators, such as a function must be labelled an operation before applying other decorators providing specifics about those operators. Should we go ahead and make these decorators follow that decision? What are your thoughts @bdice? I am inclined to do this since we seem to be moving that way and then we don't have to deprecate the change for these methods later if we do require the ordering.

flow/project.py
```diff
@@ -1193,6 +1194,8 @@ def __new__(metacls, name, bases, namespace):
         cls._OPERATION_FUNCTIONS = []
         cls._OPERATION_PRECONDITIONS = defaultdict(list)
         cls._OPERATION_POSTCONDITIONS = defaultdict(list)
+        cls._OPERATION_HOOK_REGISTRY = defaultdict(lambda: defaultdict(list))
+        cls._operation_hooks = defaultdict(list)
```
b-butler, line 1198: Why is this defined here? This is initialized in `__init__` for the `FlowProject`.
```suggestion
```

tests/test_project.py
```diff
@@ -2313,6 +2314,236 @@ def test_main_submit(self, monkeypatch):
...
+    @pytest.fixture(params=["base"])
+    def operation_name(self, request):
+        return request.param
+
+
+class TestHooksInstallCmdWithDecorators(TestHooksCmd, TestHooksInstallSetUp):
+    # Tests if project-wide hooks interfere with operation level hooks
+    # in job operations with the @cmd decorator
+    @pytest.fixture()
+    def operation_name(self):
```
b-butler, line 2476: You are inconsistent on whether you parametrize single operation fixtures or use a blank signature. Pick one style and use that for consistency. 
```suggestion
    @pytest.fixture
    def operation_name(self):
```

b-butler (changes requested):
We are really close. I have a few comments one which is more of a discussion that requested change.
Also, please re-request a review from me when ready. This escaped my notice since I was not re-requested.

tests/define_hooks_test_project.py
csadorf: I think this falls outside of the scope of this PR.

flow/hooks/hooks.py
```diff
@@ -0,0 +1,104 @@
...
+    --------
+    The example below shows an operation level decorator that prints the
+    operation name and job id at the start of the operation execution.
+
+    .. code-block:: python
+        @FlowProject.operation
+        @FlowProject.hook.on_start(lambda operation_name, job: print(
+            f"Starting operation {operation_name} on job {job.id}"))
+        def op(job):
+            pass
```
csadorf, line 45: This example here is no longer correct and we should probably also add an example for project-level hooks.

csadorf (changes requested):
I discovered some inconsistencies with the documentation, but other than that I believe this is good to go.

Speaking of documentation, https://github.com/glotzerlab/signac-docs/pull/154 is not quite up-to-date, is it?

tests/define_hooks_test_project.py
bdice: I agree that is out of scope. The PR that would introduce a required order #538 has been stalled because of the complexity of enforcing decorator ordering.

flow/project.py
bdice: Fixed. I edited the suggestion above. (Feel free to admin-edit my suggestions in the future for errors like that.)

tests/test_project.py
bdice: @b-butler I don't see any inconsistency from the changeset in this PR. I see that `test_aggregates.py` uses no parentheses here, but everywhere else in signac-flow (including this PR) uses `fixture()` as far as I can tell. I would recommend making a separate PR to fix `test_aggregates.py` if you would like. Let me know if I'm missing something.

flow/hooks/hooks.py
bdice: I fixed this example but realized there are a few missing pieces. I'll write a separate comment.

tests/define_hooks_test_project.py
b-butler: I would disagree #538 is stalled due to complexity, but we can punt on this until later.

csadorf:
>     1. Should `flow.hooks.Hooks` be public? Should it be documented in our API? It is never user-facing as far as I can tell.

No, the properties on the project that users are expected to use should be public and part of the API IMO.

>     2. We need to fix the docs for `FlowProject.operation_hooks`. I noticed it wasn't documented at all so I added it to `api.rst` but it's currently broken (shows up as "alias of `flow.project._FlowProjectClass._setup_hook_object.<locals>._HooksRegister`"). I think this used to be an issue with pre/post so a similar fix may be applicable (I can't remember what I did).

What is your suggestion? Should we try to fix this as part of this PR or not?

bdice:
> > 1. Should `flow.hooks.Hooks` be public? Should it be documented in our API? It is never user-facing as far as I can tell.
> No, the properties on the project that users are expected to use should be public and part of the API IMO.

I agree. I recommend renaming `Hooks` to `_Hooks`.

> > 2. We need to fix the docs for `FlowProject.operation_hooks`. I noticed it wasn't documented at all so I added it to `api.rst` but it's currently broken (shows up as "alias of `flow.project._FlowProjectClass._setup_hook_object.<locals>._HooksRegister`"). I think this used to be an issue with pre/post so a similar fix may be applicable (I can't remember what I did).
> What is your suggestion? Should we try to fix this as part of this PR or not?

I'm fine with fixing the docs in PR or in a subsequent PR -- just wanted to note the issue for right now. I can revisit this again later in the week but would welcome anyone who has time to work on it sooner.

b-butler:
@bdice I couldn't get the documentation quite right but I made it passable I think.

flow/project.py
```diff
@@ -1217,6 +1219,8 @@ def __new__(metacls, name, bases, namespace):
         cls._GROUPS = []
         cls._GROUP_NAMES = set()
 
+        cls.operation_hook = cls._setup_hook_object(parent_class=cls)
```
b-butler, line 1222: It seems to me that it is inconsistent to use the plural for project hooks and the singular for operation hooks.

flow/hooks/hooks.py
```diff
@@ -0,0 +1,108 @@
...
+    The example below shows an operation level decorator that prints the
+    operation name and job id at the start of the operation execution.
+
+    .. code-block:: python
+
+        def start_hook(operation_name, job):
+            print(f"Starting operation {operation_name} on job {job.id}.")
+
+        @FlowProject.operation
+        @FlowProject.operation_hook.on_start(start_hook)
```
b-butler, line 47:
```suggestion
        @FlowProject.operation_hook.on_start(start_hook)
        @FlowProject.operation
```

flow/project.py
```diff
@@ -1508,6 +1512,90 @@ def add_operation_with_directives(function):
...
+            operation name and job id at the start of the operation
+            execution.
+
+            .. code-block:: python
+
+                def start_hook(operation_name, job):
+                    print(f"Starting operation {operation_name} on job {job.id}.")
+
+                @FlowProject.operation
+                @FlowProject.operation_hook.on_start(start_hook)
```
b-butler, line 1531:
```suggestion
                @FlowProject.operation_hook.on_start(start_hook)
                @FlowProject.operation
```

flow/project.py
klywang: I think I addressed this in 594cdbf, but lmk if you had something else in mind

b-butler (approved):
:+1:

klywang (approved):
Looks good to me!

bdice:
I have resolved conflicts with other recently merged PRs (template reference data was updated). @csadorf If you would like to offer a final review, this is ready. Otherwise I think we can merge this PR.

PR #518, opened 2021-05-19
Fix: don't catch SubmitErrors, show to user.

### LINKED ISSUE #517: Submit error not shown to user
### Description
When submitting a job, the .sh template had an extra line at the top that made it so the job looked like it could be submitted but did not actually get queued. There was no error message produced which is why this problem was not caught earlier.

### To reproduce
When the job was submitted, the output said the job was ready for submission but when I checked the queue status, there was nothing queued.

### Error output
sbatch: error: This does not look like a batch script.  The first
sbatch: error: line must start with #! followed by the path to an interpreter.
sbatch: error: For instance: #!/bin/sh

I expected to see this output but I only saw this when I used sbatch directly.

### System configuration

 - Operating System [e.g. macOS]: macOS
 - Version of Python [e.g. 3.7]: 3.8.10
 - Version of signac [e.g. 1.0]: 1.6.0
 - Version of signac-flow: 0.14.0

bdice:
## Description
The cached status update context manager was hiding exceptions from users. I believe that I might have introduced this bug in a previous pull request, because I didn't know that `finally` clauses containing `return` statements would clear exceptions.

I fixed the problem:
- Remove `return` from `finally` clause to ensure that exceptions are raised.
- Add test that SubmitErrors are raised to the user.

Further explanation: https://stackoverflow.com/questions/61264374/using-tryfinally-without-except-never-generates-any-error

## Motivation and Context
Fixes #517.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -5,6 +5,17 @@ Changelog
 The **signac-flow** package follows `semantic versioning <https://semver.org/>`_.
 The numbers in brackets denote the related GitHub issue and/or pull request.
 
+Version 0.15
+============
+
+[0.15.0] -- 2021-xx-xx
+----------------------
+
+Fixed
++++++
+
+- Errors raised during submission were not being shown to users (#517, #518).
+
 Version 0.14
 ============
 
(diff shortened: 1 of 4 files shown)


mikemhenry (approved):
LGTM

bdice:
Merging without a second approval because this fixes problems seen by several users.

PR #531, opened 2021-06-02
Fixed dependency flag

### LINKED ISSUE #530: Wrong dependency flag for GreatLakes (SLURM)
### Description

The `--after` flag submits the wrong arguments for `sbatch` on GreatLakes.

### To reproduce

```bash
>>> python flowproject/project.py submit -o <operation> <other flags> --after <SLURM job id>
Querying scheduler...
Submitting cluster job '<project>/<job id>/<operation>/<some hash>':
 - Group: <operation>(<job id>)
```

This submits correctly with:
```bash
>>> python flowproject/project.py submit -o <operation> <other flags> -- -d afterany:<SLURM job id>
Querying scheduler...
Submitting cluster job '<project>/<job id>/<operation>/<some hash>':
 - Group: <operation>(<job id>)
```
In the [slurm docs](https://slurm.schedmd.com/sbatch.html), it looks like the flag should be `--dependency` or `-d`, but it looks like we're submitting `depend`.

### Error output

```bash
Submission error: Error when calling submission command sbatch <other flags> depend="afterany:20979989":
sbatch: error: Unable to open file depend="afterany:20979989"

Execute with '--show-traceback' or '--debug' to get more information.
```
### System configuration

 - Operating System [e.g. macOS]: CentOS Linux 7 (Core)
 - Version of Python [e.g. 3.7]: 3.9.1
 - Version of signac [e.g. 1.0]: 1.6.0
 - Version of signac-flow: 0.13.0

klywang:
## Description
Changed submission flag from `depend="afterany:{job_id}"` to `-d afterany:{job_id}` for SLURM.

Maybe we should make a PR to check other schedulers as well.

Checked on GreatLakes. Tried to submit with one and two job dependencies and it worked.

## Motivation and Context
Fixes #530 

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -28,6 +28,7 @@ Fixed
 +++++
 
 - Errors raised during submission were not being shown to users (#517, #518).
+- Fixed dependency flag for SLURM submissions (#530)
 
 Version 0.14
 ============
--- a/changelog.txt
+++ b/changelog.txt
@@ -30,7 +30,7 @@ Fixed
 +++++
 
 - Errors raised during submission were not being shown to users (#517, #518).
-- Fixed dependency flag for SLURM submissions (#530)
+- Fixed dependency flag for SLURM submissions (#530).
 
 Version 0.14
 ============
--- a/flow/scheduling/slurm.py
+++ b/flow/scheduling/slurm.py
@@ -140,9 +140,7 @@ class SlurmScheduler(Scheduler):
         submit_cmd = self.submit_cmd + flags
 
         if after is not None:
-            submit_cmd.extend(
-                ["-W", '-d afterany:{}'.format(after.split(".")[0])]
-            )
+            submit_cmd.extend(["-W", "-d afterany:{}".format(after.split(".")[0])])
 
         if hold:
             submit_cmd += ["--hold"]
--- a/flow/scheduling/slurm.py
+++ b/flow/scheduling/slurm.py
@@ -140,7 +140,7 @@ class SlurmScheduler(Scheduler):
         submit_cmd = self.submit_cmd + flags
 
         if after is not None:
-            submit_cmd.extend(["-W", "-d", "afterany:{}".format(after)])
+            submit_cmd.extend(["-W", "-d", f"afterany:{after}"])
 
         if hold:
             submit_cmd += ["--hold"]
--- a/flow/scheduling/slurm.py
+++ b/flow/scheduling/slurm.py
@@ -140,7 +140,7 @@ class SlurmScheduler(Scheduler):
         submit_cmd = self.submit_cmd + flags
 
         if after is not None:
-            submit_cmd.extend(["-W", "-d afterany:{}".format(after.split(".")[0])])
+            submit_cmd.extend(["-W", "-d", "afterany:{}".format(after.split(".")[0])])
 
         if hold:
             submit_cmd += ["--hold"]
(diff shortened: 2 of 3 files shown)


flow/scheduling/slurm.py
```diff
@@ -140,9 +140,7 @@ def submit(
         submit_cmd = self.submit_cmd + flags
 
         if after is not None:
-            submit_cmd.extend(
-                ["-W", 'depend="afterany:{}"'.format(after.split(".")[0])]
-            )
+            submit_cmd.extend(["-W", "-d", "afterany:{}".format(after.split(".")[0])])
```
klywang, line 143: Why do we need to split the string when there is a period? From the docs, it looks like a list of job ids is listed like `-d afterany:00000:00001:00002` etc...

vyasr: The `after` argument is largely unused AFAIK and at various points in flow's history it has definitely been outright broken. This logic is >5 years old and has not been changed in that time (I can definitely trace it back to March 2016 at least) and has survived many large changes in flow, so it's possible that this split was necessary at one point. `s.split('.')[0]` is a no-op if the string doesn't contain `'.'`, so it's not causing any harm, but it's probably safe to remove.

b-butler:
@klywang is this ready for review, or are we waiting to fix other schedulers as well.

flow/scheduling/slurm.py
b-butler, line 143: Why not `afterok:{}`?

b-butler (commented):
I am wondering the reason for using `afterany` over `afterok`.

flow/scheduling/slurm.py
bdice: I agree with @vyasr that this is probably legacy code that may not apply here. I'd remove the `split` because it's confusing and probably not helpful.
```suggestion
            submit_cmd.extend(["-W", "-d", "afterany:{}".format(after)])
```

klywang:
> @klywang is this ready for review, or are we waiting to fix other schedulers as well.

@b-butler This is ready for review. We can separate changes to other schedulers in a different pull request.

bdice:
Since #533 is merged, we could add tests for this behavior. The `--after` flag should appear in the output of `submit --pretend`. I'll work on adding a test here.

bdice:
> 
> 
> Since #533 is merged, we could add tests for this behavior. The `--after` flag should appear in the output of `submit --pretend`. I'll work on adding a test here.

Done in c0a6748. This should be ready for review, and we should have a lot more confidence in the code now. 😄

Nipuli-Gunaratne (approved):
I'm still relatively new to coding/signac but I think syntax looks good and the comments are clear.

changelog.txt
```diff
@@ -30,6 +30,7 @@ Fixed
 +++++
 
 - Errors raised during submission were not being shown to users (#517, #518).
+- Fixed dependency flag for SLURM submissions (#530)
```
bdice, line 33: Add period to changelog.
```suggestion
- Fixed dependency flag for SLURM submissions (#530).
```

bdice:
@Nipuli-Gunaratne Thank you! That's the right kind of feedback for a review. It's important that the code is relatively easy to understand for new users because that makes it easier to maintain. 👍

bdice (approved):
I'm going to approve and merge this so that it doesn't delay the release of 0.15. It looks good to me. Thanks again @klywang and @Nipuli-Gunaratne!

PR #533, opened 2021-06-03
Require a scheduler to submit, even with --pretend enabled.

bdice:
## Description
This PR forces all "pretend" submissions to go through a scheduler. Environments without a scheduler such as `StandardEnvironment` are no longer able to submit in pretend mode. It should be possible to achieve similar output using the `run --pretend` functionality (environments without schedulers can still use `run`, just not `submit`).

For testing purposes, the `TestEnvironment` uses `FakeScheduler` which always just prints the submission script. Use the `TestEnvironment` with the environment variable `SIGNAC_FLOW_ENVIRONMENT="TestEnvironment"`.

## Motivation and Context
Currently users can call `python project.py submit --pretend` even when using the `StandardEnvironment`, which renders the base template with the output script. However, this isn't very useful and the current code circumvents the use of the schedulers' support for `pretend=True`, which can provide additional information such as the submission command that would be used without `--pretend`.

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature
- [x] Breaking change<sup>1</sup>

<sup>1</sup>The change breaks (or has the potential to break) existing functionality.

## Checklist:
- [x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac-flow/blob/master/ContributorAgreement.md).
- [x] My name is on the [list of contributors](https://github.com/glotzerlab/signac-flow/blob/master/contributors.yaml).
- [x] My code follows the [code style guideline](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md#code-style) of this project.
- [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.

If necessary:
- [ ] I have updated the API documentation as part of the package doc-strings.
- [ ] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
- [x] I have updated the [changelog](https://github.com/glotzerlab/signac-flow/blob/master/changelog.txt).

PR #534, opened 2021-06-09
Add flow template create command

### LINKED ISSUE #520: Support an easy method for creating custom templates
### Feature description
Talking to @atravitz and @tommy-waltmann, we decided that we would like to lower the barrier to creating custom templates or custom input into templates. We originally discussed using a CLI option; however, after discussing it with @bdice led to the idea of creating a command line subcommand that would create a `templates` directory in the project directory and a `templates/custom.sh` template that extends the detected environment.

A second proposal that came from this discussion is adding a new block in the templates that is empty for all templates to more easily way of adding lines to a template without changing existing functionality. This is also less error-prone since no default template would use this block and a user could not forget to call `{{ super() }}`.

### Proposed solution

Add a subcommand `./project.py template create`.

b-butler:
## Description

Adds a new command for `flow/__main__.py`, `flow template create [name]` where name is an optional string the overrides the environment detection in determining a template to extend.

## Motivation and Context

This is to simplify the process of extending templates for use on different clusters, tailoring submission scripts to specific needs, and other purposes. Resolves #520.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


changelog.txt
```diff
@@ -17,6 +17,7 @@ Added
 - Add support for aggregation (operations acting on multiple jobs) via ``flow.aggregator`` (#464).
 - Add official support for Andes cluster (#500).
 - Decorator for setting directives while registering operation function ``FlowProject.operation.with_directives`` (#309, #502).
+- Add new flow command ``flow template create`` for automatic create of custom templates.
```
bdice, line 20:
```suggestion
- Add new flow command ``flow template create`` for automatic creation of custom templates.
```

flow/__main__.py
```diff
@@ -46,6 +49,43 @@ def main_init(args):
         )
 
 
+def main_template(args):
+    """Print out general information about template subcommand."""
+    print("flow template does nothing. Use of the available subcommands [create,].")
```
bdice, line 54: This feels clunky. I think there's a more argparse-friendly way to show help messages? You might need to do something with the subparser. I can look into this if you don't see a solution.

flow/__main__.py
```diff
@@ -46,6 +49,43 @@ def main_init(args):
...
+
+
+def main_template_create(args):
+    """Create custom template based on a specified or detected environment."""
+    if args.name is None:
+        env = environment.get_environment()
+        if env is environment.StandardEnvironment:
+            extend_template = "base_script.sh"
```
bdice, line 62: Instead of having a hack right here, should we define the template for the `StandardEnvironment` to be `base_script.sh`?

As a separate question, does it make any sense to allow template creation if the specified/detected environment has no scheduler? With #533, is it even possible to generate a script on environments without a scheduler? (Reasons: we removed the `script` command, `submit` can't do anything without a scheduler, and `submit --pretend` will also do nothing after #533.)

flow/__main__.py
```diff
@@ -46,6 +49,43 @@ def main_init(args):
...
+    else:
+        extend_template = args.name
+
+    project = get_project()
+    os.makedirs(project.fn("templates"), exist_ok=True)
+
+    # grab and render custom template
+    jinja_env = jinja2.Environment(loader=jinja2.PackageLoader("flow"))
+    custom_template_template = jinja_env.get_template("custom.sh")
+    new_template = custom_template_template.render(extend_template=extend_template)
```
bdice, line 74: Why is this a "template template"?
```suggestion
    custom_template = jinja_env.get_template("custom.sh")
    new_template = custom_template.render(extend_template=extend_template)
```

flow/__main__.py
```diff
@@ -46,6 +49,43 @@ def main_init(args):
...
+
+    try:
+        with open(project.fn(os.path.join("templates", "script.sh")), "x") as fh:
+            fh.write(new_template)
+    except OSError as error:
+        if error.errno == errno.EEXIST:
+            logger.error(
+                "Error while trying to create custom template. Delete "
+                "'templates/script.sh' first and rerun command."
```
bdice, line 83: Indicate the cause of the error before recommending the solution.
```suggestion
                "The file 'templates/script.sh' already exists. Delete "
                "'templates/script.sh' first and rerun command."
```

flow/templates/custom.sh
```diff
@@ -0,0 +1,9 @@
+{{ "{{% extends " + extend_template + " %}}" }}
```
bdice, line 1: I would recommend a more descriptive file name, like `custom_script_template.sh`. Maybe we should change the terminology of this PR to call it a "user environment template"?

This needs a nice comment block to explain what it's doing.
```suggestion
{#
This is a jinja template used to generate another jinja template.
This file is used by `flow template create` to detect and create a new script
template that can be modified by the user to add custom commands (e.g.
`module load ...`).
#}
{{ "{{% extends " + extend_template + " %}}" }}
```

bdice (changes requested):
Very cool concept, and implementation is pretty good. I have some comments below. The only other major need is for a test. I can look at adding tests for the existing `flow` commands in a separate PR, since there's no tests for the CLI right now. Maybe I can start that during office hours tomorrow?

flow/__main__.py
bdice: Maaaaybe. I'm not sure. I'm inclined to warn the user if `env.scheduler_type is None` with a message like:
> `"No scheduler is defined for this environment. A template can be generated, but a scheduler will be required for submission."`

This special case for `StandardEnvironment` doesn't do anything as far as I can tell. Currently the code behaves as:
```python
>>> import flow
>>> flow.environment.StandardEnvironment.template
'base_script.sh`
```

flow/__main__.py
```diff
@@ -46,6 +49,43 @@ def main_init(args):
...
+        with open(project.fn(os.path.join("templates", "script.sh")), "x") as fh:
+            fh.write(new_template)
+    except OSError as error:
+        if error.errno == errno.EEXIST:
+            logger.error(
+                "The file 'templates/script.sh' already exists. Delete "
+                "'templates/script.sh' first and rerun command."
+            )
+        else:
+            logger.error(f"Error while trying to create custom template: '{error}'.")
```
bdice, line 86: Show some message if the `try:` succeeds, to indicate success and the destination file.
```suggestion
            logger.error(f"Error while trying to create custom template: '{error}'.")
    else:
        print(f"Created user script template in '{template_path}', extending the template '{env.template}'.")
```

flow/__main__.py
```diff
@@ -46,6 +49,43 @@ def main_init(args):
...
+
+    try:
+        with open(project.fn(os.path.join("templates", "script.sh")), "x") as fh:
+            fh.write(new_template)
+    except OSError as error:
+        if error.errno == errno.EEXIST:
+            logger.error(
+                "The file 'templates/script.sh' already exists. Delete "
+                "'templates/script.sh' first and rerun command."
+            )
```
bdice, line 84: Show full paths and also don't hardcode the path in more than one place.
```suggestion
    try:
        template_path = project.fn(os.path.join("templates", "script.sh"))
        with open(template_path, "x") as fh:
            fh.write(new_template)
    except OSError as error:
        if error.errno == errno.EEXIST:
            logger.error(
                f"The file '{template_path}' already exists. Delete "
                f"'{template_path}' first and rerun this command."
            )
```

bdice (changes requested):

flow/__main__.py
bdice: Actually, we should use the same logic as this snippet. The template directory is configurable.
https://github.com/glotzerlab/signac-flow/blob/8487b3aac55966196857ddbd08776f78fe70334b/flow/project.py#L1579-L1584

changelog.txt
csadorf:
```suggestion
- Add new flow command ``flow template create`` to automatically create custom templates.
```

flow/__main__.py
b-butler: Seems like it is not, I didn't see it defined in `StandardEnvironment` (it is set in `ComputeEnvironment`), so I will remove it.

bdice:
I'd like to merge #539 first and then re-evaluate this PR once CLI tests are added.

flow/__main__.py
```diff
@@ -89,12 +131,53 @@ def main():
         help="Specify a specific to template to use.",
     )
 
+    # the flow template command
+    parser_template = subparsers.add_parser(
+        "template", help="Create and manage custom user templates."
+    )
+    # Set a flag to indicate that flow template was called without a subparser.
+    parser_template.set_defaults(bare_template=True)
```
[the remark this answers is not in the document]
b-butler: I could see `template list` potentially. I am not sure how many more we would have, but in the case we do add more this does allow us to easily do so. I do get though that we are being somewhat hacky with using `set_defaults` in this way.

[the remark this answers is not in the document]
bdice: Cool. Just wanted your thoughts on that, I also considered `list`. I am fine with the current implementation, setting a default variable is the same approach we use in other CLI code.

jennyfothergill (commented):
This is a great idea! 
I tried
```
flow template create -e bridges2
```
```
flow template create -e bridges2.sh
```
```
flow template create -e nonsense.sh
```
and all worked. Will the extends command in the script work with bridges2 and bridges2.sh? (Maybe there should be a note in the help message about whether user should include the .sh if it matters.) Also, do you want to add a check that the template exists?

Also if I don't provide a name, the script saves in `templates/s`--not `script.sh` as the argument help message suggests. (It might be because script.sh already exists in my templates folder?) And if the `templates/s` file already exists, it prints an error message to delete that file. It could be kind of cool if it incremented a file number (e.g., script1.sh, script2.sh) instead? Maybe not--I'm not sure why/if a user would need multiple templates.

Also, providing a name ```flow template create -n heck.sh``` worked as expected.

flow/__main__.py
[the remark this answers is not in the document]
jennyfothergill: I agree a `template list` command would be useful.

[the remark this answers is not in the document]
vyasr: I'm not sure why this default variable is necessary. Without it, don't you get get a (admittedly slightly less informative) error message? It should be the difference between
```
# What I think you get with your version
usage: flow template [arg list]

DESCRIPTION
```
and
```
# What you normally get, which is enough IMO
usage: flow template [arg list]
```
If I'm not wrong, I don't think the extra description is worth this hackery.

vyasr:
@b-butler now that #539 is merged can you update this PR and add tests? Then I think we'll be ready for the next round of review.

flow/__main__.py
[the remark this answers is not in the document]
b-butler: @vyasr this is the base output which does not mention create at all,
`usage: __main__.py [-h] [--debug] [--version] {init,template} ...`. A user would have to call with `--help` for any information on using the command. I am fine with removing it just let me know what you prefer.

b-butler:
@jennyfothergill it is intended behavior to allow any string specification of `--extends`.

jennyfothergill (approved):
All my previous review comments have been resolved--LGTM :)

flow/__main__.py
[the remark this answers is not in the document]
vyasr: I think that's the expected behavior for most command line tools on Unix, a complete help is only printed if requested otherwise you just get a basic usage message. It's certainly not universally followed, but it's the most common, e.g.
```
(main) it-HP-Z8-G4-Workstation% less
Missing filename ("less --help" for help)
```
However, what I think would be more in line with standard help while also being useful is giving some idea of subparser arguments:
```
(main) it-HP-Z8-G4-Workstation% flow template
usage: flow [-h] [--debug] [--version] {init,template} ...
```
What we should see here is basic help for the subparser. For example, here's what signac gives:
```
(main) it-HP-Z8-G4-Workstation% signac move
usage: signac move [-h] project job_id [job_id ...]
signac move: error: the following arguments are required: project, job_id
```
If you can modify the subparsers to provide that kind of information (in this case, the `create` argument for `flow template`), that would be fantastic.

tests/test_shell.py
```diff
@@ -86,3 +86,33 @@ def test_init_flowproject_template(self, template):
...
+        self.call("python -m flow init".split())
+        self.call("python -m flow template create".split())
+        with pytest.raises(ExitCodeError):
+            self.call("python -m flow template create".split())
+
+    @pytest.mark.parametrize("name_arg", ("-n foo.sh", "--name foo.sh"))
+    def test_template_create_custom_name(self, name_arg):
+        self.call("python -m flow init".split())
+        self.call(("python -m flow template create " + name_arg).split())
+        assert os.path.exists("templates/foo.sh")
```
bdice, line 109: Verify the file contents. Something like what's above:
```suggestion
        assert os.path.exists("templates/foo.sh")
        with open("templates/foo.sh") as fh:
            custom_script_lines = fh.read().splitlines()
        assert f'{% extends "{template}" %}' in custom_script_lines[0]
```

tests/test_shell.py
```diff
@@ -86,3 +86,33 @@ def test_init_flowproject_template(self, template):
         self.call(f"python -m flow init -t {template}".split())
         assert str(signac.get_project()) == "project"
         assert os.path.exists("project.py")
+
+    def test_template_create_base(self):
+        self.call("python -m flow init".split())
+        self.call("python -m flow template create".split())
+        assert os.path.exists("templates/script.sh")
+        template = flow.environment.get_environment().template
```
bdice, line 94: I think `get_environment` is exposed as a top-level function, so let's use that.

```suggestion
        template = flow.get_environment().template
```

Also I would prefer to force this test to use a provided environment rather than actually detecting the environment, to ensure that tests pass reliably from any system. You have to do this in a way that will affect both the current Python environment _and_ subprocesses (`python -m flow ...`). I recommend a monkeypatch that sets an environment variable. Search `SIGNAC_FLOW_ENVIRONMENT` in the tests for an example, I just used this same trick for PR #533.

tests/test_shell.py
```diff
@@ -86,3 +86,33 @@ def test_init_flowproject_template(self, template):
...
+        with open("templates/script.sh") as fh:
+            custom_script_lines = fh.read().splitlines()
+        assert ('{% extends "' + template + '" %}') in custom_script_lines[0]
+
+    def test_template_create_fail_on_recall(self):
+        self.call("python -m flow init".split())
+        self.call("python -m flow template create".split())
```
bdice, line 101: Validate that the first step succeeded.
```suggestion
        self.call("python -m flow template create".split())
        assert os.path.exists("templates/script.sh")
```

tests/test_shell.py
```diff
@@ -86,3 +86,33 @@ def test_init_flowproject_template(self, template):
         self.call(f"python -m flow init -t {template}".split())
         assert str(signac.get_project()) == "project"
         assert os.path.exists("project.py")
+
+    def test_template_create_base(self):
```
bdice, line 90: Add some tests to verify that the `--help` message is printed.

In most of the tests here, if something is written to stdout you should verify that by capturing the output, like in this sketch:
```python
out = self.call("python...")
assert "Template created!" in out
```

tests/test_shell.py
```diff
@@ -86,3 +86,33 @@ def test_init_flowproject_template(self, template):
...
+    def test_template_create_fail_on_recall(self):
+        self.call("python -m flow init".split())
+        self.call("python -m flow template create".split())
+        with pytest.raises(ExitCodeError):
+            self.call("python -m flow template create".split())
+
+    @pytest.mark.parametrize("name_arg", ("-n foo.sh", "--name foo.sh"))
+    def test_template_create_custom_name(self, name_arg):
+        self.call("python -m flow init".split())
+        self.call(("python -m flow template create " + name_arg).split())
```
bdice, line 108:
```suggestion
        self.call(f"python -m flow template create {name_arg}".split())
```

tests/test_shell.py
```diff
@@ -86,3 +86,33 @@ def test_init_flowproject_template(self, template):
...
+        assert os.path.exists("templates/foo.sh")
+
+    @pytest.mark.parametrize("extends_arg", ("-e slurm.sh", "--extends slurm.sh"))
+    def test_template_create_custom_exends(self, extends_arg):
+        self.call("python -m flow init".split())
+        self.call(("python -m flow template create " + extends_arg).split())
```
bdice, line 114:
```suggestion
        self.call(f"python -m flow template create {extends_arg}".split())
```

tests/test_shell.py
```diff
@@ -86,3 +86,33 @@ def test_init_flowproject_template(self, template):
...
+            self.call("python -m flow template create".split())
+
+    @pytest.mark.parametrize("name_arg", ("-n foo.sh", "--name foo.sh"))
+    def test_template_create_custom_name(self, name_arg):
+        self.call("python -m flow init".split())
+        self.call(("python -m flow template create " + name_arg).split())
+        assert os.path.exists("templates/foo.sh")
+
+    @pytest.mark.parametrize("extends_arg", ("-e slurm.sh", "--extends slurm.sh"))
+    def test_template_create_custom_exends(self, extends_arg):
```
bdice, line 112:
```suggestion
    def test_template_create_custom_extends(self, extends_arg):
```

tests/test_shell.py
```diff
@@ -86,3 +86,33 @@ def test_init_flowproject_template(self, template):
...
+        self.call("python -m flow init".split())
+        self.call(("python -m flow template create " + extends_arg).split())
+        assert os.path.exists("templates/script.sh")
+        with open("templates/script.sh") as fh:
+            custom_script_lines = fh.read().splitlines()
+        assert ('{% extends "slurm.sh" %}') in custom_script_lines[0]
```
bdice, line 118: I'm a bit surprised that the extra parentheses passed our style formatters...
```suggestion
        assert '{% extends "slurm.sh" %}' in custom_script_lines[0]
```

tests/test_shell.py
```diff
@@ -86,3 +86,33 @@ def test_init_flowproject_template(self, template):
...
         assert os.path.exists("project.py")
+
+    def test_template_create_base(self):
+        self.call("python -m flow init".split())
+        self.call("python -m flow template create".split())
+        assert os.path.exists("templates/script.sh")
+        template = flow.environment.get_environment().template
+        with open("templates/script.sh") as fh:
+            custom_script_lines = fh.read().splitlines()
+        assert ('{% extends "' + template + '" %}') in custom_script_lines[0]
```
bdice, line 97:
```suggestion
        assert f'{% extends "{template}" %}' in custom_script_lines[0]
```

flow/templates/custom_script_template.sh
```diff
@@ -0,0 +1,15 @@
...
+This is a jinja template used to generate another jinja template.
+This file is used by `flow template create` to detect the current environment
+and create a new script template that can be modified by the user to add
+custom commands (e.g. `module load ...`).
+#}
+{{ '{% extends "' + extend_template + '" %}' }}
+{% raw %}
+{% block header %}
+    {{- super () -}}
+{% endblock header %}
```
bdice, line 11: I wonder if there are other blocks that we should add in this template, with recommendations for usage? Like maybe `preamble` or `project_header`? Not sure where the best place is for a `module load` command, but `module load` is the specific use case I would try to simplify.

Maybe we could add a comment block `{# ... #}` that is in the "raw" section and give instructions for how to use the template effectively.

flow/__main__.py
```diff
@@ -46,6 +50,46 @@ def main_init(args):
...
+        with open(script_path, "x") as fh:
+            fh.write(new_template)
+    except OSError as error:
+        if error.errno == errno.EEXIST:
+            logger.error(
+                f"Error while trying to create custom template. Delete '{script_path}' "
+                f"first and rerun command."
+            )
+        else:
+            logger.error(f"Error while trying to create custom template: '{error}'.")
```
bdice, line 84: How could we trigger this error in our tests, so we have 100% coverage? 😉 Don't spend a lot of time on this, unless you have an idea of how to trigger the error. Perhaps this would be triggered if the user lacks permissions to create a new file?

flow/__main__.py
```diff
@@ -46,6 +50,46 @@ def main_init(args):
...
+        raise
+    else:
+        print(
```
bdice, line 87: It looks like all the other subparser commands print this kind of information to `sys.stderr`. I'm not sure if I like that, but we might want to be consistent with stdout or stderr among these types of messages.

https://github.com/glotzerlab/signac-flow/blob/5dc0a60f9b4c51f9a48dadc265b4a60a60373c4a/flow/template.py#L119

https://github.com/glotzerlab/signac-flow/blob/5dc0a60f9b4c51f9a48dadc265b4a60a60373c4a/flow/__main__.py#L36-L40

flow/__main__.py
```diff
@@ -90,12 +134,46 @@ def main():
...
+    # the flow template command
+    parser_template = subparsers.add_parser(
+        "template", help="Create and manage custom user templates."
+    )
+    template_subparsers = parser_template.add_subparsers()
+
+    # flow template create command
+    parser_template_create = template_subparsers.add_parser(
+        "create",
+        help="Create a new custom template based on the detected or selected environment",
```
bdice, line 146:
```suggestion
        help="Create a new custom template based on the detected or selected environment.",
```

flow/__main__.py
```diff
@@ -90,12 +134,46 @@ def main():
...
+        "without a scheduler (or an unsupported scheduler) this is 'base_script.sh'.",
+    )
+
+    parser_template_create.add_argument(
+        "--name",
+        "-n",
+        nargs=1,
+        type=str,
+        default="script.sh",
+        help="Optional specify a name to use as the the name of the custom user "
```
bdice, line 165:
```suggestion
        help="Optionally specify a name to use as the the name of the custom user "
```

flow/__main__.py
```diff
@@ -90,12 +134,46 @@ def main():
...
+        "template. Defaults to 'script.sh' which will be used by default in flow "
```
bdice, line 166:
```suggestion
        "template. Defaults to 'script.sh', which will be used by default in flow "
```

flow/__main__.py
```diff
@@ -90,12 +134,46 @@ def main():
...
+        "create",
+        help="Create a new custom template based on the detected or selected environment",
+    )
+    parser_template_create.add_argument(
+        "--extends",
+        "-e",
+        nargs=1,
+        type=str,
+        default=None,
+        help="Specify a template to extend (including the file extension if present). "
```
bdice, line 154:
```suggestion
        help="Optionally specify a template to extend (including the file extension if present). "
```

bdice (changes requested):

tests/test_shell.py
b-butler: You have to escape the `{` then using `{{`, so this suggestion won't work, but I can refactor to use f-strings.

flow/templates/custom_script_template.sh
b-butler: I thought of this, but didn't know what advice to exactly offer, so punted it for now. If we think this needs to be there to start, I can try my hand at adding a message.

flow/__main__.py
b-butler: I think this is a mistake, `stderr` should be used for errors. I would say we should remove output on success if we don't want this in `stdout`.

bdice: Some tools print information to stderr that isn't an error. For example, signac-flow's progress bars are printed to stderr. Some suggest that stderr can be used for "diagnostic output": https://unix.stackexchange.com/questions/331611/do-progress-reports-logging-information-belong-on-stderr-or-stdout

Basically, I don't agree that "`stderr` should [_only_] be used for errors." Either stdout or stderr is fine with me, so long as we are consistent about the choice with the other diagnostic messages linked in my first comment.

b-butler: The thread seems to be 6 one way and half a dozen the other with regard should outputting to stderr require an error exit code, and we fall on different sides of that. In this case, then if we want consistency, I can just switch to stderr.

flow/__main__.py
b-butler: That would work, but I don't know of an easy way to trigger this except to potentially overwrite the `open` function to just raise an `OSError`.

flow/templates/custom_script_template.sh
bdice: I created #545 to track this. I think it would increase the value of the feature but we don't have to solve it right now.

changelog.txt
```diff
@@ -17,6 +17,7 @@ Added
 - Add support for aggregation (operations acting on multiple jobs) via ``flow.aggregator`` (#464, #516, #542).
 - Add official support for Andes cluster (#500).
 - Decorator for setting directives while registering operation function ``FlowProject.operation.with_directives`` (#309, #502).
+- Add new flow command ``flow template create`` for automatic creation of custom templates.
```
bdice, line 20:
```suggestion
- Add new flow command ``flow template create`` for automatic creation of custom templates (#520, #534).
```

bdice (approved):
LGTM! I'll update the changelog and set this to automerge.

PR #495, opened 2021-04-08
Improve jinja template readability.

bdice:
## Description
Improving jinja template readability by adding block names to `{% endblock %}` tags and indenting the logic (but not the script contents). This should help with visually identifying what logical paths are followed in the code. This page from the Jinja docs has more information about whitespace control: https://jinja.palletsprojects.com/en/2.11.x/templates/#whitespace-control

Collaborated with @AKerr9509 and @AlainKadar.

## Motivation and Context
Resolves #461.

Existing tests should be sufficient to catch any major problems in the submission scripts, there are regression tests that ensure the output from before and after this change are the same. Status output templates are not yet tested, see #435.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change<sup>1</sup>

<sup>1</sup>The change breaks (or has the potential to break) existing functionality.

## Checklist:

- [x] I am familiar with the [**Contributing Guidelines**](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**Contributor Agreement**](https://github.com/glotzerlab/signac-flow/blob/master/ContributorAgreement.md).
- [x] My name is on the [list of contributors](https://github.com/glotzerlab/signac-flow/blob/master/contributors.yaml).
- [x] My code follows the [code style guideline](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md#code-style) of this project.
- [x] The changes introduced by this pull request are covered by existing or newly introduced tests.

If necessary:
- [ ] I have updated the API documentation as part of the package doc-strings.
- [ ] I have created a separate pull request to update the [framework documentation](https://docs.signac.io/) on [signac-docs](https://github.com/glotzerlab/signac-docs) and linked it here.
- [x] I have updated the [changelog](https://github.com/glotzerlab/signac-flow/blob/master/changelog.txt).

PR #537, opened 2021-06-10
Expanse template

### LINKED ISSUE #493: Create cluster template for SDSC Expanse
### Feature description

We need to create a cluster template for SDSC Expanse. The hardware is similar to Bridges-2 and it uses SLURM, so that can be a good template. Assigning @cbkerr @b-butler @bdice since they are being granted access to the system.

User guide: https://www.sdsc.edu/support/user_guides/expanse.html
Example PR for Bridges-2: https://github.com/glotzerlab/signac-flow/pull/441

cbkerr:
Add Expanse template

## Motivation and Context

Resolves #493 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


flow/environments/xsede.py
```diff
@@ -215,8 +215,69 @@ def add_args(cls, parser):
...
+    cores_per_node = 128
+    gpus_per_node = 4
+
+    @classmethod
+    def _get_mpi_prefix(cls, operation):
+        if operation.directives["nranks"] == 0:
+            return ""
+        # We must specify the pmi2 option as the SLURM configuration of Expanse uses the
+        # special pmi2 preprocessing. See `man srun` on an Expanse login node.
+        base_prefix = "srun --mpi=pmi2"
```
joaander, line 235: This will work for codes built on expanse with their MPI library, but it does not work with the `glotzerlab-software` containers. Those need to be launched with `mpirun`. I have tried but not succeeded in the past to get `srun` to pass MPI configurations through to singularity.

I haven't tested if codes built on the system also work with `mpirun` or not.

flow/environments/xsede.py
```diff
@@ -215,8 +215,69 @@ def add_args(cls, parser):
...
+            return base_prefix + rank_option
+        cpus_per_task = f"--cpus-per-task={operation.directives['omp_num_threads']} "
+        return base_prefix + cpus_per_task + rank_option
+
+    @classmethod
+    def _get_omp_prefix(cls, operation):
+        # The current user states to run OpenMP jobs with mpirun even when not using MPI
+        base_prefix = super()._get_omp_prefix(operation)
+        if operation.directives["nranks"] == 0:
+            return base_prefix + f"mpirun -np {operation.directives['num_omp_threads']}"
```
joaander, line 249: This seems very strange to me. `mpirun` will launch one process for each rank requested in `-np`. OpenMP programs expect just a single process with the `OMP_NUM_THREADS` environment variable set. Like in the "Basic OpenMP Job" example from the Expanse documentation.

```
#SET the number of openmp threads
export OMP_NUM_THREADS=24

#Run the job
./hello_openmp
```

flow/environments/xsede.py
```diff
@@ -215,8 +215,69 @@ def add_args(cls, parser):
...
+        # We must specify the pmi2 option as the SLURM configuration of Expanse uses the
+        # special pmi2 preprocessing. See `man srun` on an Expanse login node.
+        base_prefix = "srun --mpi=pmi2"
+        rank_option = f"-n {operation.directives['nranks']} "
+        # We have to handle the case with and without OpenMP differently due to the
+        # --cpus-per-task option.
+        if operation.directives["omp_num_threads"] == 0:
+            return base_prefix + rank_option
+        cpus_per_task = f"--cpus-per-task={operation.directives['omp_num_threads']} "
+        return base_prefix + cpus_per_task + rank_option
```
joaander, line 242: Are the MPI and OMP prefixes stacked? If not, this also needs to set the OMP_NUM_THREADS environment variable.

joaander:
@cbkerr @b-butler we now have a deadline to support Expanse with signac-flow. XSEDE XRAC projects will automatically be transferred from Comet to Expanse in "mid-July". Comet will set the filesystem to read only and will not run jobs past July 15.

flow/environments/xsede.py
b-butler: Yes, this is done automatically.

flow/environments/xsede.py
b-butler: I remember thinking this was strange too, but the User documentation doesn't show this any more, so maybe I just hallucinated this, but I removed this.

doc/supported_environments.rst
```diff
@@ -31,6 +31,7 @@ In addition, the package provides specialized submission templates for the follo
    supported_environments/stampede2
    supported_environments/comet
```
bdice, line 32: > Comet will set the filesystem to read only and will not run jobs past July 15.

Because Comet will be removed from service in 3 days, I would like to remove Comet from the supported environments. This can be done in this PR or in a separate PR, but should be done in the near term -- ideally before the next release, if it comes after July 15.

flow/environments/xsede.py
```diff
@@ -215,8 +215,67 @@ def add_args(cls, parser):
...
+    gpus_per_node = 4
+
+    @classmethod
+    def _get_mpi_prefix(cls, operation, parallel):
+        # The Expanse user guide recommends the usage of `srun`. However, this will not
+        # forward the MPI configuration to the container, meaning it cannot be used with
+        # one of flow's most common use cases, HPC with singularity containers. Thus we
+        # are using the default mpi command `mpiexec` with the other recommendations
+        # given by the Expanse user guide. Native MPI programs with Expanse also require
+        # the `--mpi=pmi2` option (or the documentation suggests as such), but this
```
bdice, line 236: In my understanding, the flag `--mpi=pmi2` may depend on the MPI module being used. The docs list "OpenMPI, MVAPICH2, and IntelMPI" and my guess is that `pmi2` corresponds to MVAPICH2. @joaander may know more.
https://www.sdsc.edu/support/user_guides/expanse.html#compiling

flow/environments/xsede.py
```diff
@@ -215,8 +215,67 @@ def add_args(cls, parser):
...
+        # would cause errors for the container.
+        #
+        # This has been tested with conda environments to work.
+        if operation.directives["nranks"] == 0:
+            return ""
+        rank_option = f" -n {operation.directives['nranks']} "
+        # We have to handle the case with and without OpenMP differently due to the
+        # --cpus-per-task option recommended by the Expanse user guide.
```
bdice, line 244: Has this behavior been tested? I believe the `--cpus-per-task` is an option for `srun`, and I do not expect it to affect the behavior if we use `mpiexec` instead of `srun` as the MPI command. It may cause an error.

flow/templates/expanse.sh
```diff
@@ -0,0 +1,54 @@
+{# Templated in accordance with: https://www.sdsc.edu/support/user_guides/expanse.html #}
+{% extends "slurm.sh" %}
+{% block tasks %}
+{% set threshold = 0 if force else 0.9 %}
```
bdice, line 4: Please indent this script for readability. See #495 or the [Bridges2 template](https://github.com/glotzerlab/signac-flow/blob/master/flow/templates/bridges2.sh) for examples.

flow/templates/expanse.sh
```diff
@@ -0,0 +1,54 @@
...
+    {% if 'gpu' in partition and not force %}
+        {% raise "Requesting GPU partition, but no GPUs requested!" %}
+    {% endif %}
+    {% set nn = nn|default(cpu_tasks|calc_num_nodes(128), true) %}
+{% endif %}
+{% if 'gpu' in partition %}
+    {% set gpus_per_node = (gpu_tasks / nn)|round(0, 'ceil')|int %}
+    {% set cpus_per_node = (cpu_tasks / nn)|round(0, 'ceil')|int %}
+    {% if cpus_per_node > gpus_per_node * 5 and not force %}
+        {% raise "Cannot request more than 5 CPUs per GPU." %}
```
bdice, line 24: It appears there are 2x20 = 40 cores in the GPU nodes. That would mean that we should replace this with 10 cores / GPU instead of 5.

image
```suggestion
    {% if cpus_per_node > gpus_per_node * 10 and not force %}
        {% raise "Cannot request more than 10 CPUs per GPU." %}
```

flow/templates/expanse.sh
```diff
@@ -0,0 +1,54 @@
...
+#SBATCH -N {{ nn|check_utilization(gpu_tasks, 4, threshold, 'gpu') }}
+#SBATCH --ntasks-per-node={{ cpus_per_node }}
+#SBATCH --gpus={{ gpu_tasks }}
+{% elif partition == 'gpu-shared' %}
+#SBATCH -N {{ nn|check_utilization(gpu_tasks, 1, threshold, 'GPU') }}
+#SBATCH --ntasks-per-node={{ cpus_per_node }}
+#SBATCH --gpus={{ gpu_tasks }}
+{% elif partition == 'shared' %}
+    {% if nn|default(1, true) > 1 %}
+        {% raise "Cannot request shared with multiple nodes." %}
```
bdice, line 38:
```suggestion
        {% raise "Cannot request shared partition with resources spanning multiple nodes." %}
```

bdice (changes requested):
Thanks for your work on this @b-butler @cbkerr! I have some requested changes below.

flow/templates/expanse.sh
```diff
@@ -0,0 +1,54 @@
...
+    {% if 'gpu' in partition %}
+        {% set gpus_per_node = (gpu_tasks / nn)|round(0, 'ceil')|int %}
+        {% set cpus_per_node = (cpu_tasks / nn)|round(0, 'ceil')|int %}
+        {% if cpus_per_node > gpus_per_node * 10 and not force %}
+            {% raise "Cannot request more than 10 CPUs per GPU." %}
+        {% endif %}
+    {% endif %}
+
+    {% if partition == 'gpu' %}
+#SBATCH -N {{ nn|check_utilization(gpu_tasks, 4, threshold, 'gpu') }}
```
bdice, line 29: We're inconsistent with `'gpu'` vs. `'GPU'` and `'CPU'`. Make these consistent with other scripts.

flow/templates/expanse.sh
```diff
@@ -0,0 +1,54 @@
...
+    {% set cpu_tasks = operations|calc_tasks('np', parallel, force) %}
+    {% set gpu_tasks = operations|calc_tasks('ngpu', parallel, force) %}
+    {% if gpu_tasks %}
+        {% if not ('gpu' in partition or force) %}
+            {% raise "GPU operations require a GPU partition!" %}
+        {% endif %} {# End check for correct partition #}
+        {# GPU nodes have 4 NVIDIA V100-32GB SMX2 #}
+        {% set nn_gpu = gpu_tasks|calc_num_nodes(4) %}
+        {% set nn = nn_gpu %}
+    {% else %} {# if no GPU tasks #}
```
bdice, line 14: Personally I find "endif" comments uninformative and a bit confusing when the template already has indented logic. It only begins the `if` block 7 lines above, which is much less than one screen-full, so it's not like the original condition is out of sight.
```suggestion
    {% else %}
```

flow/templates/expanse.sh
```diff
@@ -0,0 +1,54 @@
+{# Templated in accordance with: https://www.sdsc.edu/support/user_guides/expanse.html #}
+{% extends "slurm.sh" %}
+{% block tasks %}
+    {% set threshold = 0 if force else 0.9 %}
+    {% set cpu_tasks = operations|calc_tasks('np', parallel, force) %}
+    {% set gpu_tasks = operations|calc_tasks('ngpu', parallel, force) %}
+    {% if gpu_tasks %}
+        {% if not ('gpu' in partition or force) %}
+            {% raise "GPU operations require a GPU partition!" %}
+        {% endif %} {# End check for correct partition #}
```
bdice, line 10: Doesn't seem necessary since the `if` is only two lines above.
```suggestion
        {% endif %}
```

bdice (approved):
Some minor comments. Please address and fix the template test data. If this is submitting as expected for MPI and GPU jobs, then I think it's fine to merge.

joaander:
This is looking better. Why are the tests failing though?

bdice:
@cbkerr @b-butler Let's add changelog lines for adding Expanse / removing Comet, and then merge.

changelog.txt
```diff
@@ -15,6 +15,7 @@ Added
 +++++
 
 - Added a ``custom_content`` block to templates for user customization (#553).
+- Added official support for XSEDE Expanse cluster (#537)
```
bdice, line 18:
```suggestion
- Added official support for XSEDE Expanse cluster (#537).
```

changelog.txt
```diff
@@ -26,6 +27,7 @@ Removed
 
 - Removed deprecated environment classes (#556).
 - Removed support for signac < 1.3.0 (#558).
+- Removed support for the XSEDE comet environment ($537)
```
bdice, line 30:
```suggestion
- Removed support for decommissioned XSEDE Comet cluster (#537).
```

bdice (approved):
@cbkerr I fixed the formatting and updated the changelog messages to match what we wrote for Bridges-2 / Bridges in a previous version. I'll merge this.

PR #542, opened 2021-06-11
Refactor aggregates cursors.

### LINKED ISSUE #516: Refactor _AggregatesCursor
### Feature description

This issue exists as a reminder that some refactoring proposals in #464 are incomplete and need to be fixed in a follow-up pull request.

Review the open comments on #464 for more information. The basic idea is that we should have two classes -- one that acts as a `JobsCursor` that returns aggregates and one that acts as an iterator over an aggregate store. There are some more details and questions to resolve in #464.

### LINKED ISSUE #519: Support job/aggregate abbreviations in the -j option
### Feature description
Given the new feature `aggregation`, I noticed that I could not easily grab only the aggregates with running `project.py status -d`. We support selecting operations using `-o` which supports regex. I think we should support regex in job selection on the CLI. This matters much less in Python as filters and for loops generally more useful. This would help without aggregation as well. @bdice mentioned that he frequently likes to grab all jobs that start with some value in order to partition the data space. Furthermore this would allow grabbing single jobs by only a specifying a few characters with something like `^ef9d31`.

### Proposed solution

Just change the way we compute jobs for the `-j` argument.

bdice:
## Description
- Refactor aggregates cursors into two classes.
- Add support for abbreviated aggregate ids.
- Require operations argument to `_run_operations` (no code paths used the default behavior of `None`).

## Motivation and Context
Resolves #516. Resolves #519.

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/flow/aggregates.py
+++ b/flow/aggregates.py
@@ -667,7 +667,7 @@ class _JobAggregateCursor(_AggregatesCursor):
     """
 
     def __init__(self, project, filter=None, doc_filter=None):
-        self._project = project
+        super().__init__(project)
         self._cursor = project.find_jobs(filter, doc_filter)
 
     def __eq__(self, other):
--- a/flow/aggregates.py
+++ b/flow/aggregates.py
@@ -392,7 +392,7 @@ class _AggregateStore(_BaseAggregateStore):
 
     def __init__(self, aggregator, project):
         self._aggregator = aggregator
-        super().__init__(project)
+        self._project = project
 
         # We need to register the aggregates for this instance using the
         # provided project. After registering, we store the aggregates mapped
@@ -663,7 +663,6 @@ class _JobAggregateCursor(_AggregatesCursor):
     """
 
     def __init__(self, project, filter=None, doc_filter=None):
-        super().__init__(project)
         self._cursor = project.find_jobs(filter, doc_filter)
 
     def __eq__(self, other):
--- a/flow/aggregates.py
+++ b/flow/aggregates.py
@@ -611,10 +611,6 @@ class _AggregatesCursor(Collection):
     """
 
     @abstractmethod
-    def __init__(self, project):
-        self._project = project
-
-    @abstractmethod
     def __eq__(self, other):
         pass
 
--- a/flow/aggregates.py
+++ b/flow/aggregates.py
@@ -8,7 +8,8 @@ jobs are grouped before being passed as arguments to the operation. The
 default aggregator produces individual jobs.
 """
 import itertools
-from collections.abc import Iterable, Mapping
+from abc import abstractmethod
+from collections.abc import Collection, Iterable, Mapping
 from hashlib import md5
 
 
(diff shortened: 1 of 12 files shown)


flow/project.py
```diff
@@ -3519,8 +3469,21 @@ def _get_aggregate_from_id(self, id):
                 return aggregate_store[id]
             except KeyError:
                 pass
-        # Raise error as didn't find the id in any of the stored objects
-        raise LookupError(f"Did not find aggregate with id {repr(id)}.")
+        else:
```
kidrahahjo, line 3472: I am not super familiar with this flow, can you please redirect me to someplace where this is mentioned?
```
if condition:
    # do something
# code to compute something
else:
    # do something if condition is False
```

kidrahahjo (commented):
Thank you @bdice for this PR. The abbreviated id feature looks super useful. I just have a few minor comments/questions.

flow/aggregates.py
```diff
@@ -597,3 +598,91 @@ def get_aggregate_id(aggregate):
...
+    """Abstract class defining iterators over aggregates stored in a FlowProject.
+
+    Parameters
+    ----------
+    project : :class:`~.FlowProject`
+        A FlowProject whose jobs are aggregated.
+
+    """
+
+    @abstractmethod
```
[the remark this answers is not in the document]
bdice: We want the `_AggregatesCursor` parent class to define an interface. By deriving from the abstract base class `Collection` and marking the constructor `__init__` and equality operator `__eq__` as abstract, we can create subclasses `_AggregateStoreCursor` and `_JobAggregateCursor` that are required to implement `__init__`, `__eq__`, `__iter__`, `__len__`, and `__contains__`. Specifically we mark the constructor as abstract because we don't want users to be able to construct the base class `_AggregatesCursor` - the constructor behavior should be implemented by the specific cursor classes derived from this abstract base class.

Here's an article explaining the purpose and usage of abstract classes: https://www.geeksforgeeks.org/abstract-classes-in-python/

flow/project.py
bdice: `for:` / `else:` is a syntax allowed by Python that is somewhat uncommon in other languages. Here's an article explaining what it does. https://book.pythontips.com/en/latest/for_-_else.html#for-else

In this specific case, since we don't have any `break` statements in the body of the `for` loop, I can remove this `else:` block and remove one level of indention (the `else:` doesn't do anything special here). That would be cleaner, so I'll make that change.

flow/project.py
```diff
@@ -3519,8 +3469,20 @@ def _get_aggregate_from_id(self, id):
...
-        # Raise error as didn't find the id in any of the stored objects
-        raise LookupError(f"Did not find aggregate with id {repr(id)}.")
+        if check_abbrevations:
+            # No direct match was found, so check for abbreviated ids. Requires
+            # iteration over all elements of all aggregate stores.
+            matches = set()
+            for aggregate_store in self._group_to_aggregate_store.inverse:
+                for full_id in aggregate_store:
+                    if full_id.startswith(id):
+                        matches.add(aggregate_store[full_id])
```
kidrahahjo, line 3479: This implementation is linear in time. I wonder if we can make aggregation search using ids logarithmic by storing aggregate ids in a sorted manner in the aggregate stores? This is surely a tradeoff between time and space, hence, I think this should be a point to discuss in our next office hours.
PS: This is not completely related to this PR, so feel free to merge this one.

tests/test_project.py
```diff
@@ -2116,6 +2125,20 @@ def test_main_run(self):
...
 
+    def test_main_run_abbreviated(self):
+        project = self.mock_project()
+        assert len(project)
+        for job in project:
+            assert not job.doc.get("op2", False)
+            assert not job.doc.get("op3", False)
+
+        # Use an abbreviated aggregate id
+        self.call_subcmd(f"run -o group_agg -j {get_aggregate_id(project)[:-5]}")
```
kidrahahjo, line 2136: I think adding a test for duplicates here might make sense.
```
self.call_subcmd(f"run -o some_operation -j {get_aggregate_id(project)} {get_aggregate_id(project)[:-5]}")
```
We should expect the aggregate operation to run just once.

kidrahahjo, line 2136: I think adding a test for duplicates here might make sense.
```
self.call_subcmd(f"run -o some_operation -j {get_aggregate_id(project)} {get_aggregate_id(project)[:-5]}")
```
We should expect the aggregate operation to run just once.

kidrahahjo (approved):
The changes look good to me, and I think we can merge this. I just have a few more comments, but I am going to approve this PR.

tests/test_project.py
bdice: Done in 67dbb3f2740a8118ea1eef8c97214821d777a409.

flow/aggregates.py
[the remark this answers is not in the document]
vyasr: This doesn't make sense to me. I agree that `_AggregatesCursor` should be an abstract class that can't be instantiated, but that is already automatically true by virtue of inheriting from `Collection`, which declares `__contains__`, `__iter__`, and `__len__` abstract. We also need `__eq__` to be defined by subclasses. Declaring the _constructor_ abstract, on the other hand, prevents a user from defining a subclass with a trivial constructor. While the subclasses that we define happen to require stateful constructors, that doesn't make sense as a requirement.

flow/project.py
vyasr: The optimal (or at least the best standard) solution for this particular use-case would be to implement a trie for the aggregate stores, but that is definitely out of scope of this PR and we'd also need to make sure that it doesn't degrade performance in any other way. It's certainly an interesting idea though, and if you want to brainstorm at office hours that would be cool.

tests/define_aggregate_test_project.py
```diff
@@ -13,6 +13,8 @@ def statepoint_i_even_odd_aggregator(jobs):
 
 def set_all_job_docs(jobs, key, value):
     for job in jobs:
+        if key in job.document and job.document[key] == value:
+            raise RuntimeError(f"Key {key} is already set to {value}!")
```
vyasr, line 17:
```suggestion
        if job.document.get(key) == value:
            raise RuntimeError(f"Key {key} is already set to {value}!")
```

vyasr (approved):
This is exactly what I had in mind. I made a couple of comments that should be addressed (no need for an abstract constructor and one minor change to a test function) but unless you disagree with my suggestions I'm approving conditional on those changes being made. Thanks for taking care of this!

flow/aggregates.py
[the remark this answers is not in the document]
bdice: Okay. I removed the abstract constructor in 51f9aef31d956bfc7a26e25e559d7bd3c18156a5.

tests/define_aggregate_test_project.py
bdice: This refactor isn't quite the same, because it would not work if `value is None`. That was why I wrote it like this in the first place. However, since we aren't using `None` as a value and this is a bit simpler, I will apply this change.

PR #544, opened 2021-06-19
Raise error for groups that decorate non-operation functions

### LINKED ISSUE #538: Error for groups that decorate non-operation functions
### Description

I created a group like this and then assigned it to a function but forgot to mark the function as an operation. Like this:

```python
from flow import FlowProject

class MyProject(FlowProject):
    pass

my_group = MyProject.make_group("my_group")

@my_group
def my_operation(job):
    print(job)
```

We agreed during my live demo in office hours that this should probably throw an error.

One way to do this might be to track how many times the group is used to decorate a function, and then make sure that it has the same number of operations on FlowProject instantiation, during group registration.

### System configuration

 - Version of signac-flow: 0.14

kidrahahjo:
## Description

Throw an error when a group decorates a non-operation function
Example
```
from flow import FlowProject

group = FlowProject.make_group("foo")

@group
def op1(job): # This function is not an operation function because FlowProject.operation decorator is not being used
    pass

if __name__ == "__main__":
    FlowProject()
```

## Motivation and Context

Fix #538 

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


bdice (commented):
I revised the error message text and pushed a change. I had a question about the `__name__` attribute.

Also we should consider whether invalid `FlowProject` definitions should raise a `ValueError` or a `RuntimeError`, I put a message in Slack about that and copied it below.

> FlowProject definition errors: Should we prefer `RuntimeError` or `ValueError`? Current behavior:
> 
> `ValueError`s:
> - Attempting to put more than one `@FlowProject.operation` decorator on the same function
> - Attempting to define another operation with the same name as as existing operation
> - Registering a condition function as an operation
> - Registering an operation as a condition
> - Defining a single function as an operation for two projects `A` and `B`, where `B` inherits from `A`
> - Using `FlowProject.pre.after` with a function isn't an operation
> - Creating an operation with the same name as an existing group
> - Creating a group with the same name as an existing operation
> - (Proposed in this PR): Registering a group on a function that is not an operation
> 
> `RuntimeError`s:
> - Using `@cmd` before `@with_job`
> - Using `@aggregator` and `@with_job` on the same operation (in either order)

bdice:
This code currently fails when used with custom operation names.

```python
from flow import FlowProject


class Project(FlowProject):
    pass


group = Project.make_group("group")


@group
@Project.operation("custom_operation_name")
def print_job(job):
    print(job)


if __name__ == "__main__":
    Project().main()
```

Raises error:
```
ValueError: Cannot add the function 'print_job' to a FlowGroup because it is not registered as an operation. Add the `@FlowProject.operation` decorator to register the operation.
```

flow/project.py
```diff
@@ -680,6 +684,7 @@ def __call__(self, func):
             func._flow_groups.append(self.name)
         else:
             func._flow_groups = [self.name]
+        self.operations.append(func.__name__)
```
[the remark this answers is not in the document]
vyasr: There are two distinct problems here, both of which have roughly the same solution:
1. Anonymous functions (lambdas) have a `__name__`, it is just implicitly set to "\<lambda\>". The issue of naming conflicts is not unique to anonymous functions, it occurs any time that two operations have the same `__name__`. For example, if a users import operation functions from separate modules they could conflict e.g. `import foo; import bar; ... Project.operation(foo.baz), Project.operation(bar.baz)`. The handle we provide to users to circumvent this issue is naming the operation via that `name` parameter `Project.operation(foo.baz, "foo_baz"); Project.operation(bar.baz, "bar_baz")`, but it falls to users to make that distinction if necessary. Users wishing to use multiple anonymous functions have the same requirement.
2. Objects that actually don't have a `__name__` will fail [here](https://github.com/glotzerlab/signac-flow/blob/master/flow/project.py#L1415). This issue is possible because instances of `functools.partial`, for example, do not have a name: 
```
>>> x = partial(sum, [1, 2, 3])
>>> x.__name__
  File "<stdin>", line 1, in <module>
AttributeError: 'functools.partial' object has no attribute '__name__'
```
I know @b-butler jumped through quite a few hoops in parts of the groups code to enable partials in certain cases. I would be fine requiring users to provide a name in that case as well, probably by wrapping the line I linked above with something like
```
try:
    name = func.__name__
except AttributeError:
    raise ValueError("A name could not be inferred from the passed callable, please provide the ``name`` parameter to ``Project.operation``.")

File: flow/project.py [hunk 2]
```diff
@@ -655,6 +655,10 @@ def __init__(self, name, options="", group_aggregator=None):
         # decorator placement in terms of `@FlowGroupEntry`, `@aggregator`, or
         # `@operation`.
         self.group_aggregator = group_aggregator
+        # We assign operations to a group using a decorator. In order to identify
+        # whether a function is assigned to a group but not as an operation, we need
+        # to store the names of the functions.
+        self.operations = []
```
vyasr, line 661: If we're only using this object to check if an operation name exists, this may as well be a set rather than a list.

vyasr:
> ...
> Edit: I thought for a while and experimented with some code but I don't see a solution for this. If we assume the function is registered as an operation first (`@FlowProject.operation("name")` is below the group decorator) then we could inject an attribute like `_flow_operation_name` and read that instead of `__name__` when the group decorator is applied. But if the decorators are reversed and the group decorator is first, I don't see how we could identify the custom operation name...

My suggestion on the corresponding issue [here](https://github.com/glotzerlab/signac-flow/issues/538#issuecomment-862944062) was also predicated on enforcing an ordering for the decorators. At the time I said that this might not be desirable, but I think we should consider requiring that the `operation` decorator be applied before the `group` decorator. We already have a precedent (although in a significantly less visible context) for the `with_job` and `cmd` decorators. The more validation logic we push to "run-time" rather than "definition-time", the less performant we're likely to be. Moreover, you could argue that conceptually it makes sense to have decorators nested in a consistent order to make the code readable. Too much flexibility (especially when it doesn't actually enable anything new) can also be a bad thing and it obviously hamstrings us internally in some ways.

kidrahahjo:
I don't think it's too much of an effort to add a check for the case which @bdice mentioned. The solution which immediately comes to my mind requires storing the `__name__` property of the operation functions as an instance variable. But as @vyasr said, too much flexibility is not always great and the positioning of the decorators conceptually makes sense too. @bdice let me know what you think?

flow/project.py
kidrahahjo: The `operations` variable here is only used for iterating over.
```
for operation in entry.operations:
  if operation not in self._operations:
      # raise error if not found.
```
The `_operation` variable is an instance variable for the `FlowProject` class which is a set.

vyasr: Got it, I didn't check that code carefully enough. This looks fine then.

bdice:
@kidrahahjo @vyasr I think it is fine to require an order for the decorators.

flow/project.py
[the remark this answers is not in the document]
vyasr: @kidrahahjo I think enforcing this requirement explicitly is the easiest way for you to move forward.

vyasr:
I think the path forward then is to require an order for the decorators and then just look up the names in `_OPERATION_FUNCTIONS`.

bdice:
I'm taking a look at this PR and will try to finalize it tonight. @b-butler said he and @kidrahahjo are planning to release 0.16 tomorrow.

flow/project.py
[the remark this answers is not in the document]
bdice: I added tests that raise `FlowProjectDefinitionError` for anonymous functions added to groups without first being added as operations. Resolving.

flow/project.py
bdice: I think you might be saying different things in this comment thread. It sounds like a `set` is actually a viable replacement here, and is already how the `FlowProject` tracks its operations. I carefully reviewed the code and it appears that all the uses of `FlowOperation._flow_groups` and `FlowGroupEntry._operations` can be replaced by `set` instead of `list`. I made this change in ee92b196694fa3befc519ae7f0f7a7a126a10d1e.

bdice (approved):
This looks good to me. @b-butler @kidrahahjo Feel free to merge this for the 0.16 release if you agree that it is complete.

bdice:
@b-butler @kidrahahjo I forgot to handle decorator ordering when I worked on this last night, and have retracted my approval. There is still some additional work needed for that, so I suggest that we bump this to a later version. 👍

b-butler:
@bdice sounds good. I will change the milestone for this PR and associated issue then. :smiley:

tests/test_project.py
```diff
@@ -1772,6 +1788,37 @@ def bar(job):
...
+
+        group = A.make_group("foo")
+
+        with pytest.raises(FlowProjectDefinitionError):
+
+            @group
+            def test_op(job):
+                pass
+
+        # Make test_op into an operation, then project creation should succeed.
```
b-butler, line 1805:
```suggestion
        # Make test_op into an operation, then group adding should succeed.
```

flow/project.py
```diff
@@ -1455,11 +1466,13 @@ def __call__(self, func, name=None):
...
-                self._parent_class._GROUPS.append(FlowGroupEntry(name=name, options=""))
+                self._parent_class._GROUPS.append(
+                    FlowGroupEntry(name=name, project=self._parent_class, options="")
+                )
                 if hasattr(func, "_flow_groups"):
-                    func._flow_groups.append(name)
+                    func._flow_groups.add(name)
                 else:
-                    func._flow_groups = [name]
+                    func._flow_groups = {name}
```
b-butler, line 1475: It should always be the case now that `_flow_groups` does not exist. If it does exist, it should be an error, unless we are allowing different parent classes to also declare an operation without error.

flow/project.py
```diff
@@ -675,14 +678,22 @@ def __call__(self, func):
...
         """
+        if not any(
+            func == op_func for _, op_func in self._project._OPERATION_FUNCTIONS
+        ):
+            raise FlowProjectDefinitionError(
+                f"Cannot add function '{func}' to group without making the function an "
+                f"operation. Add @MyProjectClass.operation below group decorator."
+            )
+
         if hasattr(func, "_flow_groups"):
```
b-butler, line 689: Since `FlowProject.operation` creates a `_flow_groups` attribute we should just check for the existence of that and error if it is not found. Although, if it passes the operation test, then the attribute should always exist.

flow/project.py
b-butler: We should still keep the above check though because another project could set the `_flow_groups` attribute.

b-butler (changes requested):
This is looking good. A few more changes and this should be ready.

This could also be considered a breaking change, so what are thoughts on merging this without a deprecation cycle? The error message does give the fix which helps mitigate any pain in switching so I am fine with merging.

vyasr:
I'm going to look into the outstanding status error here and see if I can help resolve it so that we can get this merged.

vyasr:
@bdice I don't think there's anything wrong here. If you look at the [test project definition](https://github.com/glotzerlab/signac-flow/pull/544/files#diff-04581b2a24930660df81af15374e1effd73d2d717aa7de2b051c4c5e69a17b76) the old version of `op1` was actually not decorated with the operation decorator. I assume that this was an oversight, but as a result the outputs of the status commands now contain an extra operation which throws off alignment (e.g. in the `eligible_jobs_max_lines=2` case that was the first failure that I saw) and total lines output (in the `unroll=False` case that is failing now after I merged in master). We should be safe to update the reference data for status tests.

flow/project.py
vyasr: @b-butler I'm confused, is there an action item here? I think after your second comment that the answer is no.

flow/project.py
vyasr: Are you saying that if `_OPERATION_FUNCTIONS` is populated then we must have also populated `_flow_groups`? If so, I agree. What would you like to see here now, then?

vyasr:
> This is looking good. A few more changes and this should be ready.
> 
> This could also be considered a breaking change, so what are thoughts on merging this without a deprecation cycle? The error message does give the fix which helps mitigate any pain in switching so I am fine with merging.

Yeah, I'm fine just merging this. To an extent this is almost like a bug fix, it's resolving some undesirable edge cases in behavior.

flow/project.py
b-butler: I think we should error if `func._flow_groups` exists already.

flow/project.py
b-butler: We could remove the reference to the project that this adds and makes the class slightly more complicated and just check for the existence of `_flow_groups` and error if not found.

vyasr:
@b-butler let me know if you're happy with the way that I addressed your comments, I had to make some adaptations to deal with inherited operations between projects and to avoid conflicts.

flow/project.py
```diff
@@ -1469,10 +1466,10 @@ def __call__(self, func, name=None):
...
                     FlowGroupEntry(name=name, project=self._parent_class, options="")
                 )
-                if hasattr(func, "_flow_groups"):
-                    func._flow_groups.add(name)
-                else:
-                    func._flow_groups = {name}
+                if not hasattr(func, "_flow_groups"):
+                    func._flow_groups = {}
+                if self._parent_class not in func._flow_groups:
+                    func._flow_groups[self._parent_class] = {name}
```
b-butler, line 1472: This check should be unnecessary as if this is true then this function is already in another group with the same project which shouldn't happen.
```suggestion
                func._flow_groups[self._parent_class] = {name}
```

b-butler (changes requested):
I have one suggestion and I will approve.

b-butler (approved):
:+1:

PR #550, opened 2021-06-30
Fix --job-output command for all schedulers.

bdice:
## Description
User @salazardetroya noticed that the `--job-output` command wasn't supported by most environments. This PR makes it a universal feature for all schedulers.

*Reviewers:* should we split this up and have two commands for output and error instead of always combining them?

## Motivation and Context
Adds support for custom `--job-output` values.

## Types of Changes
- [ ] Documentation update
- [x] Bug fix
- [x] New feature

## Checklist:

If necessary:


vyasr:
Part of the reason that we've punted on improving support for this feature in the past is that there are a couple gotchas to be aware of that should at least be documented:
- If users provide a relative path, what root directory is it relative to? Do we make any promises that it will always be relative to the project root? If not, are we exclusively recommending absolute paths? What should users who want to put outputs in the job directories do, and how could they construct appropriate absolute paths (if necessary)?
- How does this work with groups/bundles/aggregates?

I'm not a reviewer on this PR so I'll leave it to others to make final decisions, but answers to this question (at least in the form of documentation) are definitely in scope for this PR since adding this feature as is will likely lead to confusion without some attempt at providing an explanation.

I don't have a strong opinion on the separate vs. combined filename question. I would be fine with merging this PR without having that addressed, but IMO it's fine to just combine the streams for now.

atravitz (approved):
I think it's fine to leave things combined for now

Charlottez112 (approved):

bdice:
@vyasr I'm not sure how you want to act on these answers to your questions. What else do you think needs to be done on this PR, if anything?

> If users provide a relative path, what root directory is it relative to? Do we make any promises that it will always be relative to the project root? If not, are we exclusively recommending absolute paths? What should users who want to put outputs in the job directories do, and how could they construct appropriate absolute paths (if necessary)?

This just sets the scheduler flags for output / error files. How the scheduler interprets paths is not something we have to worry about. I would guess that relatives paths are relative to the directory from which the submission command is executed.

> How does this work with groups/bundles/aggregates?

It doesn't. This is not job-specific. It just sets a scheduler flag for all the scripts being generated.

tcmoore3:
Are there more changes that need to be made on this?

bdice:
@tcmoore3 Nope, it's ready! Just lost track of this PR during my defense. Thanks for the reminder, I set it to merge once tests pass. 😄

PR #558, opened 2021-07-29
Use id property, require signac>=1.3.0.

bdice:
## Description
This PR updates all usage of `job.get_id()` to the property `job.id`.

Changes in the API of signac for version 2.0 will remove the `job.get_id()` and `project.get_id()` methods in favor of the properties `job.id` and `project.id`. This PR for signac-flow updates its use of these ids, and should allow seamless compatibility for signac>=1.3.0, including major version 2.

Additionally, this PR adds nightly CI testing for the `next` branch of signac, which will become version 2.0.

## Motivation and Context
See comment: https://github.com/glotzerlab/signac/pull/578#issuecomment-886080077

## Types of Changes
- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ from setuptools import find_packages, setup
 
 requirements = [
     # The core package.
-    "signac>=1.0.0",
+    "signac>=1.3.0",
     # For the templated generation of (submission) scripts.
     "jinja2>=2.10",
     # To enable the parallelized execution of operations across processes.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,7 +108,7 @@ jobs:
     <<: *test-template
     environment:
       DEPENDENCIES: "OLDEST"
-      SIGNAC_VERSION: "signac==1.0.0"
+      SIGNAC_VERSION: "signac==1.3.0"
     docker:
       - image: circleci/python:3.6
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,21 +81,26 @@ jobs:
           path: test-reports
           destination: test-reports
 
+  linux-python-39-signac-16:
+    <<: *test-template
+    environment:
+      DEPENDENCIES: "NEWEST"
+      SIGNAC_VERSION: "signac~=1.6.0"
   linux-python-39-signac-15:
     <<: *test-template
     environment:
       DEPENDENCIES: "NEWEST"
       SIGNAC_VERSION: "signac~=1.5.0"
-  linux-python-39-signac-14:
-    <<: *test-template
-    environment:
-      DEPENDENCIES: "NEWEST"
-      SIGNAC_VERSION: "signac~=1.4.0"
   linux-python-39-signac-latest:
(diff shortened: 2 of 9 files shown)


setup.py
```diff
@@ -7,7 +7,7 @@
 
 requirements = [
     # The core package.
-    "signac>=1.0.0",
+    "signac>=1.3.0",
```
csadorf, line 10: Is the assumption that this version is going to be fully compatibel with signac 2? Then, I'd suggest:
```suggestion
    "signac~=1.3,~=2.0",
```
Otherwise:
```suggestion
    "signac~=1.3",
```

csadorf (commented):
Since signac (core) is _mostly_ following semantic versioning, I'd recommend a slight change in the dependency specification.

setup.py
bdice: I currently expect it to be compatible with signac 2.0, yes. However, there are a few catches to discuss.

First, I think the proposed syntax `~=1.3,~=2.0` may be invalid. [PEP 440](https://www.python.org/dev/peps/pep-0440/#version-specifiers) says that "The comma (",") is equivalent to a logical **and** operator", which seems that it would break `~=1.3,~=2.0`. Based on the definition of a compatible release specifier, `~V.N` is like `>= V.N, == V.*`. No release could match both criteria `~=1.3` and `~=2.0` -- we would want an "or" separating them, if anything.

As an alternative, it might also be equivalent to specify `>=1.3.0,<3.0.0`. However, I recently read a perspective on why blocking a major version like this may pose a significant detriment to users: https://hynek.me/articles/semver-will-not-save-you/

> On the other hand, if a new major version of a package surprisingly breaks your package, they can always add a pin themselves (see step 4 above) until you fix your package. But there’s no practical way for them to remove your pin.
_Don’t ever pin major versions, unless you know they’re broken._ 

Since we won't truly know compatibility with signac 2.0 (or 3.0) until some later point in time when those releases are made, I recommend that we only bump the minimum version requirement for now, from `>=1.0.0` to `>=1.3.0`.

csadorf: You are absolutely, right, my suggestion with `~=1.3,~=2.0` is nonsensical, sorry about that.

I read the article you linked, certainly an interesting read although I don't fully agree with all of its conclusions. That said, I agree that updating our dependency to `>=1.3.0` is likely the best solution.

bdice: Sounds good. I agree -- the article takes a strong stance on some topics, but I think there may be good reasons to pause before blocking major versions. 👍

tommy-waltmann:
Did signac 1.3 introduce the id property? I don't understand why these two seemingly unrelated updates are a part of the same PR.

bdice:
> Did signac 1.3 introduce the id property? I don't understand why these two seemingly unrelated updates are a part of the same PR.

Yes, exactly. The id property was added in signac 1.3.

csadorf (approved):

tommy-waltmann (approved):

PR #562, opened 2021-08-12
Add tests for print_status output

### LINKED ISSUE #435: Add tests for print_status output

Charlottez112:
## Motivation and Context
This pull request resolves #435 

## Description
* Generate status and save in files
* Run tests and check against the status generated

## Checklist:

If necessary:

### DIFF
--- a/tests/define_status_test_project.py
+++ b/tests/define_status_test_project.py
@@ -1,0 +1,92 @@
+import os
+
+import flow
+from flow import FlowProject
+
+
+class _TestProject(FlowProject):
+    pass
+
+
+group1 = _TestProject.make_group(name="group1")
+group2 = _TestProject.make_group(name="group2", options="--num-passes=2")
+
+
+@_TestProject.label
+def default_label(job):
+    return True
+
+
+@_TestProject.label
+def negative_default_label(job):
+    return False
+
+
+@_TestProject.label("named_label")
+def anonymous_label(job):
+    return True
+
+
+@_TestProject.label
+def b_is_even(job):
+    try:
+        return job.sp.b % 2 == 0
+    except AttributeError:
+        return False
+
+
+@_TestProject.operation.with_directives(
+    {
+        # Explicitly set a "bad" directive that is unused by the template.
+        # The submit interface should warn about unused directives.
+        "bad_directive": 0,
+        # But not this one:
+        "np": 1,
+    }
+)
+@flow.cmd
+@_TestProject.pre(b_is_even)
+@group1
+@_TestProject.post.isfile("world.txt")
+def op1(job):
+    return 'echo "hello" > {job.ws}/world.txt'
+
+
+def _need_to_fork(job):
+    return job.doc.get("fork")
+
+
+@_TestProject.operation.with_directives({"fork": _need_to_fork})
+@group1
+@_TestProject.post.true("test")
+def op2(job):
+    job.document.test = os.getpid()
+
+
+@_TestProject.operation.with_directives({"ngpu": 1, "omp_num_threads": 1})
+@group2.with_directives(dict(omp_num_threads=4))
+@_TestProject.post.true("test3_true")
+@_TestProject.post.false("test3_false")
+@_TestProject.post.not_(lambda job: job.doc.test3_false)
+def op3(job):
+    job.document.test3_true = True
+    job.document.test3_false = False
+
+
+# class _DynamicTestProject(_TestProject):
+#     pass
+
+
+# group3 = _DynamicTestProject.make_group(name="group3")
+
+
+# @_DynamicTestProject.operation
+# @group3
+# @_DynamicTestProject.pre.after(op1)
+# @_DynamicTestProject.post(lambda job: job.sp.get("dynamic", False))
+# def op4(job):
+#     job.sp.dynamic = True  # migration during execution
+
+
+if __name__ == "__main__":
+    _TestProject().main()
--- a/tests/generate_status_reference_data.py
+++ b/tests/generate_status_reference_data.py
@@ -1,0 +1,192 @@
+#!/usr/bin/env python
+# Copyright (c) 2018 The Regents of the University of Michigan
+# All rights reserved.
+# This software is licensed under the BSD 3-Clause License.
+import argparse
+import contextlib
+import io
+import itertools
+import operator
+import os
+import sys
+from hashlib import sha1
+
+import jinja2
+import signac
+from define_status_test_project import TestProject
+from test_project import redirect_stdout
+
+import flow
+import flow.environments
+from flow.scheduling.fake_scheduler import FakeScheduler
+
+PROJECT_NAME = "StatusTest"
+ARCHIVE_DIR = os.path.normpath(
+    os.path.join(os.path.dirname(__file__), "./status_reference_data.tar.gz")
+)
+PROJECT_DIRECTORY = "/home/user/project/"
+MOCK_EXECUTABLE = "/usr/local/bin/python"
+
+
+def init(project):
+    """Initialize the data space for the given project."""
+
+    for a in range(2, 4):
+        for b in range(2):
+            project.open_job(dict(a=a, b=b)).init()
+            project.open_job(dict(a=dict(a, a), b=b)).init()
+
+
+@contextlib.contextmanager
+def get_masked_flowproject(p, environment=None):
+    """Mock environment-dependent attributes and functions. Need to mock
+    sys.executable before the FlowProject is instantiated, and then modify the
+    root_directory and project_dir elements after creation."""
+    try:
+        old_executable = sys.executable
+        sys.executable = MOCK_EXECUTABLE  # Not sure if this is necessary
+        fp = TestProject.get_project(root=p.root_directory())
+        if environment is not None:
+            fp._environment = environment
+        fp._entrypoint.setdefault("path", "generate_status_reference_data.py")
+        fp.config.project_dir = PROJECT_DIRECTORY
+        old_generate_id = flow.project.FlowGroup._generate_id
+
+        def wrapped_generate_id(self, aggregate, *args, **kwargs):
+            """Mock the root directory used for id generation.
+
+            We need to generate consistent ids for all operations. This
+            mocking has to happen within this method to avoid affecting other
+            methods called during the test that access the project root directory.
+            """
+            old_root_directory = fp.root_directory
+            fp.root_directory = lambda: PROJECT_DIRECTORY
+            operation_id = old_generate_id(self, aggregate, *args, **kwargs)
+            fp.root_directory = old_root_directory
+            return operation_id
+
+        flow.project.FlowGroup._generate_id = wrapped_generate_id
+        yield fp
+
+    finally:
+        sys.executable = old_executable
+        flow.project.FlowGroup._generate_id = old_generate_id
+
+
+def main(args):
+    # If the ARCHIVE_DIR already exists, only recreate if forced.
+    if os.path.exists(ARCHIVE_DIR):
+        if args.force:
+            print(f"Removing existing archive '{ARCHIVE_DIR}'.")
+            os.unlink(ARCHIVE_DIR)
+        else:
+            print(
+                "Archive '{}' already exists, exiting. "
+                "Use `-f/--force` to overwrite.".format(ARCHIVE_DIR)
+            )
+            return
+
+    with signac.TemporaryProject(name=PROJECT_NAME) as p:
+        init(p)
+        with get_masked_flowproject(p) as fp:
+            # Here we set the appropriate executable for all the operations. This
+            # is necessary as otherwise the default executable between submitting
+            # and running could look different depending on the environment.
+            for group in fp.groups.values():
+                for op_key in group.operations:
+                    if op_key in group.operation_directives:
+                        group.operation_directives[op_key][
+                            "executable"
+                        ] = MOCK_EXECUTABLE
+            for job in fp:
+                with job:
+                    kwargs = job.statepoint()
+                    env = get_nested_attr(flow, kwargs["environment"])
+                    # We need to set the scheduler manually. The FakeScheduler
+                    # is used for two reasons. First, the FakeScheduler prints
+                    # scripts to screen on submission and we can capture that
+                    # output. Second, the FakeScheduler won't try to call any
+                    # cluster executable (e.g. squeue) associated with the real
+                    # schedulers used on supported clusters. Otherwise
+                    # submission would fail when attempting to determine what
+                    # jobs already exist on the scheduler.
+                    env.scheduler_type = FakeScheduler
+                    fp._environment = env
+                    parameters = kwargs["parameters"]
+                    if "bundle" in parameters:
+                        bundle = parameters.pop("bundle")
+                        fn = "script_{}.sh".format("_".join(bundle))
+                        tmp_out = io.TextIOWrapper(io.BytesIO(), sys.stdout.encoding)
+                        with redirect_stdout(tmp_out):
+                            try:
+                                fp.submit(
+                                    jobs=[job],
+                                    names=bundle,
+                                    pretend=True,
+                                    force=True,
+                                    bundle_size=len(bundle),
+                                    **parameters,
+                                )
+                            except jinja2.TemplateError as e:
+                                print(
+                                    "ERROR:", e
+                                )  # Shows template error in output script
+
+                        # Filter out non-header lines
+                        tmp_out.seek(0)
+                        with open(fn, "w") as f:
+                            with redirect_stdout(f):
+                                print(tmp_out.read(), end="")
+                    else:
+                        for op in {**fp.operations, **fp.groups}:
+                            if "partition" in parameters:
+                                # Don't try to submit GPU operations to CPU partitions
+                                # and vice versa.  We should be able to relax this
+                                # requirement if we make our error checking more
+                                # consistent.
+                                if operator.xor(
+                                    "gpu" in parameters["partition"].lower(),
+                                    "gpu" in op.lower(),
+                                ):
+                                    continue
+                            fn = f"script_{op}.sh"
+                            tmp_out = io.TextIOWrapper(
+                                io.BytesIO(), sys.stdout.encoding
+                            )
+                            with redirect_stdout(tmp_out):
+                                try:
+                                    fp.submit(
+                                        jobs=[job],
+                                        names=[op],
+                                        pretend=True,
+                                        force=True,
+                                        **parameters,
+                                    )
+                                except jinja2.TemplateError as e:
+                                    print(
+                                        "ERROR:", e
+                                    )  # Shows template error in output script
+
+                            # Filter out non-header lines and the job-name line
+                            tmp_out.seek(0)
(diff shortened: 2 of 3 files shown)


tests/extract_status.py
```diff
@@ -0,0 +1,39 @@
...
+        os.makedirs(PROJECT_DIR)
+    elif args.force:
+        import shutil
+
+        shutil.rmtree(PROJECT_DIR)
+        os.makedirs(PROJECT_DIR)
+    else:
+        return
+
+    p = signac.init_project(name=gen.PROJECT_NAME, root=PROJECT_DIR)
```
bdice, line 27: There are two projects and we're conflating the names. In the other file, `PROJECT_NAME` is used for the project holding state points with `a` and `b` (the FlowProject whose status is printed). In this file, `PROJECT_NAME` is being used for the project holding status options in the state point. I suggest creating two variables in the generation script, perhaps something like:
```python
STATUS_OPTIONS_PROJECT_NAME = "StatusOptionsData"
PROJECT_NAME = "StatusTest"
```
Then this file would use `gen.STATUS_OPTIONS_PROJECT_NAME` on this line.
```suggestion
    p = signac.init_project(name=gen.STATUS_OPTIONS_PROJECT_NAME, root=PROJECT_DIR)
```

tests/generate_status_reference_data.py
```diff
@@ -0,0 +1,91 @@
...
+
+    for a in range(2, 4):
+        for b in range(2):
+            project.open_job(dict(a=a, b=b)).init()
+            project.open_job(dict(a=dict(a=a), b=b)).init()
+
+
+def init_status_options(project):
+
+    options = [{}, {"detailed": True}, {"parameters": ["a"]}, {"expand": True}]
```
bdice, line 33: Add more options here. I would recommend first checking their outputs and making sure that you understand the difference that each option makes. List of options: https://docs.signac.io/projects/flow/en/latest/api.html#flow.FlowProject.print_status

Here is a suggested list of options from the Python and CLI options. I would check on the "max lines/width" options to see if the values I suggested are reasonable or not. We might need to tweak the FlowProject or state points to properly test some of these options.
```suggestion
    options = [
        {},  # default options
        {"overview": False},  # --no-overview
        {"overview_max_lines": 2},  # --overview-max-lines 2
        {"detailed": True},  # -d, --detailed
        {"parameters": ["a"]},  # -p a, --parameters a
        {"param_max_width": 1},  # --param-max-width 1
        {"expand": True},  # -e, --expand
        {"all_ops": True},  # -a, --all-operations
        {"detailed": True, "all_ops": True},  # --full
        {"only_incomplete": True},  # --only-incomplete-operations
        {"dump_json": True},  # --json
        {"unroll": False},  # --stack
        {"compact": True},  # -1, --one-line
        {"pretty": True},  # --pretty
        {"profile": True},  # No CLI flag
        {"eligible_jobs_max_lines": 2},  # --eligible-jobs-max-lines 2
        {"output_format": "markdown"},  # -o markdown, --output-format markdown
        {"output_format": "html"},  # -o html, --output-format html
    ]
```

tests/generate_status_reference_data.py
```diff
@@ -0,0 +1,91 @@
...
+        fp = _TestProject.get_project(root=p.root_directory())
+        env = flow.environment.TestEnvironment
+        # We need to set the scheduler manually. The FakeScheduler
+        # is used for two reasons. First, the FakeScheduler prints
+        # scripts to screen on submission and we can capture that
+        # output. Second, the FakeScheduler won't try to call any
+        # cluster executable (e.g. squeue) associated with the real
+        # schedulers used on supported clusters. Otherwise
+        # submission would fail when attempting to determine what
+        # jobs already exist on the scheduler.
```
bdice, line 66:
```suggestion
        # We need to set the scheduler manually. The FakeScheduler won't try to
        # call any cluster executable (e.g. squeue) associated with the real
        # schedulers used on supported clusters. Otherwise status checks would
        # fail when attempting to determine what jobs exist on the scheduler.
```

tests/generate_status_reference_data.py
```diff
@@ -0,0 +1,91 @@
...
+        env.scheduler_type = FakeScheduler
+        fp._environment = env
+        for job in status_pr:
+            with job:
```
bdice, line 70: This `with job:` can be removed because we're using `job.fn("status.txt")` below.

tests/generate_status_reference_data.py
```diff
@@ -0,0 +1,91 @@
...
+import signac
+from define_status_test_project import _TestProject
+from test_project import redirect_stdout
+
+import flow
+import flow.environments
+from flow.scheduling.fake_scheduler import FakeScheduler
+
+PROJECT_NAME = "StatusTest"
+ARCHIVE_DIR = os.path.normpath(
```
bdice, line 17: I think this would make more sense if it were named `ARCHIVE_PATH`. It is a file, not a directory. If you agree, please make the same change to the template tests and update the CLI help messages referring to this variable at the bottom.
```suggestion
ARCHIVE_PATH = os.path.normpath(
```

tests/test_status.py
```diff
@@ -0,0 +1,41 @@
...
+import generate_status_reference_data as gen
+import signac
+from test_project import redirect_stderr, redirect_stdout
+
+
+def test_print_status():
+
+    # Force asserts to show the full file when failures occur.
+    # Useful to debug errors that arise.
+
```
bdice, line 17: I don't know what this means. Let's remove it unless you know. Also make this change to the template tests if relevant.
```suggestion
```

tests/test_status.py
```diff
@@ -0,0 +1,41 @@
...
+                    with redirect_stdout(tmp_out):
+                        fp.print_status(**kwargs)
+
+                    tmp_out.seek(0)
+                    msg = f"---------- Status options {kwargs} for job {job}."
+                    generated.extend([msg] + tmp_out.read().splitlines())
+
+                    with open(job.fn("status.txt")) as file:
+                        reference.extend([msg] + file.read().splitlines())
+        assert "\n".join(generated) == "\n".join(reference)
```
bdice, line 41: Let's move the "reference == generated" check into the for loop, so that we can verify the status output for each set of status options individually. I did some additional refactoring based on other suggestions above, too.
```suggestion
        for job in status_pr:
            reference = []
            generated = []
            kwargs = job.statepoint()
            tmp_out = io.TextIOWrapper(io.BytesIO(), sys.stdout.encoding)
            with open(os.devnull, "w") as devnull:
                with redirect_stderr(devnull):
                    fp.print_status(**kwargs, file=tmp_out)
                    tmp_out.seek(0)
                    msg = f"---------- Status options {kwargs} for job {job}."
                    generated = [msg] + tmp_out.read().splitlines()
                    with open(job.fn("status.txt")) as status_file:
                        reference = [msg] + status_file.read().splitlines())
            assert "\n".join(generated) == "\n".join(reference)
```

tests/generate_status_reference_data.py
```diff
@@ -0,0 +1,91 @@
...
+        env.scheduler_type = FakeScheduler
+        fp._environment = env
+        for job in status_pr:
+            with job:
+                kwargs = job.statepoint()
+                fn = job.fn("status.txt")
+
+                with open(fn, "w") as f:
+                    with redirect_stdout(f):
+                        fp.print_status(**kwargs)
```
bdice, line 76:
```suggestion
                with open(job.fn("status.txt"), "w") as status_file:
                    fp.print_status(**kwargs, file=status_file)
```

bdice (changes requested):
@Charlottez112 Good work so far! I have some suggestions for cleaning this up and expanding the test data. Once you have a set of status options that you think are appropriate to test (which may be more or less extensive than what I suggested in my review), then please commit the `.tar.gz` file and see if tests pass on CI.

tests/define_status_test_project.py
```diff
@@ -0,0 +1,92 @@
...
+        # The submit interface should warn about unused directives.
+        "bad_directive": 0,
+        # But not this one:
+        "np": 1,
+    }
+)
+@flow.cmd
+@_TestProject.pre(b_is_even)
+@group1
+@_TestProject.post.isfile("world.txt")
```
atravitz, line 50:
```suggestion
@flow.cmd
@group1
@_TestProject.pre(b_is_even)
@_TestProject.post.isfile("world.txt")
```
reordering just for clarity

tests/generate_status_reference_data.py
```diff
@@ -0,0 +1,91 @@
...
+import flow.environments
+from flow.scheduling.fake_scheduler import FakeScheduler
+
+PROJECT_NAME = "StatusTest"
+ARCHIVE_DIR = os.path.normpath(
+    os.path.join(os.path.dirname(__file__), "./status_reference_data.tar.gz")
+)
+
+
+def init(project):
```
atravitz, line 22: is it important here to check non-numeric key values, or is it handled elsewhere that you don't have to here?

atravitz (commented):
this looks good! please request my review again once you've added more options

tests/generate_status_reference_data.py
Charlottez112: Sorry it took a while. I'm not sure what you mean. Could you explain what non-numeric key values you're referring to?

tests/generate_status_reference_data.py
Charlottez112: I've added all the options. I have a question about `{'profile': True}` - could you explain what this does? The script `generate_status_reference_data.py` fails when I add this option.

The error I got:
```
  File "/Users/zshiqi/code/signac-flow/tests/generate_status_reference_data.py", line 103, in <module>
    main(parser.parse_args())
  File "/Users/zshiqi/code/signac-flow/tests/generate_status_reference_data.py", line 88, in main
    fp.print_status(**kwargs, file=status_file)
  File "/Users/zshiqi/code/signac-flow/flow/project.py", line 2684, in print_status
    if re.match(profile, module_fn):
  File "/Users/zshiqi/miniconda3/envs/signac-flow-dev/lib/python3.9/re.py", line 191, in match
    return _compile(pattern, flags).match(string)
  File "/Users/zshiqi/miniconda3/envs/signac-flow-dev/lib/python3.9/re.py", line 303, in _compile
    raise TypeError("first argument must be string or compiled pattern")
TypeError: first argument must be string or compiled pattern
```

bdice: @Charlottez112 From what I saw on Slack, it sounded like you might have discussed the `profile` option with @atravitz during office hours. If there are a small number of options like `profile` that are problematic, we can just skip them in the tests.

tests/generate_status_reference_data.py
bdice: @Charlottez112 I think @atravitz might have meant that the state points initialized in this loop only contain integer values. While it would be nice to test non-integer values (e.g. strings or floats), I don't think that is important enough to delay this PR being merged.

tests/generate_status_reference_data.py
```diff
@@ -0,0 +1,103 @@
...
+        {"parameters": ["a"]},  # -p a, --parameters a
+        {"param_max_width": 1},  # --param-max-width 1
+        {"expand": True},  # -e, --expand
+        {"all_ops": True},  # -a, --all-operations
+        {"only_incomplete": True},  # --only-incomplete-operations
+        {"dump_json": True},  # --json
+        {"unroll": False},  # --stack
+        {"compact": True},  # -1, --one-line
+        {"pretty": True},  # --pretty
+        #    {"profile": True},  # No CLI flag
```
bdice, line 47: Update this comment to indicate why you skipped `profile`. Maybe just indicate there is a problem of some kind.

bdice (changes requested):
@Charlottez112 This is really close -- please commit the test data so that we can see if CI passes!

tests/generate_status_reference_data.py
Charlottez112: When I set the option `profile` to `True`, I got the following error:
```
  File "/Users/zshiqi/code/signac-flow/tests/generate_status_reference_data.py", line 103, in <module>
    main(parser.parse_args())
  File "/Users/zshiqi/code/signac-flow/tests/generate_status_reference_data.py", line 88, in main
    fp.print_status(**kwargs, file=status_file)
  File "/Users/zshiqi/code/signac-flow/flow/project.py", line 2684, in print_status
    if re.match(profile, module_fn):
  File "/Users/zshiqi/miniconda3/envs/signac-flow-dev/lib/python3.9/re.py", line 191, in match
    return _compile(pattern, flags).match(string)
  File "/Users/zshiqi/miniconda3/envs/signac-flow-dev/lib/python3.9/re.py", line 303, in _compile
    raise TypeError("first argument must be string or compiled pattern")
TypeError: first argument must be string or compiled pattern
```

I discussed this with @atravitz during the office hour, and we think this might be a bug in the original code.

b-butler: This does seem to be a bug. Rerequest a review from Bradley, and we should create an issue for this unless you want to fix the bug in this PR. @Charlottez112

Charlottez112: > This does seem to be a bug. Rerequest a review from Bradley, and we should create an issue for this unless you want to fix the bug in this PR. @Charlottez112

I'd like to help fix this but I don't really know what `re.match` is trying to match. I'm not sure what `prof.merged_file_dict` is. I'll create the issue after Bradley takes a look at this.

Charlottez112:
According to @bdice the `profile` option should take a string (the name of the functions to profile), instead of a boolean (in the documentation). So I added that option with a function name and the tests passed locally. Now the tests are failing here because of `ImportError: Profiling requires the pprofile package`. I'm not sure which files I need to change.

tests/generate_status_reference_data.py
```diff
@@ -0,0 +1,103 @@
...
+        {"dump_json": True},  # --json
+        {"unroll": False},  # --stack
+        {"compact": True},  # -1, --one-line
+        {"pretty": True},  # --pretty
+        {"profile": "b_is_even"},  # No CLI flag
+        {"eligible_jobs_max_lines": 2},  # --eligible-jobs-max-lines 2
+        {"output_format": "markdown"},  # -o markdown, --output-format markdown
+        {"output_format": "html"},  # -o html, --output-format html
+    ]
+    # options = [{}, {"detailed": True}, {"parameters": ["a"]}, {"expand": True}]
```
bdice, line 52: Delete.
```suggestion
```

bdice:
@Charlottez112 Go ahead and remove the tests for profiling, then this PR should be nearing completion. We discussed in the signac dev meeting and will probably remove the profiling feature entirely. There are issues with installing `pprofile` with `setuptools>58` (https://github.com/vpelletier/pprofile/issues/41) and we don't need the feature badly enough to continue supporting it.

tests/test_status.py
```diff
@@ -0,0 +1,33 @@
...
+# This software is licensed under the BSD 3-Clause License.
+import io
+import os
+import sys
+
+import generate_status_reference_data as gen
+import signac
+
+
+def test_print_status():
```
b-butler, line 12: I think that using fixtures to make this a test that is parametrized by the status arguments used would be cleaner, but this should hold the PR up: it would just make it cleaner.

tests/test_status.py
```diff
@@ -0,0 +1,33 @@
...
+    # Must import the data into the project.
+    with signac.TemporaryProject(name=gen.PROJECT_NAME) as p, signac.TemporaryProject(
+        name=gen.STATUS_OPTIONS_PROJECT_NAME
+    ) as status_pr:
+        gen.init(p)
+        fp = gen._TestProject.get_project(root=p.root_directory())
+        status_pr.import_from(origin=gen.ARCHIVE_PATH)
+        for job in status_pr:
+            kwargs = job.statepoint()
+            msg = f"---------- Status options {kwargs} for job {job}."
```
b-butler, line 23: Why is this this message added to both the generated and reference data?

b-butler (changes requested):
We should document how to update the status tests. Is there a corresponding docs PR?

tests/define_status_test_project.py
```diff
@@ -0,0 +1,64 @@
...
+@_TestProject.post.true("test")
+def op2(job):
+    job.document.test = os.getpid()
+
+
+@_TestProject.operation
+@group2
+@_TestProject.post.true("test3_true")
+@_TestProject.post.false("test3_false")
+@_TestProject.post.not_(lambda job: job.doc.test3_false)
```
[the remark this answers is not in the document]
bdice: @b-butler If I recall correctly, the project that this was copy-pasted from included `not_` for the purposes of increasing test coverage. We had no other tests of that condition.

It is equivalent to the `post.false("test3_false")` condition but will not raise an error because it is the last postcondition. The earlier postconditions are not met until after the operation runs, so the "early exit" on an unmet postcondition prevents this from running (and raising). However, this is not great design.

For this PR, we can simplify the conditions. I don't think status information will be affected if we were to simplify to the following:

```python
@_TestProject.operation
@group2
@_TestProject.post.true("test2")
def op3(job):
    job.document.test2 = True
```

Separately, we could change the implementation of the `not_` condition in the test file this came from so that the potential error is eliminated:

```python
@_TestProject.post.not_(lambda job: job.doc.get("test3_false", None))
```

tests/generate_status_reference_data.py
```diff
@@ -0,0 +1,101 @@
...
+def main(args):
+    # If the ARCHIVE_PATH already exists, only recreate if forced.
+    if os.path.exists(ARCHIVE_PATH):
+        if args.force:
+            print(f"Removing existing archive '{ARCHIVE_PATH}'.")
+            os.unlink(ARCHIVE_PATH)
+        else:
+            print(
+                "Archive '{}' already exists, exiting. "
+                "Use `-f/--force` to overwrite.".format(ARCHIVE_PATH)
```
bdice, line 65:
```suggestion
                f"Archive '{ARCHIVE_PATH}' already exists, exiting. "
                "Use `-f/--force` to overwrite."
```

tests/generate_status_reference_data.py
```diff
@@ -0,0 +1,101 @@
...
+            )
+            return
```
bdice, line 67: This `return` doesn't do anything.
```suggestion
```

tests/test_status.py
bdice: This is just a separator / header that makes it easier to tell how the outputs are generated.

tests/test_status.py
```diff
@@ -0,0 +1,33 @@
...
+            with open(os.devnull, "w") as devnull:
+                tmp_out = io.TextIOWrapper(io.BytesIO(), sys.stdout.encoding)
+                # with redirect_stderr(devnull):
+                #     with redirect_stdout(tmp_out):
```
bdice, line 27: Remove unused code.
```suggestion
```

bdice (approved):
I have left some minor suggestions. Overall I think this is ready after applying my suggestions.

bdice:
@b-butler Can you re-review / reply in comment threads? I think @Charlottez112 has handled some of your change requests and I commented on the other ones. Thanks!

PR #572, opened 2021-09-07
Quote project dir in base_script template when changing directory

javierbg:
Added quotes around the project directory path when changing directory in the base_script template to avoid errors when using paths with spaces.

## Description

Just added a couple of quotes

## Motivation and Context

I recently started a project where the path contained spaces and started having problems when submitting jobs. This was the cause.

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:


bdice:
We should add tests for a case with spaces and special characters that require escaping as well. We may be able to use the built-in “simple scheduler” (or another means) to test execution of the submitted shell script, or at least we can add template tests like we do for directives.

tommy-waltmann:
> We should add tests for a case with spaces and special characters that require escaping as well. We may be able to use the built-in “simple scheduler” (or another means) to test execution of the submitted shell script, or at least we can add template tests like we do for directives.

Agreed, add tests on this and then I will give a full review.

mikemhenry:
Thanks for the PR @javierbg :wave: Thank you for bringing this issue to our attention.  The tests are failing since we will need to re-generate the output of the scripts with quotes. Are you interested in adding a few tests for this PR?

javierbg:
@mikemhenry No problem, but it might take some days until I got some time to do it

javierbg:
Ok, I fixed the reference tests, but sadly I'm not familiarized enough with the testing framework to know where to add the checks for projects with directories containing spaces. I can give it a go some time later, but I'm not sure if I'll be capable.

b-butler:
@bdice @tommy-waltmann, do either of you want to add the necessary tests? If not, I suggest merging and making an issue to test this behavior.

tommy-waltmann:
I would merge and make the issue then.

bdice:
@b-butler @tommy-waltmann I hoped that adding tests would make it easier to investigate a potential bug: if paths contain a double quote, this will fail. Rather than adding quotes around the path, this PR should probably use some string/path method to properly escape the path. I would recommend investigating that route before we merge this PR. I can take a look eventually, but I have a somewhat long backlog at the moment.

javierbg:
> There is probably a function in the Python standard library for this exact use case

I have searched for a couple hours, but I cannot find anything of the sort. Essentially, the problem is making sure that the path is interpreted correctly by bash as a single argument. The approach now is to always double quote this command and escape both backslashes and double quotes. This is done directly in the base script template.

This time I have modified the mock test project directory path in the template generator script, which is where the environment test takes this path. I find this a bit deceiving: of course, the same code will generate the same script, we are just saving it in a .tar.gz and running it again when the test is ran. Also, the environment test feels quite monolithic, more like an integration test. I initially tried adding a specific test for this case, but I would have to essentially copy over the same test and adding a different directory name.

While I think this approach is better than before, I am still not fully satisfied about the test and the path escaping. Maybe this is fine?

bdice:
@javierbg Sorry, I should have looked up the function in my previous comment. I found it here. `shlex.quote`: https://docs.python.org/3/library/shlex.html Hopefully this function resolves any uncertainty about whether we are handling the path safely. 👍 I think (please check) that you do not need surrounding quotes if you use this function, it will add them only if necessary.

You’re correct that the template tests are basically a massive regression test. We have no way to externally validate the submission scripts that we generate in our CI, so we rely on saving a tarball full of output scripts that test various options and manually checking that they appear to be valid SLURM/PBS/LSF scripts. Your alteration to the project directory looks like the right thing to do here.

javierbg:
@bdice No, thank you for tour work on signac! I will finish this up in a couple of days when I am back home

javierbg:
I tried to follow the same scheme that I observed for adding new items to the changelog for future releases, apologies if I did it incorrectly.

mikemhenry (approved):
LGTM! Thanks! :tada:

Issue #506, opened 2021-04-26
Establish clear flow schema


vyasr:
This issue is part of glotzerlab/signac#527 but requires a more focused thread.

signac-flow currently stores a number of pieces of internal data, such as bundles, submission status, and configuration information for FlowProjects. The current data organization is ad hoc and relies on signac's config, and we want to decouple flow from these signac internals. #504 is the first step, establishing an explicit schema versioning for flow's internal data and enabling migration. The next step is defining a clear schema for signac-flow. 

Concrete proposals:
- Running `flow init` in a directory should run `signac init` if necessary (already happens), then create an additional directory `.flow` and a file `.flow/config` for flow config information.
- Bundles should be stored in a file `.flow/${FLOWPROJECT_NAME}_bundles`.
- Status should be stored in a file `.flow/${FLOWPROJECT_NAME}_status`. It _should not_ be stored in the project document, which is user-facing.
- Submission ids should use FlowProject subclass names rather than signac project names. This change is also related to glotzerlab/signac#541.

Possible proposals:
- Config information should be stored per-FlowProject, not per signac project. Many config values conceptually belong to a FlowProject more than a signac project.
- Config information should be stored directly in the FlowProject class definition rather than in a config files.
  - Pros: 
    - Removes the need to maintain an external config file
    - Clearly and visibly associates config data with a FlowProject
  - Cons:
    - Would require storing possibly "private" data like account names on supercomputers into the "public" `project.py` files (note that we already do this in signac.rc, but maybe we can find a way to avoid that). One possibility is that _some_ information ends up belonging to the FlowProject class, while other information (like accounts) goes in a config file.
    - This may make it harder to copy over from one machine to another. The problem could be ameliorated by formalizing an association between accounts and environments so that users have a way to specify the account on one supercomputer vs another (which we could internally resolve using our hostname pattern logic), but it will invariably be more verbose. 
- Bundle and status files can use JSON to get the migration going, but we should consider using alternative formats (e.g. a compressed binary file) to reduce I/O and speed up flow's internals. One option would be to formalize something like the gzip format used for the signac project cache into a new synced collections backend.

Open questions:
[...]

PR #573, opened 2021-09-26
Replace monkey-patching of flow config with post-load validation

vyasr:
## Description

This PR removes the code patching signac-flow configuration schema and defaults into the configobj schema stored in signac. Instead, we now rely on signac solely for the purpose of loading in the schema from wherever it is stored and perform validation of all flow-related keys within signac-flow upon instantiation of a FlowProject. This change also means that project-specific information in the config is now immutable once a FlowProject has been instantiated. This is not true for environment information, which is still read from disk when the environment class needs it.

## Motivation and Context

The monkey-patching of the config is the primary reason that signac-flow is currently subject to potential breakage if signac changes the way it stores configuration information. The purpose of the monkey-patching is to ensure that flow's schema is validated when a config file is read from disk by signac. By moving the validation logic into the FlowProject, we no longer need to perform this monkey-patching, which means that in the future signac-flow can rely on signac's config API alone with no dependence on implementation details like the file where the config is stored. 

This PR contributes to glotzerlab/signac#527 and #506 and should make it easier to move forward on these two packages independently.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


bdice (changes requested):

vyasr:
Thanks for the review @bdice. The immediate plan going forward is to:
1. Clean up the issues with this PR and get it merged largely as is.
2. When working on glotzerlab/signac#527, and in particular glotzerlab/signac#283 for signac 2.0, we will decouple the reading of the config file from validation of its contents.
3. Expose a public API in signac (exact form TBD) for validating the contents of a configuration against a schema.
4. Once flow is updated to rely on signac >= 2.0, modify it to use signac to handle validation.

flow/__init__.py
```diff
@@ -42,7 +43,10 @@
 ]
 
 
-if _get_config_value("import_packaged_environments", default=True):
+if _get_config_value(
```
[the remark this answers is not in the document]
vyasr: In light of our plans for moving more of the validation logic (hopefully including default handling) into signac's config API, let's leave this as is in this PR and revisit it later. Most aspects of our config handling feel like they need some work. I think the way that this should probably work is that signac-flow should wrap signac's `read_config` function to perform flow-specific validation so that we can put all default-handling into a standardized validate-with-defaults step, but I haven't thought that all the way through yet.

flow/util/config.py
```diff
@@ -6,15 +6,36 @@
...
+# [flow]
+# import_packaged_environments = boolean()
+# status_performance_warn_threshold = float(default=0.2)
+# show_traceback = boolean()
+# eligible_jobs_max_lines = int(default=10)
+# status_parallelization = string(default='none')
+# """
+
+
+_FLOW_SCHEMA = {
```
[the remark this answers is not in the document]
bdice: Discussed during dev meeting -- we're going to move forward with jsonschema. Resolving.

.circleci/ci-oldest-reqs.txt
```diff
@@ -2,3 +2,4 @@ cloudpickle==1.1.1
 deprecation==2.0.0
 jinja2==2.10
 tqdm==4.48.1
+jsonschema>=3.0.0
```
bdice, line 5: 1. Oldest test requirements should be exact pinnings (not `>=`).
2. [requirements/requirements-test.txt](https://github.com/glotzerlab/signac-flow/blob/master/requirements/requirements-test.txt) should be pinned to the latest version, `jsonschema==4.0.1`. 4.0.1 was just released: https://pypi.org/project/jsonschema/

flow/project.py
```diff
@@ -1546,6 +1547,33 @@ def hello(job):
...
+        # supporting in-place config modification, flow can make use of the
+        # signac Project config dictionary directly and that can be updated by
+        # any flow config APIs. For now, we store the flow config separately to
+        # avoid any side effects associated with modifying instances of
+        # signac.contrib._ProjectConfig.
+        self._flow_config = {
+            **flow_config._FLOW_CONFIG_DEFAULTS,
+            **self._config.get("flow", {}),
+        }
+        # TODO: Is there a more elegant solution to casting strings directly?
```
bdice, line 1564: No, this is fine. Remove the TODO.
```suggestion
```

flow/project.py
```diff
@@ -4787,7 +4813,7 @@ class MyProject(FlowProject):
 
         # Read the config file and set the internal flag.
         # Do not overwrite with False if not present in config file
-        if flow_config.get_config_value("show_traceback"):
+        if self._flow_config["show_traceback"]:
```
bdice, line 4816: Is this being cast correctly to bool? The other config keys were handled with specific casts in `FlowProject.__init__`. We don't want `"False"` (or `"off"` or whatever configobj supports as boolean strings) to be interpreted as truthy.

bdice (changes requested):

flow/project.py
vyasr: Another good idea. I've largely adapted your snippet as is aside from the fact that we can safely assume that the input is a string since configobj always loads it that way. However, I did uncover a deeper issue with our current config loading: we're not actually doing anything with validation. If validation fails, all that happens is that a [debug-level logger message](https://github.com/glotzerlab/signac/blob/master/signac/common/config.py#L104) is emitted. If a valid value is provided then everything works fine, and configobj will coerce things to the appropriate types as requested by our schema. If an invalid value is provided, though, we're silently passing it through. I've opted not to do anything about it at this time since the immediate next step in the config overhaul is to rewrite this part of signac anyway, but it's something that I wanted documented somewhere.

doc/requirements.txt
```diff
@@ -1 +1,2 @@
 signac
+jsonschema
```
bdice, line 2: This file can be deleted, it's missing several other dependencies. I updated the ReadTheDocs config to use the root `requirements.txt`.

flow/project.py
bdice: Thanks for the fix, and thanks for looking into this further.

bdice (approved):
We can delete `doc/requirements.txt`, I changed the ReadTheDocs config to use `requirements.txt` instead.

doc/requirements.txt
vyasr: @bdice RTD is failing because we need to pin docutils. Where would you like to put that pinning since we don't currently bother to pin anything docs related explicitly?

bdice: I'll fix this upstream. We just need to use a newer RTD build environment.

bdice: I'll delete this file and merge once I get the RTD config fixed.

bdice: Just kidding. I'll change it on this branch so we know it works.

bdice:
@vyasr I fixed the documentation issues. I'm going to merge this once tests pass, then apply similar changes to signac and related repos.

PR #586, opened 2021-11-15
CI: Test whether installation works as described in documentation.

csadorf:
Add steps to nightly CI work chain that tests whether the package can be installed as described in the documentation.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -249,6 +249,19 @@ workflows:
   version: 2
   test:
     jobs:
+      # DEBUG: MOVE THIS BACK TO NIGHTLY AFTER TESTING
+      - test-install-pip-python-36
+      - test-install-pip-python-37
+      - test-install-pip-python-38
+      - test-install-pip-python-39
+      - test-install-pip-python-310
+      - test-install-conda
+      - test-install-conda-python-36
+      - test-install-conda-python-37
+      - test-install-conda-python-38
+      - test-install-conda-python-39
+      - test-install-conda-python-310
+      # END DEBUG
 #     - pre-checks
       - linux-python-310
       - linux-python-39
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -121,6 +121,88 @@ jobs:
...
+          command: |
+            pip install --progress-bar off signac signac-flow
+
+      - run: &test-run-and-import
+          name: test-run-and-import
+          command: |
+            signac --version
+            python -c 'import signac'
+            flow --version
+            python -c 'import flow'
```
bdice, line 141: Do we know if importing is enough to catch the issues we want to catch? This would not catch issues that rely on actually executing code.

csadorf: It's a smoke test. We can expand that to more tests if we like, but I would prefer to leave it as is.

.circleci/config.yml
```diff
@@ -121,6 +121,88 @@ jobs:
     docker:
       - image: cimg/python:3.6
 
+  test-install-pip-python-36: &test-install-pip
```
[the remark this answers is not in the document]
csadorf: That does not seem to be a thing:

```console
$ docker run cimg/python python --version
Unable to find image 'cimg/python:latest' locally
docker: Error response from daemon: manifest for cimg/python:latest not found: manifest unknown: manifest unknown.
See 'docker run --help'.
```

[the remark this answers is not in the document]
bdice: Sounds good. Resolving.

.circleci/config.yml
```diff
@@ -206,6 +288,17 @@ workflows:
     jobs:
       - linux-python-310-signac-latest
       - linux-python-310-signac-next
+      - test-install-pip-python-36
+      - test-install-pip-python-37
+      - test-install-pip-python-38
+      - test-install-pip-python-39
+      #- test-install-pip-python-310  # not yet supported
+      #- test-install-conda  # selects Python 3.0 (not yet supported)
```
csadorf, line 296:
```suggestion
      #- test-install-conda  # currently selects Python 3.10 (not yet supported)
```

csadorf:
@bdice I would appreciate if you could renew your approval, all tests pass now.

bdice (approved):

Issue #547, opened 2021-06-27
Add groups in status output


bdice:
### Feature description

I submitted groups containing multiple operations, but the status output does not reflect this. It appears that groups are currently not integrated with the status output code, resulting in unknown `[U]` status codes even when the operations are contained in actively running groups. (See screenshot.)

![image](https://user-images.githubusercontent.com/3943761/123550550-a9fa7700-d733-11eb-9cfb-50065ce4d352.png)

### Proposed solution

Perhaps we need another status code for operations like `[GA]` for "group active" or `[GQ]` for "group queued." We could also add a separate line in the status output for each group present in the project. I think it might even make sense to do both, so that it would look like this. Assume `my_group` contains `my_operation1` and `my_operation2`. 10 jobs were submitted as `my_group`, 5 of which are queued and 5 of which are running.

```
operation        number of eligible jobs  submission status
-------------  -------------------------  -------------------------
my_operation1                         10  [U]: 10, [GQ]: 5, [GA]: 5
my_operation2                         10  [U]: 10, [GQ]: 5, [GA]: 5
my_group                              10  [U]: 10, [Q]: 5, [A]: 5
```

### Additional context

Related to #469.

csadorf:
Maybe the status output should only show groups. I've argued before that operations are in essence just groups of size 1. I think internally we still need to treat operations as first-class entities, but for the status view, everything could be a group?

b-butler:
@csadorf I like the idea in principle, but I think that only showing groups would be confusing to users. Since every operation has its singleton group, the operations would seem to be there regardless (as they are properly groups). However, when submitting through a user created group, they would show as unsubmitted, but would not be eligible for submission with the only indication of their status being the status of the composing group. That may be fine since multi-operation groups are an opt-in feature, but we would just need to be very clear in our documentation.

I would also potentially want more distinction from singleton operations and user/multi-operation groups since in this case `[GA], [GQ]` could never apply to them. I do think having a separate status for user group submission makes sense as just because a group an operation belongs to is submitted or active does not mean that that operation will run/complete.

I am in favor of @bdice's solution with an additional heading/separation for user groups (even though semantically multi-operation groups are not different from single operation groups)

csadorf:
[...]

PR #593, opened 2021-12-13
Group Status Output

b-butler:
- Add groups to status output
- Add operation within group statuses.

## Description

This implements the status output suggested and discussed in #547.

Status test outputs have been regenerated, and locally checked using a project I am working on that utilizes groups.

_Note_: I still need to work on the detailed status output.

## Motivation and Context

Having non-default groups in the status output is very helpful for users.

## Types of Changes

- [ ] Documentation update
- [ ] Bug fix
- [x] New feature

## Checklist:

If necessary:


flow/scheduling/base.py
```diff
@@ -45,12 +49,47 @@ class JobStatus(enum.IntEnum):
     error = 8
     """The cluster job is in an error or failed state."""
 
+    group_registered = 9
+    """The operation via a group is registered with the scheduler.but no other status is known."""
```
bdice, line 53: What does "via a group" mean? I would prefer phrasing like:
```suggestion
    """The operation is in a group that is registered with the scheduler but no other status is known."""
```

flow/project.py
```diff
@@ -108,6 +108,13 @@
     JobStatus.active: "A",
     JobStatus.error: "E",
     JobStatus.placeholder: " ",
```
bdice, line 110: Re-order this to put placeholder at the bottom, like in `scheduling/base.py`.

flow/scheduling/base.py
```diff
@@ -45,12 +49,47 @@ class JobStatus(enum.IntEnum):
...
 
     user = 128
     """All user-defined states must be >=128 in value."""
 
+    @classmethod
+    def _to_group(cls, status):
+        """Convert to an operation within a group status."""
+        if status == JobStatus.unknown or int(status) > 127:
+            raise ValueError(f"No equivalent group status for {status}.")
+        return cls(int(status) + 7)
```
bdice, line 91: We should not do addition here. This should use an explicit dictionary mapping like:
```python
to_group_mapping = {
    JobStatus.registered: JobStatus.group_registered,
    JobStatus.inactive: JobStatus.group_inactive,
    # ...
}
return cls(to_group_mapping[status])
```

bdice (changes requested):
Overall this looks pretty good so far. Nice work @b-butler. I have two major requests for this PR:

1. Let's benchmark this carefully. I spent a lot of time ensuring that the status code is performant, so we should measure this to ensure we haven't regressed.

2. Let's add test coverage for as many of the missing lines as we can. Poor test coverage in status code has been a longstanding problem for signac-flow that we've recently begun to improve upon.

flow/util/template_filters.py
```diff
@@ -356,3 +356,13 @@ def get_account_name(environment, required=False):
...
             )
             _GET_ACCOUNT_NAME_MESSAGES_SHOWN.add(env_name)
+
+
+def format_operation_name(operation_name, project):
+    """Return the operation name or group name with number of operations."""
+    # both checks are necessary to ensure the operations omitted line passes through
+    if operation_name not in project._operations and operation_name in project._groups:
+        return (
+            operation_name + f" ({len(project._groups[operation_name].operations)} ops)"
```
csadorf, line 366: Why not only use an f-string?
```suggestion
            f"{operation_name} ({len(project._groups[operation_name].operations)} ops)"
```

flow/util/template_filters.py
```diff
@@ -356,3 +356,13 @@ def get_account_name(environment, required=False):
...
+        )
+    return operation_name
```
csadorf, line 368: I'd prefer a slightly more functional style:
```suggestion
    return (
        f"{operation_name} ({len(project._groups[operation_name].operations)} ops)"
        if (
            # both checks are necessary to ensure the operations omitted line
            # passes through
            operation_name in project._groups
            and operation_name not in project._operations
        )
        else str(operation_name)
    )
```

flow/util/template_filters.py
```diff
@@ -356,3 +356,13 @@ def get_account_name(environment, required=False):
                 file=sys.stderr,
             )
             _GET_ACCOUNT_NAME_MESSAGES_SHOWN.add(env_name)
+
+
+def format_operation_name(operation_name, project):
+    """Return the operation name or group name with number of operations."""
+    # both checks are necessary to ensure the operations omitted line passes through
```
csadorf, line 363: I don't understand what is meant by that.

csadorf (changes requested):

flow/util/template_filters.py
b-butler: No reason just didn't occur to me in the moment.

changelog.txt
```diff
@@ -5,6 +5,17 @@ Changelog
...
+Version 0.18
+============
+
+[0.18.0] -- ????-??-??
+----------------------
+
+Changed
++++++++
+
+- Added user defined ``FlowGroups`` to status output (#593, #547).
```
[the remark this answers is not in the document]
b-butler: Got it. Thanks.

flow/project.py
```diff
@@ -2482,21 +2487,23 @@ def _fetch_status(
...
                 # one postcondition is not met).
-                status["eligible"] = not completed and group._eligible(
-                    aggregate, ignore_conditions=IgnoreConditions.POST
-                )
+                if is_user_defined_group:
+                    status["eligible"] = group._eligible(aggregate)
+                else:
+                    status["eligible"] = not completed and group._eligible(
+                        aggregate, ignore_conditions=IgnoreConditions.POST
+                    )
```
bdice, line 2506: Good catch @b-butler. It seems like this should use the condition `len(group.operations) > 1` instead of `is_user_defined_group = group.name not in self._operations`. A user-defined group with a single operation would still be able to use `not completed` as a shortcut for determining its eligibility, in my understanding. That would help avoid some extra calls to `group._eligible`.

Edge case I have spent less than ten seconds considering: can a group have zero operations? (Probably, by defining it and assigning no operations.) Is the eligibility/completion state of such a group defined in a consistent way for status checks and execution?

b-butler: The reason I used that check rather than your suggestion was to only check once since `is_user_defined_group` is used specifically in the user defined sense below, but we can do the length check here to utilize the short cut, but I am not sure if that will be much change in most cases as a single operation user defined group is likely rare.

Also, yes such operations would register right now as being complete and not eligible (according to the logic of the builtins `any` and `all`).

flow/util/template_filters.py
```diff
@@ -356,3 +356,17 @@ def get_account_name(environment, required=False):
...
+    return (
+        f"{operation_name} ({len(project._groups[operation_name].operations)} ops)"
+        if (
+            # both checks in conditional are necessary to ensure that the status line that prints
+            # out the number of operations omitted passes through without causing errors.
+            operation_name in project._groups
+            and operation_name not in project._operations
+        )
+        else operation_name
+    )
```
bdice, line 372: This logic is correct but it's written awkwardly with comments interspersed within the return statement. I suggest:
```suggestion
    # If the name is from a group that is not an operation, we append the
    # number of operations to its name in the status.
    if operation_name in project._groups and operation_name not in project._operations:
        num_operations = len(project._groups[operation_name].operations)
        return f"{operation_name} ({num_operations} ops)"
    return operation_name
```

b-butler: That is how it was but @csadorf argued for a more functional style.

lyrivera:
I am not familiar with how the status output works in signac-flow, so I am unable to provide a useful review ATM.

flow/util/template_filters.py
bdice: I'm going to apply my suggestion on the grounds that it's fewer lines of code and that it was originally written that way. Resolving.

flow/project.py
```diff
@@ -2619,32 +2625,52 @@ def compute_status(data):
...
+                {
+                    "aggregate_id": aggregate_id,
+                    "group_name": group.name,
+                    "status": status,
+                    "_error": error_text,
+                }
+            ]
+            if (
+                group.name not in self._operations
+                and scheduler_status > JobStatus.unknown
```
bdice, line 2638: "unknown" is the first status value in that enum. I think it would be clearer to check `!=`.
```suggestion
                and scheduler_status != JobStatus.unknown
```

flow/project.py
```diff
@@ -2664,6 +2690,20 @@ def compute_status(data):
...
+            group_statuses = {}
+            for status_result in aggregate_status_results:
+                group_name = status_result["group_name"]
+                # Only true when both a user group and singleton group report a status. We use the
+                # one that is not unknown.
+                if (
+                    group_name in group_statuses
+                    and status_result["status"]["scheduler_status"] == JobStatus.unknown
+                ):
+                    continue
```
bdice, line 2703: Could this be flipped to `break` when the one that is not unknown is found? If there is only one, we should break after finding it, rather than `continue` on all the other cases, to avoid doing more iterations than needed. I can't quite wrap my head around whether this is equivalent or not, without testing this manually and examining the data structures.

flow/util/template_filters.py
```diff
@@ -356,3 +356,13 @@ def get_account_name(environment, required=False):
                 file=sys.stderr,
             )
             _GET_ACCOUNT_NAME_MESSAGES_SHOWN.add(env_name)
+
+
+def format_operation_name(operation_name, project):
```
bdice, line 361: Rather than calling this filter in the Jinja templates, could we call it as a pre-processing step someplace like here, replacing `group_name` with something like `format_operation_name(group_name)` where this becomes a method declared inline inside of `FlowProject.print_status`? https://github.com/glotzerlab/signac-flow/blob/50bc29eec23b8bb2c8c277b10dfefd9c7afbdc03/flow/project.py#L3100

My goal is to minimize the complexity of the Jinja templates, at the cost of making `print_status` have another helper function (it already has several).

bdice (commented):
I have some comments attached. @b-butler If you're targeting the 0.18 release, we'll want to merge this soon. I think it's close to ready. cc: @kidrahahjo as release manager.

flow/project.py
b-butler: The break is not equivalent as we are iterating over the status with multiple operations. This check just ensures that we don't overwrite the `group_statuses` dict when the status from a user group or singleton group would overwrite the status of the other with a `JobStatus.unknown` potentially losing information.

flow/project.py
```diff
@@ -2922,10 +2960,24 @@ def _has_any_eligible_group(status_entry):
...
+            """Return the operation name or group name with number of operations."""
+            # If the name is from a group that is not an operation, we append the
+            # number of operations to its name in the status.
+            if group_name not in self._operations:
+                num_operations = len(self._groups[group_name].operations)
+                return f"{group_name} ({num_operations} ops)"
+            return group_name
+
+        statuses = {}
+        # We store the name for display in statuses[aggregate_id][group_name][display_name] to
```
bdice, line 2973: The last key is not a variable.
```suggestion
        # We store the name for display in statuses[aggregate_id][group_name]["display_name"] to
```

bdice (approved):
I have one small comment. Otherwise we just need to measure performance. I don’t expect the performance to change dramatically but it would be good to verify that. See #399 for an example project and #416 for a summary of the benchmarks I ran last time I was working on this kind of code. I haven’t finished porting signac-flow to use asv yet.

b-butler:
@bdice The performance for your default test project in #399 seems to be identical.

bdice:
@csadorf Please take a peek when you're able to review. @b-butler I think this is ready to merge with a second approval.

changelog.txt
```diff
@@ -19,6 +19,7 @@ Added
 Changed
 +++++++
 
+- Add user defined ``FlowGroups`` to status output (#593, #547).
```
csadorf, line 22:
```suggestion
- Add user defined ``FlowGroups`` to status output (#547, #593).
```

csadorf (approved):
It looks fine to me, but I did not test it.

csadorf:
I did not see this before, but is the patch sufficiently tested? codecov/patch is complaining.

b-butler:
@csadorf the code coverage is due to status test not testing submission status at all. This means code that adds group statuses can't be tested currently. I think we should eventually add these tests but I think doing so in another PR so this makes it through for the next release makes the most sense.

PR #601, opened 2022-01-10
Raise `UserOperationError` for @flow.cmd operations

### LINKED ISSUE #571: @flow.cmd operations should raise `UserOperationError` like Python operations.
It looks like the execution logic for `run` will raise `UserOperationError` if a Python operation fails, but not a `@flow.cmd` operation.

_Originally posted by @bdice in https://github.com/glotzerlab/signac-flow/pull/508#r696881917_

kidrahahjo:
## Description
If a Python operation fails, signac-flow will now throw `UserOperationError` for every type of operation. 
Previously, this wasn't consistent with @flow.cmd operations.

## Motivation and Context
Fixes #571 

## Types of Changes

- [ ] Documentation update
- [x] Bug fix
- [ ] New feature

## Checklist:

If necessary:


tests/test_project.py
```diff
@@ -442,6 +447,39 @@ def op2(job):
...
+        def test_with_job_cmd_error(job):
+            raise Exception
+
+        project = self.mock_project(A)
+        with add_cwd_to_environment_pythonpath():
+            with switch_to_directory(project.root_directory()):
+                with pytest.raises(UserOperationError):
+                    starting_dir = os.getcwd()
+                    with redirect_stderr(StringIO()):
+                        project.run()
```
b-butler, line 480: This should stop after the first error. In other words, the other operations would not be tested. You can either use the `operation` argument in `run` or create multiple tests.

In other words, leave the two tests below and just change the error type for the `cmd` one.

b-butler (changes requested):
I would prefer to see the `subprocess.run` wrapped in a `try` block checking for a `CalledProcessError` instead of wrapping all the logic in a `try` block. In general, having as small a `try` block as possible is best. This also would decrease the general indent level of most of the code resulting in a smaller diff.k

flow/project.py
```diff
@@ -3321,8 +3321,16 @@ def _execute_operation(self, operation, timeout=None, pretend=False):
...
+            try:
+                with self._run_with_hooks(operation):
+                    subprocess.run(
+                        operation.cmd, shell=True, timeout=timeout, check=True
+                    )
+            except subprocess.CalledProcessError as error:
+                raise UserOperationError(
+                    f"An exception was raised during operation {operation.name} "
+                    f"for job or aggregate with id {get_aggregate_id(operation._jobs)}."
+                ) from error
```
b-butler, line 3333:
```suggestion
            with self._run_with_hooks(operation):
                try:
                    subprocess.run(
                        operation.cmd, shell=True, timeout=timeout, check=True
                    )
                except subprocess.CalledProcessError as error:
                    raise UserOperationError(
                        f"An exception was raised during operation "
                        "{operation.name} for job or aggregate with id "
                        "{get_aggregate_id(operation._jobs)}."
                    ) from error
```

b-butler (changes requested):

flow/project.py
kidrahahjo: This may not read the user-defined error messages, hence the tests would fail eventually. We may need to fix the behaviour of how we show errors to our users. I guess this relates to my comment here https://github.com/glotzerlab/signac-flow/issues/571#issuecomment-1008668958

A quick fix to this problem is simply to change the hooks tests but I guess this can be a good topic of discussion.

b-butler: That is a good point. Looking at the code it seems that the `on_fail` hook catches the error and re-raises it. Thanks for catching that @kidrahahjo!

kidrahahjo:
@b-butler I guess we can merge this now?

b-butler (changes requested):
One small fix and I will approve.

bdice (approved):
Nice work @kidrahahjo. I have resolved @b-butler's comment and made a slight refactor (defining the error message once for both cmd operations and Python function operations). I approve and will set this to automerge once CI passes.

bdice:
## Test project

```python
import flow
from flow import FlowProject


class Project(FlowProject):
    pass

@flow.cmd
@Project.operation
def asdf(job):
    return "exit 1"

if __name__ == '__main__':
    Project().main()
```

## Traceback before

```
Using environment configuration: StandardEnvironment
ERROR: Encountered error during program execution: 'Command 'exit 1' returned non-zero exit status 1.'

  File "/home/bdice/code/signac-flow/flow/project.py", line 5022, in main
    args.func(args)
  File "/home/bdice/code/signac-flow/flow/project.py", line 4702, in _main_run
    run()
  File "/home/bdice/code/signac-flow/flow/project.py", line 3583, in run
    self._run_operations(
  File "/home/bdice/code/signac-flow/flow/project.py", line 3178, in _run_operations
    self._execute_operation(operation, timeout, pretend)
  File "/home/bdice/code/signac-flow/flow/project.py", line 3325, in _execute_operation
    subprocess.run(operation.cmd, shell=True, timeout=timeout, check=True)
  File "/home/bdice/mambaforge/envs/dice/lib/python3.9/subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command 'exit 1' returned non-zero exit status 1.
```

## Traceback after

```
Using environment configuration: StandardEnvironment
ERROR: An exception was raised during operation asdf for job or aggregate with id 4d82d61b4aebcb2400b9567fc1919941.

  File "/home/bdice/code/signac-flow/flow/project.py", line 3332, in _execute_operation
    subprocess.run(
  File "/home/bdice/mambaforge/envs/dice/lib/python3.9/subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command 'exit 1' returned non-zero exit status 1.

The above exception was the direct cause of the following exception:

  File "/home/bdice/code/signac-flow/flow/project.py", line 5030, in main
    args.func(args)
  File "/home/bdice/code/signac-flow/flow/project.py", line 4710, in _main_run
    run()
  File "/home/bdice/code/signac-flow/flow/project.py", line 3591, in run
    self._run_operations(
  File "/home/bdice/code/signac-flow/flow/project.py", line 3178, in _run_operations
    self._execute_operation(operation, timeout, pretend)
  File "/home/bdice/code/signac-flow/flow/project.py", line 3336, in _execute_operation
    raise UserOperationError(execution_error_message) from error
flow.errors.UserOperationError: An exception was raised during operation asdf for job or aggregate with id 4d82d61b4aebcb2400b9567fc1919941.
```

bdice:
This also allows users of the Python API to reliably catch `UserOperationError` if any operation fails.

Issue #607, opened 2022-02-13
Status check fails for Project with met postconditions


bdice:
### Description

Minimal project with all postconditions met fails in signac-flow 0.18.0.

### To reproduce

Project file:
```python
# project.py
from flow import FlowProject


class Project(FlowProject):
    pass


@Project.operation
@Project.post(lambda job: True)
def op(job):
    pass


if __name__ == "__main__":
    Project().main()
```

### Error output
With `--show-traceback`:
```bash
$ python project.py status
Using environment configuration: StandardEnvironment

Fetching status:   0%|          | 0/1006 [00:00<?, ?it/s]
Fetching status: 100%|██████████| 1006/1006 [00:00<00:00, 458538.34it/s]

Fetching labels:   0%|          | 0/1006 [00:00<?, ?it/s]
Fetching labels: 100%|██████████| 1006/1006 [00:00<00:00, 851901.84it/s]
ERROR:flow.project:Error during status update: 'display_name'
Use '--ignore-errors' to complete the update anyways.
Traceback (most recent call last):
  File "/home/bdice/code/signac-flow/flow/project.py", line 4679, in _main_status
    self.print_status(jobs=aggregates, **args)
  File "/home/bdice/code/signac-flow/flow/project.py", line 3136, in print_status
    display_name = group_status["display_name"]
KeyError: 'display_name'
```

Without `--show-traceback`, I see another issue in how we're handling tracebacks when the traceback originates in signac-flow rather than in user code:
```bash
ERROR:flow.project:Error during status update: 'display_name'
Use '--ignore-errors' to complete the update anyways or '--show-traceback' to show the full traceback.
ERROR: Encountered error during program execution: ''NoneType' object has no attribute '__traceback__''
```

### System configuration

Please complete the following information:

 - Operating System [e.g. macOS]: Ubuntu 20.04
 - Version of Python [e.g. 3.7]: 3.8.10
 - Version of signac [e.g. 1.0]: 1.7.0
 - Version of signac-flow: 0.18.0

PR #608, opened 2022-02-13
Fix traceback handling for status checks

bdice:
## Description
This PR fixes an error in our custom traceback handling for status checks. The problem occurs when an error is raised from signac-flow's internals, rather than from user code.

**Once this PR and #609 are merged, we should create a patch release v0.18.1.**

## Motivation and Context
While investigating #607, I noticed the traceback handling for status checks had an error. It tried to get the `__cause__` from an error that was occurring in signac-flow (rather than a `UserConditionError` or `UserOperationError`, which use `raise ... from` to indicate that a user error is the underlying cause). However, the `__cause__` is `None` if the error is raised internally and not the result of a user error. This made the traceback say:

```bash
ERROR: Encountered error during program execution: ''NoneType' object has no attribute '__traceback__''
```

I also simplified our other traceback handling code so that the logic is slightly easier to read and reduced duplication of error messages / traceback printing. I verified these changes locally.

## Types of Changes
- [x] Bug fix

## Checklist:

If necessary:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -4689,7 +4689,9 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
                     "complete the update anyways or '--show-traceback' to show "
                     "the full traceback."
                 )
-                error = error.__cause__  # Always show the user traceback cause.
+                if isinstance(error, (UserOperationError, UserConditionError)):
+                    # Always show the user traceback cause.
+                    error = error.__cause__
             traceback.print_exception(type(error), error, error.__traceback__)
         else:
             if aggregates is None:
@@ -5060,21 +5062,16 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
         logging.basicConfig(level=max(0, logging.WARNING - 10 * args.verbose))
 
         def _show_traceback_and_exit(error):
-            if args.show_traceback:
-                traceback.print_exception(type(error), error, error.__traceback__)
-            elif isinstance(error, (UserOperationError, UserConditionError)):
+            is_user_error = isinstance(error, (UserOperationError, UserConditionError))
+            if is_user_error:
                 # Always show the user traceback cause.
                 error = error.__cause__
+            if args.show_traceback or is_user_error:
                 traceback.print_exception(type(error), error, error.__traceback__)
+            if not args.show_traceback:
                 print(
                     "Execute with '--show-traceback' or '--debug' to show the "
                     "full traceback.",
-                    file=sys.stderr,
-                )
-            else:
-                print(
-                    "Execute with '--show-traceback' or '--debug' to get more "
-                    "information.",
                     file=sys.stderr,
                 )
             sys.exit(1)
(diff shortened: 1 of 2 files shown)


flow/project.py
```diff
@@ -4689,7 +4689,9 @@ def _main_status(self, args):
                     "complete the update anyways or '--show-traceback' to show "
                     "the full traceback."
                 )
-                error = error.__cause__  # Always show the user traceback cause.
+                if isinstance(error, (UserOperationError, UserConditionError)):
+                    # Always show the user traceback cause.
+                    error = error.__cause__
```
vyasr, line 4694: Can we add a test for an invalid user operation? I'm kind of surprised that we don't have one, maybe this missing coverage is erroneous?

vyasr (changes requested):

flow/project.py
bdice: We don't have a test for this. I think we'd have to create a new test class with a `FlowProject` subclass that raises an error in a condition function, and then assert that some substring is in the error output, which takes some work. That felt messy and prone to breaking so I skipped adding proper tests.

bdice: I spent 20 minutes trying to mock this -- it's messy. I think I'd need to add a new `define_project_invalid_condition.py` and configure its entrypoint and some other steps just for this test of tracebacks. I think this has to be triggered through the `Project.main()` interface, so a self-contained Python test class won't cut it (unless I'm missing something easier).

vyasr: Why do you think it would be prone to breaking? I think your description of what test would need to be is correct, but that doesn't feel any harder than what you added in #609. You'd just have to define the operation as raise an exception (just `assert False`) and then use try-except in the test, right? We've had multiple recent issues related to exception handling with user operations so I think it's important for us to improve our coverage here if we can.

csadorf: We're probably not taking good advantage of pytest's powerful fixture system at the moment, because the test framework was setup before we switched to pytest. Rather than wasting a lot effort on trying to get this to work, maybe we should consider to refactor the test framewor to use a more module and fixture hierarchy that should make it easier to create these kind of scenarios. What do you think?

csadorf (approved):

flow/project.py
kidrahahjo: @bdice  I'll be happy to try refactoring the testing framework!

b-butler (approved):
LGTM. We should create an issue (perhaps several) which seeks to take full advantage of pytest's features.

flow/project.py
csadorf: @kidrahahjo I think that would be great, but we should make sure it's a targeted effort. Would you mind doing a brief analysis of what could potentially be improved and then record that in an issue before making any changes?

vyasr: As far as testing this specific case, @bdice your analysis of how you need to write this test is correct, and if you've attempted it and found it challenging I am fine postponing. I agree that any refactoring should be discussed before we move forward. Since the vast majority of signac-flow users employ our CLI, though, I think improving both our test coverage and our ability to test these kind of CLI-only issues is valuable and worth exploring.

vyasr (approved):
We'll skip the test for now.

Issue #611, opened 2022-02-16
Refactor testing code


bdice:
### Motivation

While working on PR #608, I had some difficulty writing tests for traceback handlers. That sparked a [discussion](https://github.com/glotzerlab/signac-flow/pull/608#discussion_r805515803) about how we could improve our testing code in signac-flow. This is a meta-issue meant for discussion of potential improvements to the testing code.

### Potential Improvements

(Please suggest other improvements in comments on this issue. We can identify which are most actionable and create separate issues / PRs to address.)

- Both status tests and template tests have a complicated set of scripts to create outputs for the current code and compare them to data stored in a tarball (for identifying regressions). Maybe we can unify the code between status tests and template tests, and minimize duplication.
- Testing the signac-flow CLI (e.g. python project.py run) currently requires a file on disk (e.g. define_test_project.py) that can be passed as an entrypoint. This involves a lot of mocking code to create an environment that acts as if it a user FlowProject acting on user data. I wonder if we can do something that is more self-contained or reduces the amount of mocking needed. Perhaps even consolidating those define_project modules into a subfolder would help.

csadorf:
Here are some thoughts on how we could approach this:

1. Use a highly modular fixture hierarchy for everything. The fixture concept is the most powerful pytest feature IMO and we should take full advantage of that. I would identify the tests with complicated "data" setups and then start replacing them one by one.
1. Fixtures should be modular. A `project` fixture should be built on top of `job` fixtures, which should be built on `docs` and `statepoint` fixtures which are built with individual data fixtures and so on.
1. Fixtures that are used across all test modules should be defined in `conftest.py`. Fixtures that are autoused and never explicitly used in the tests functions should be private.
1. Essentially all tests that are operating on data structures (individual documents, jobs, projects) should use standard (parameterized) fixtures. That has the huge advantage that any test that is operating on some kind of job, can automatically be be tested against all kinds of jobs with varying data structures. Filter queries can be fixtures, too!
1. Isolate the test environment. A fixture with monkeypatch and autouse can be employed to automatically create an isolated test environment.
1. Use a cli-runner fixture for CLI tests. The click project provides the [CLIRunner](https://click.palletsprojects.com/en/8.0.x/testing/#basic-testing) which makes it very easy to write CLI tests. Unfortunately the previous attempt to refactor our CLI with click failed, but I would hope that there is a similar fixture provided by a pytest plugin that provides a similar functionality (if not we should write it).

[...]

PR #645, opened 2022-06-01
Refactor Aggregation Tests

kidrahahjo:
## Description
Refactor tests written for `aggregates.py`

## Motivation and Context

Refer to #611 

## Checklist:


b-butler:
@kidrahahjo I like the idea of making our test files more modular, but I would caution against using a directory named `common`, `misc`, etc. as these names are fairly ambiguous. I think that matching the module structure generally works best. Here where we have one monster file, I think splitting these tests into multiple files is fine.

kidrahahjo:
This PR should then do some additional work as well. I think refactoring aggregation tests would be great for the scope of this PR.

requirements/requirements-test.txt
```diff
@@ -1,5 +1,5 @@
 click==8.1.3
-coverage==6.3.2
+coverage==6.4
```
kidrahahjo, line 2:
```suggestion
coverage==6.3.2
```

requirements/requirements-dev.txt
```diff
@@ -1,3 +1,3 @@
 click>=7.0
 ruamel.yaml>=0.16.12
-pre-commit==2.18.1
+pre-commit==2.19.0
```
kidrahahjo, line 3:
```suggestion
pre-commit==2.18.1
```

kidrahahjo:
Requesting @b-butler to review this, since I've written most of the aggregation tests with him.

tests/test_aggregates.py
```diff
@@ -1,362 +1,375 @@
...
+    def test_groupby_with_valid_default_key_for_Iterable(self, project):
+        aggregate_store = self.create_aggregate_store(
+            aggregator.groupby(["half", "even"], default=[-1, -1]), project
+        )
+        assert len(aggregate_store) == 6
+        for aggregate in aggregate_store.values():
+            half = aggregate[0].sp.get("half", -1)
+            even = aggregate[0].sp.get("even", -1)
+            assert all(
+                half == job.sp.get("half", -1) and even == job.sp.get("even", -1)
```
b-butler, line 186:
```suggestion
                half == job.sp.get("half", -1) and even == job.sp["even"]

```

tests/test_aggregates.py
```diff
@@ -1,362 +1,375 @@
...
+        def keyfunction(job):
+            return job.sp["half"]
+
+        with pytest.raises(KeyError):
+            # We will attempt to generate aggregates but will fail in
+            # doing so due to the invalid key
+            self.create_aggregate_store(aggregator.groupby(keyfunction), project)
+
+    def test_valid_select(self, project):
+        def _select(job):
```
b-butler, line 212: No need for underline.

tests/test_aggregates.py
```diff
@@ -1,362 +1,375 @@
...
+            return job.sp.i > 5
+
+        aggregate_store = self.create_aggregate_store(
+            aggregator.groupsof(1, select=_select), project
+        )
+        selected_jobs = []
+        for job in project:
+            if _select(job):
+                selected_jobs.append((job,))
+        assert self.get_aggregates_from_store(aggregate_store) == tuple(selected_jobs)
```
b-butler, line 222:
```suggestion
        selected_jobs = tuple(job for job in project if _selected(job))
        assert self.get_aggregates_from_store(aggregate_store) == selected_jobs
```

tests/test_aggregates.py
```diff
@@ -1,362 +1,375 @@
...
+    def test_store_hashing(self, project):
+        # Since we need to store groups on a per aggregate basis in the project,
+        # we need to be sure that the aggregates are hashing and compared correctly.
+        # This test ensures this feature.
+        total_aggregators = AggregateFixtures.list_of_aggregators()
+        list_of_stores = [
+            self.create_aggregate_store(aggregator, project)
+            for aggregator in total_aggregators
+        ]
+        assert len(list_of_stores) == len(total_aggregators)
```
b-butler, line 233: This is guaranteed to be true
```suggestion
```

tests/test_aggregates.py
```diff
@@ -1,362 +1,375 @@
...
+
+    def test_store_hashing(self, project):
+        # Since we need to store groups on a per aggregate basis in the project,
+        # we need to be sure that the aggregates are hashing and compared correctly.
+        # This test ensures this feature.
+        total_aggregators = AggregateFixtures.list_of_aggregators()
+        list_of_stores = [
+            self.create_aggregate_store(aggregator, project)
+            for aggregator in total_aggregators
+        ]
```
b-butler, line 232:
```suggestion
        list_of_stores = [
            self.create_aggregate_store(aggregator, project)
            for aggregator in AggregateFixtures.list_of_aggregators()
        ]
```

tests/test_aggregates.py
```diff
@@ -1,362 +1,375 @@
...
+        selected_jobs = []
+        for job in project:
+            if _select(job):
+                selected_jobs.append((job,))
+        assert self.get_aggregates_from_store(aggregate_store) == tuple(selected_jobs)
+
+    def test_store_hashing(self, project):
+        # Since we need to store groups on a per aggregate basis in the project,
+        # we need to be sure that the aggregates are hashing and compared correctly.
+        # This test ensures this feature.
```
b-butler, line 227:
```suggestion
```

tests/test_aggregates.py
```diff
@@ -1,362 +1,375 @@
...
+        "aggregator_instance", AggregateFixtures.list_of_aggregators()
+    )
+    def test_aggregates_are_tuples(self, project, aggregator_instance):
+        # This test ensures that all aggregator functions return tuples. All
+        # aggregate stores are expected to return tuples for their values, but
+        # this also tests that the aggregator functions (groupsof, groupby) are
+        # generating tuples internally.
+        aggregate_store = self.create_aggregate_store(aggregator_instance, project)
+        if not isinstance(aggregate_store, _DefaultAggregateStore):
+            for aggregate in aggregate_store._generate_aggregates():
```
b-butler, line 253: On creation of the store, we register/compute the aggregates, so we don't need separate branches.

tests/test_aggregates.py
```diff
@@ -1,362 +1,375 @@
...
+                assert isinstance(aggregate, tuple)
+                assert all(isinstance(job, signac.contrib.job.Job) for job in aggregate)
+        for aggregate in aggregate_store.values():
+            assert isinstance(aggregate, tuple)
+            assert all(isinstance(job, signac.contrib.job.Job) for job in aggregate)
```
b-butler, line 258:
```suggestion
        for aggregate in aggregate_store.values():
            assert isinstance(aggregate, tuple)
            assert all(isinstance(job, signac.contrib.job.Job) for job in aggregate)
```

tests/test_aggregates.py
```diff
@@ -1,362 +1,375 @@
...
+    @pytest.mark.parametrize("agg_value", [(1, 2, 3, 4, 5), (), [1, 2, 3, 4, 5], []])
+    def test_default_init(self, agg_value):
+        aggregate_instance = aggregator()
+        assert not aggregate_instance._is_default_aggregator
+        assert aggregate_instance._sort_by is None
+        assert aggregate_instance._sort_ascending
+        assert aggregate_instance._select is None
+        assert list(aggregate_instance._aggregator_function(agg_value)) == [
+            tuple(agg_value)
+        ]
```
b-butler, line 305:
```suggestion
        assert next(aggregate_instance._aggregator_function(agg_value)) == tuple(agg_value)
```
Also, this check seems a bit off to me. I get that we just wrap anything passed in a tuple by default, but this isn't really what the default aggregator is for. At the very least we should comment why this check exists.

tests/test_aggregates.py
```diff
@@ -1,362 +1,375 @@
...
+        with pytest.raises(TypeError):
+            aggregator(select=select)
+
+    @pytest.mark.parametrize("param", ["str", 1, None])
+    def test_invalid_call(self, param):
+        aggregator_instance = aggregator()
+        with pytest.raises(FlowProjectDefinitionError):
+            aggregator_instance(param)
+
+    def test_call_without_decorator(self):
```
b-butler, line 328:
```suggestion
    def test_call_without_argument(self):

```

tcmoore3 (approved):
Looks good. Approving since addressing Brandon's comments should be sufficient.

tests/test_aggregates.py
kidrahahjo: I don't think we should be choosing the first element here using `next` because we need to make sure that the output of `_aggregator_function` is everything grouped into 1.

tests/test_aggregates.py
```diff
@@ -1,362 +1,375 @@
...
+            aggregator.groupsof(4),
+            aggregator.groupby("even"),
+            aggregator.groupby("even"),
+            aggregator.groupby("half", -1),
+            aggregator.groupby("half", -1),
+            aggregator.groupby(["half", "even"], default=[-1, -1]),
+        ]
+
+    def create_aggregate_store(self, aggregator_instance, project):
+        return aggregator_instance._create_AggregateStore(project)
```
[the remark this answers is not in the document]
kidrahahjo: I could call it directly, but it makes the code look more clean IMO. Happy to change this if you want

tests/test_aggregates.py
```diff
@@ -1,362 +1,375 @@
...
+    def _get_all_job_aggregate(cls, jobs):
+        return (tuple(jobs),)
+
+    @pytest.fixture
+    def get_single_job_aggregate(self):
+        return AggregateFixtures._get_single_job_aggregate
+
+    @pytest.fixture
+    def get_all_job_aggregate(self):
+        return AggregateFixtures._get_all_job_aggregate
```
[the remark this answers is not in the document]
kidrahahjo: This was done in so that we could easily compute things in `list_of_aggregators` below without duplicating code.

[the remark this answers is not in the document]
b-butler: If using fixtures, fixtures can consume other fixtures, meaning you can do this without any code duplication.

tests/test_aggregates.py
b-butler: I get your thoughts, just testing the aggregator on random input feels off. We also don't need to parameterize this test.

tests/test_aggregates.py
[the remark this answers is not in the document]
kidrahahjo: I think that's right, I forgot about fixtures consuming other features. I will make that change.

tests/test_aggregates.py
kidrahahjo: Hmm, let's remove the parameters but are you suggesting to remove this test instead?

kidrahahjo: I removed the parameters, let me know how it looks

tests/test_aggregates.py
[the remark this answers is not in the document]
kidrahahjo: The reason I'm stating this though is that if I remove this then I'll have to add multiple lines of code for initializing the aggregator_instance first. It might be painful to do that but if the current one is against best practices then I'm happy to change this.
@b-butler what do you suggest?

vyasr:
@b-butler how does everything here look to you now? It looks like you're pretty deep into the review already and it's probably easier for you to finish than to bring in anyone else, but feel free to ping if it needs another eye.

b-butler (approved):
:+1:

kidrahahjo:
Merging this as it has two approvals.

PR #650, opened 2022-06-16
Allow for options to be passed to exec through groups

b-butler:
## Description
Enables execution of exec commands with arbitrary options appended through the `FlowGroup` attribute `run_options`. Also renames `options` to `submit_options`.

## Motivation and Context
To debug or pass flags to `project.py exec` is very difficult currently. This simplifies the process greatly which enhances our usefulness for containerized workflows.

## Checklist:

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -651,18 +651,24 @@ class FlowGroupEntry:
         The name of the :class:`FlowGroup` to be created.
     project : flow.FlowProject
         The project the group is associated with.
-    options : str
+    submit_options : str
         The :meth:`FlowProject.run` options to pass when submitting the group.
         These will be included in all submissions. Submissions use run
         commands to execute.
+    run_options : str
+        The options to pass to ``entrypoint exec`` when running the group. These will only be used
+        when the operation is forked instead of running in the same Python interpretor.
     group_aggregator : :class:`~.aggregator`
         aggregator object associated with the :class:`FlowGroup` (Default value = None).
     """
 
-    def __init__(self, name, project, options="", group_aggregator=None):
+    def __init__(
+        self, name, project, submit_options="", run_options="", group_aggregator=None
+    ):
         self.name = name
         self._project = project
-        self.options = options
+        self.submit_options = submit_options
+        self.run_options = run_options
         # We register aggregators associated with operation functions in
         # `_register_groups` and we do not set the aggregator explicitly.
         # We delay setting the aggregator because we do not restrict the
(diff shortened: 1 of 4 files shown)


flow/project.py
```diff
@@ -785,17 +791,26 @@ def op2(job):
...
         used. To prevent this, set the directives to an empty dictionary for
         that operation.
-    options : str
-        A string of options to append to the output of the object's call method.
-        This allows options like ``--num_passes`` to be given to a group.
-
+    run_options : str
+        The options to pass to ``entrypoint exec`` when running the group. These will only be used
+        when the operation is forked instead of running in the same Python interpretor.
+    group_aggregator : :class:`~.aggregator`
```
bdice, line 797: I think you copied the wrong docstrings from above? `submit_options` is missing and `group_aggregator` shouldn't be listed here.

flow/project.py
```diff
@@ -785,17 +791,26 @@ def op2(job):
         the directives of the singleton group containing that operation are
         used. To prevent this, set the directives to an empty dictionary for
         that operation.
-    options : str
-        A string of options to append to the output of the object's call method.
-        This allows options like ``--num_passes`` to be given to a group.
-
+    run_options : str
+        The options to pass to ``entrypoint exec`` when running the group. These will only be used
+        when the operation is forked instead of running in the same Python interpretor.
```
bdice, line 796: Alternatively, should we alter the "should fork?" logic to always fork if `run_options` are supplied?

bdice (commented):
This is pretty much what I had in mind! A couple of suggestions are attached.

flow/project.py
b-butler: Yep, I did. I am currently fighting off COVID, so I am not at peak mental form. :laughing:

flow/project.py
b-butler: That is an interesting question. We could definitely do this, and I almost implemented this behavior. I can see the argument that this is a promise, so we should fork if provided.

flow/project.py
```diff
@@ -651,18 +651,26 @@ class FlowGroupEntry:
...
-    options : str
+    submit_options : str
         The :meth:`FlowProject.run` options to pass when submitting the group.
         These will be included in all submissions. Submissions use run
         commands to execute.
+    run_options : str
+        The options to pass to ``entrypoint exec`` when running the group. Specifying this will
+        cause the operation to be forked even if it otherwise would run in the current Python
+        interpreter.
+        when the operation is forked instead of running in the same Python interpretor.
```
bdice, line 662:
```suggestion
```

flow/project.py
```diff
@@ -840,33 +859,36 @@ def _resolve_directives(self, name, defaults, env):
...
 
     def __iter__(self):
         yield from self.operations.values()
 
     def __repr__(self):
         return (
             "{type}(name='{name}', operations='{operations}', "
-            "operation_directives={directives}, options='{options}')".format(
+            "operation_directives={directives}, submit_options='{submit_options}', "
+            "run_options='{run_options}')".format(
```
bdice, line 885: Generally you want to call `repr` to generate representations of data when it's used in a `__repr__`. For example, the single quotes wrapping `'{submit_options}'` won't be correct if the string itself contains single quotes. Using `repr` will handle quoting/escaping the string correctly for you. This example is kind of unusual because we don't generate a repr that is literally interpretable -- the string list of operations breaks that.

We probably ought to replace `name='{name'}` with `name={repr(name)}` and similar for the other string arguments like `submit_options`, `run_options`.

bdice (approved):
Great. I have a couple small comments but otherwise LGTM.

PR #666, opened 2022-08-20
Fix formatting for cmd operations

### LINKED ISSUE #630: Incorrect command formatting in @flow.cmd
### Description

When testing the `@flow.cmd` operator I noticed something strange in the docs. The example code shows the following:
```python
@FlowProject.operation
@flow.cmd
def compute_volume(job):
    return "matlab -r 'prog {job.sp.foo} {job.sp.bar}' > {job.ws}/output.txt"
```
I noticed that the returned string is not an f-string, so whatever function calls `compute_volume` has the job of populating the command arguments.  That function is the `decorated()` function nested inside `flow.operations.with_job()`. The example works because both `decorated` and `compute_volume` name the job parameter `job`, but if I change the job parameter name for e.g. `j` and change the function to:
```python
@FlowProject.operation
@flow.cmd
def compute_volume(j):
    return "matlab -r 'prog {j.sp.foo} {j.sp.bar}' > {j.ws}/output.txt"
```
Then the code fails because `j` is not a local variable to `decorated`, although it would be difficult for a user to notice the problem from the traceback below.

I think the expected behavior of `@flow.cmd` is that the operation must return the full command already populated, as the user cannot know what the job parameter in the calling function is named (or, in that case, it should be documented). The example code should reflect that and return a fully formatted string (just adding an `f` right before the return value). Additionally, we should avoid the calling function from formatting the returned value, protecting it somehow.

I'm willing to implement and submit a fix but I wanted to know your opinions on how it should be resolved first.

### To reproduce

* Create a project and add a job (e.g. with statepoint `{"test_parameter": 1}`).
* Create a project file "project.py" containing the following:
```python
from flow import FlowProject
import flow

@FlowProject.operation
@flow.cmd
def testop(j):
  return 'echo "{j.sp.a}"'

if __name__ == '__main__':
  FlowProject().main()
```
* Run with traceback: `python project.py run --show-traceback`

Note that if the return value of `testop` is changed to `'echo "{job.sp.a}"'` (no need to change the name of the job parameter, mind you), the error disappears.

### Error output

The error traceback:
```
ERROR: Encountered error during program execution: ''j''

  File "/home/javierbg/miniconda3/envs/python39/lib/python3.9/site-packages/flow/project.py", line 4870, in main
    args.func(args)
  File "/home/javierbg/miniconda3/envs/python39/lib/python3.9/site-packages/flow/project.py", line 4550, in _main_run
    run()
  File "/home/javierbg/miniconda3/envs/python39/lib/python3.9/site-packages/flow/project.py", line 3438, in run
    self._run_operations(
  File "/home/javierbg/miniconda3/envs/python39/lib/python3.9/site-packages/flow/project.py", line 3058, in _run_operations
    self._execute_operation(operation, timeout, pretend)
  File "/home/javierbg/miniconda3/envs/python39/lib/python3.9/site-packages/flow/project.py", line 3179, in _execute_operation
    operation.cmd,
  File "/home/javierbg/miniconda3/envs/python39/lib/python3.9/site-packages/flow/project.py", line 356, in cmd
    return self._cmd()
  File "/home/javierbg/miniconda3/envs/python39/lib/python3.9/site-packages/flow/project.py", line 838, in _run_cmd
    return operation(*jobs).lstrip()
  File "/home/javierbg/miniconda3/envs/python39/lib/python3.9/site-packages/flow/project.py", line 576, in __call__
    return cmd.format(**format_arguments)
KeyError: 'j'
```

### System configuration

 - Operating System: Debian 10 (buster)
 - Version of Python: 3.9
 - Version of signac: 1.7.0
 - Version of signac-flow: 0.17.0

kidrahahjo:
## Description
`FlowCmdOperation` assumes that the command returned by the operation will either contain `job` or `jobs` to be formatted but as noted in #630 there may be cases when users want to use abbreviations instead. This PR fixes the formatting for `FlowCmdOperation`.

## Motivation and Context
Fixes #630 

## Checklist:


kidrahahjo:
Aaaahhhhhh! Black and flake8 are not compatible, wow! Black adds space before `:` and Flake8 throws error, nice :)

flow/operations.py
```diff
@@ -101,12 +101,13 @@ def hello_cmd(job):
         )
 
     @wraps(func)
-    def decorated(*jobs):
-        with jobs[0] as job:
+    def decorated(__flow_internal_with_job_argument):
```
kidrahahjo, line 104: this was done because when `cmd` is used with `with_job`, the formatting logic assumes that user specified `*jobs` variable argument, but that's us not the user :)

flow/project.py
```diff
@@ -556,7 +559,7 @@ class FlowCmdOperation(BaseFlowOperation):
     ----------
     cmd : str or callable
         The command to execute the operation. Callable values will be
-        provided one or more positional arguments (``*jobs``) that are
```
kidrahahjo, line 559: should not necessarily be `*jobs`

kidrahahjo:
Note: We can also deprecate the whole formatting thing in this PR and remove later. This one just fixes the things which are already present in the code.

mikemhenry:
You are correct, we need to adjust the flake8 settings to ignore E203 https://black.readthedocs.io/en/stable/faq.html#why-are-flake8-s-e203-and-w503-violated

vyasr:
@kidrahahjo @mikemhenry now that #670 is merged and our style checks are passing is this PR ready for review/merge? The pre-commit flake8 check is fine now.

kidrahahjo:
@vyasr yes, this PR is ready to review

flow/operations.py
bdice: I don't understand this change. If you call `func(*jobs)` with a single argument like `func(job)`, it will get `jobs = [job]` as an argument.

flow/project.py
```diff
@@ -578,9 +581,60 @@ def __str__(self):
...
+                format_arguments[argspec.varargs] = jobs[len(args) :]
+
+            if format_arguments.get("jobs", None) is None:
+                if re.search("{jobs}", cmd) or re.search(r"{jobs\..*}", cmd):
+                    warnings.warn(
+                        "Argument names for the flow cmd operations must match that of "
+                        "in the function definition. Auto assignment of the jobs "
+                        "argument has been deprecated as of 0.21.0 "
+                        "and will be removed in 0.23.0.",
+                        DeprecationWarning,
```
bdice, line 617: We should use `FutureWarning`s everywhere.

flow/project.py
```diff
@@ -578,9 +581,60 @@ def __str__(self):
...
+                    format_arguments["jobs"] = jobs
+
+            if format_arguments.get("job", None) is None:
+                if len(jobs) == 1 and (
+                    re.search("{job}", cmd) or re.search(r"{job\..*}", cmd)
+                ):
+                    warnings.warn(
+                        "Argument names for the flow cmd operations must match that of "
+                        "in the function definition. Auto assignment of the job "
+                        "argument has been deprecated as of 0.21.0 "
```
bdice, line 628:
```suggestion
                        "argument has been deprecated as of 0.22.0 "
```

bdice (commented):
Initial comments from a quick look -- I'll need to take another pass or two.

flow/project.py
mikemhenry: :+1:

mikemhenry (changes requested):

flow/project.py
```diff
@@ -580,9 +580,54 @@ def __str__(self):
     def __call__(self, *jobs):
         """Return the command formatted with the supplied job(s)."""
         cmd = self._cmd(*jobs) if callable(self._cmd) else self._cmd
-        format_arguments = {"jobs": jobs}
-        if len(jobs) == 1:
-            format_arguments["job"] = jobs[0]
+        # Inconclusive test if cmd is a format string. If {} with anything in the interior is found
+        # though it is guarenteed not to be a format string.
```
bdice, line 584:
```suggestion
        # though it is guaranteed not to be a format string.
```

flow/project.py
```diff
@@ -580,9 +580,54 @@ def __str__(self):
     def __call__(self, *jobs):
         """Return the command formatted with the supplied job(s)."""
         cmd = self._cmd(*jobs) if callable(self._cmd) else self._cmd
-        format_arguments = {"jobs": jobs}
-        if len(jobs) == 1:
-            format_arguments["job"] = jobs[0]
+        # Inconclusive test if cmd is a format string. If {} with anything in the interior is found
```
bdice, line 583: One way we could conclusively test if the string is formatted is by applying the format operation and seeing if it modifies the string. e.g.
```python
cmd_input = "arbitrary string"
formatted = cmd_input.format(**format_arguments)
if cmd_input != formatted:
    warnings.warn("Formatting is deprecated.", FutureWarning)
```

flow/project.py
```diff
@@ -580,9 +580,54 @@ def __str__(self):
...
+        if not callable(self._cmd):
+            format_arguments["jobs"] = jobs
+            if len(jobs) == 1:
+                format_arguments["job"] = jobs[0]
+            return cmd.format(**format_arguments)
+
+        argspec = inspect.getfullargspec(self._cmd)
+        if argspec.varkw or argspec.kwonlyargs:
+            raise RuntimeError(
+                "FlowProject cmd operations doesn't support keyword only arguments"
```
bdice, line 602:
```suggestion
                "FlowProject cmd operations do not support keyword-only arguments."
```

bdice (commented):
I haven't spent enough time to parse the `inspect.signature` bits, but I have an idea that may help with detecting format arguments. I'll revisit later.

flow/project.py
b-butler: Good suggestion.

flow/project.py
```diff
@@ -580,10 +580,60 @@ def __str__(self):
...
+        format_arguments = {}
+        if not callable(self._cmd):
+            format_arguments["jobs"] = jobs
+            if len(jobs) == 1:
+                format_arguments["job"] = jobs[0]
+            formatted_cmd = cmd.format(**format_arguments)
+            if formatted_cmd != cmd:
+                warnings.warn(
+                    "Returning format strings in a cmd operation is deprecated as of version "
+                    "0.22.0 and will be removed in  0.23.0",
```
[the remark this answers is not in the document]
bdice: Same below. (Can this warning and the formatting step be performed in only one place?)

tests/test_project.py
```diff
@@ -355,6 +355,110 @@ def test_context(job):
...
+    def test_cmd_operation_argument_as_command_invalid(self):
+        class A(FlowProject):
+            pass
+
+        @A.operation(with_job=True, cmd=True)
+        def test_cmd(job):
+            return "echo '{jobabc}' > output.txt"
+
+        project = self.mock_project(A)
+        with _add_cwd_to_environment_pythonpath():
```
bdice, line 386: Maybe we should make one context manager that does all this… at least up to the pytest.raises.

flow/project.py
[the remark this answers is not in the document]
b-butler: Yes. I did this twice to avoid nesting the rest of the lines.

flow/project.py
```diff
@@ -580,10 +580,60 @@ def __str__(self):
...
+                    FutureWarning,
+                )
+            return formatted_cmd
+
+        argspec = inspect.getfullargspec(self._cmd)
+        if argspec.varkw or argspec.kwonlyargs:
+            raise RuntimeError(
```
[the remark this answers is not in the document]
b-butler: I am not completely sure to be honest. I think this is what was intended. I can remove it though as it seems restrictive now that you mention it. I was going for quick changes without effecting the logic in my recent refactor.

> Also does all this logic go away when formatting is removed?

Yes, the entire function essentially becomes the first line.

flow/project.py
```diff
@@ -580,10 +580,45 @@ def __str__(self):
     def __call__(self, *jobs):
         """Return the command formatted with the supplied job(s)."""
         cmd = self._cmd(*jobs) if callable(self._cmd) else self._cmd
-        format_arguments = {"jobs": jobs}
-        if len(jobs) == 1:
-            format_arguments["job"] = jobs[0]
-        return cmd.format(**format_arguments)
+        format_arguments = {}
```
bdice, line 583: It might be nice to add a note about how this function should change once the deprecated behavior is removed. There is a fair bit of `inspect` going on, which I don't think is needed, but I also don't want to fight it if we're going to remove it in the next release.

tests/test_project.py
```diff
@@ -56,6 +56,27 @@ def suspend_logging():
         logging.disable(logging.NOTSET)
 
 
+@contextmanager
+def setup_project_subprocess_execution(
+    project, capture_stderr=True, stderr_output=None
```
bdice, line 61: An alternative would be to let the default behavior be something like `stderr_output=sys.stderr` and eliminate the `capture_stderr` flag?

Also "capture" seems like the wrong word if `stderr_output is None`, because the "captured" result is just discarded. Hmm.

Maybe the default can be `None` or `sys.stderr`, and if we don't want the output to be shown or captured at all, we do something like redirect to `os.devnull`?

tests/test_project.py
```diff
@@ -348,12 +369,72 @@ def test_context(job):
...
+            return "echo '{joba} {jobb}' > output.txt"
+
+        project = self.mock_project(A)
+        with setup_project_subprocess_execution(project):
+            project.run()
+        for job in project:
+            assert os.path.isfile(job.fn("output.txt"))
+            with open(job.fn("output.txt")) as f:
+                lines = f.readlines()
+                assert f"{job.id} test\n" == lines[0]
```
bdice, line 392: Try to perform `assert` statements _outside_ the context manager if possible. Here, the `lines` have already been read, so we can close the file before comparing to `lines[0]`.
```suggestion
            assert f"{job.id} test\n" == lines[0]
```

bdice (approved):
I have a few small comments. Please address them however you see fit, otherwise I approve.

flow/project.py
b-butler: Generally, I agree. I feel though that given @kidrahahjo did the work for this fix we should merge it to have for this version.

tests/test_project.py
b-butler: I dislike the amount of `stderrr` capturing we do as it makes debugging tests difficult, so I would be happy to change the default behavior.

mikemhenry (changes requested):
LGTM

mikemhenry (approved):

PR #667, opened 2022-08-22
Enable document parameters in status display

rosecers:
Hey team! This is just a small change to allow rendering document parameters in the status call. Given that people often have large documents, I made the renderer only pull the document if necessary, then used the same verbiage as statepoint pulls. Shouldn't break anything that I'm aware of.

## Motivation and Context
- Right now I'm running a lot of molecular simulations where the molecule size differs. It's good to see in the status call the molecule size (which is stored in the document) because that gives me a relatively good idea of how long the simulation takes. 
- 
## Checklist:

- [X] I am familiar with the [Contributing Guidelines](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md).
- [X] I agree with the terms of the [Contributor Agreement](https://github.com/glotzerlab/signac-flow/blob/master/ContributorAgreement.md).
- [X] The [package documentation](https://github.com/glotzerlab/signac-flow/tree/master/doc) and [framework documentation](https://docs.signac.io/) in [signac-docs](https://github.com/glotzerlab/signac-docs) are up to date with these changes.


kidrahahjo (commented):
Thank you so much @rosecers for this pull request. The changes looks great to me. I only have one question as of why do performance degrade for large documents in the first place. I see that `signac` caches the document for a `Job` and hence we're mostly saved from loading documents everytime we perform something on the lines of `sp.get("doc", None)`. So in your case, if the parameters starting with "doc." is present should only fetch document once and hence this should already be optimised.

@b-butler @bdice can you please have a look at this?

kidrahahjo:
So I investigated more and I see that in `statepoint` we do not explicitly cache document when fetching things from the `__getitem__` method for `_StatePointDict`. Which is maybe why this is causing issues.

kidrahahjo:
Maybe we can change the `dotted_get` function instead? It'd be something on the lines of:
```python
cache = {}
def dotted_get(mapping, key):
    """Fetch a value from a nested mapping using a dotted key."""
    if mapping is None:
        return None
    
    tokens = key.split(".")
    if len(tokens) > 1:
        if tokens[0] in cache:
            value = cache[tokens[0]]
        else:
            value = mapping.get(tokens[0])
        
        cache[tokens[0]] = value
        key = ".".join(tokens[1:])
        return dotted_get(value, ".".join(tokens[1:]))
    else:
        if key in cache:
            return cache[key]

        cache[key] = mapping.get(key)
        return cache[key]
```

b-butler:
If we change `dotted_get` I would want to see something like
```python
def dotted_get(mapping, key, sep="."):
    """Fetch a value from a nested mapping using a dotted key."""
    if mapping is None:
        return None
    
    tokens = key.split(".")
    return _dotted_get(mapping, key)

def _dotted_get(mapping, key): 
    if len(key) > 1:
        if key[0] in cache:
            value = cache[key[0]]
        else:
            value = mapping.get(key[0])
        
        cache[key[0]] = value
        return _dotted_get(value, key[1:])
    else:
        key = key[0]
        if key in cache:
            return cache[key]

        cache[key] = mapping.get(key)
        return cache[key]
```
which removes the need to join the list of keys.

b-butler:
We should also require a status test that uses this in `tests/test_status.py`. It should involve just adding an entry in `tests/generate_status_reference_data.py`.

flow/project.py
```diff
@@ -3065,10 +3066,18 @@ def dotted_get(mapping, key):
...
                 for parameter in parameters:
-                    status["parameters"][parameter] = shorten(
-                        str(self._alias(dotted_get(statepoint, parameter))),
-                        param_max_width,
-                    )
+                    if not parameter.startswith("doc."):
+                        status["parameters"][parameter] = shorten(
+                            str(self._alias(dotted_get(statepoint, parameter))),
+                            param_max_width,
+                        )
```
bdice, line 3073:
```suggestion
                        if statepoint is None:
                            statepoint = job.statepoint()
                        status["parameters"][parameter] = shorten(
                            str(self._alias(dotted_get(statepoint, parameter))),
                            param_max_width,
                        )
```

flow/project.py
```diff
@@ -3053,6 +3053,7 @@ def _add_parameters(status):
                     # TODO: Fill parameters with empty values (or shared values?)
                     raise ValueError("Cannot show parameters for aggregates.")
                 statepoint = self.open_job(id=aggregate_id).statepoint()
+                document = None
```
bdice, line 3056: We should only load the state point if it is needed (and not create a copy of the statepoint via the call operator `job.statepoint()` if only document keys are used).
```suggestion
                job = self.open_job(id=aggregate_id)
                # Cache the job state point and document if used to render status parameters.
                statepoint = None
                document = None
```

flow/project.py
```diff
@@ -3065,10 +3066,18 @@ def dotted_get(mapping, key):
...
-                        param_max_width,
-                    )
+                    if not parameter.startswith("doc."):
+                        status["parameters"][parameter] = shorten(
+                            str(self._alias(dotted_get(statepoint, parameter))),
+                            param_max_width,
+                        )
+                    else:
+                        if document is None:
+                            document = self.open_job(id=aggregate_id).document()
```
bdice, line 3076: Reusing the `job` from above should be faster than opening the job again.
```suggestion
                            document = job.document()
```

bdice (changes requested):
Nice feature! Thanks for the contribution @rosecers! I would suggest a few small refactors that should improve clarity and performance.

flow/project.py
bdice: @kidrahahjo To your question above, the document is not cached by default because it is not "fixed" like the state point. The state point for a Job is meant to be treated as immutable, since signac's storage is based on the hash of that state point. However, we can't guarantee that a document is unchanged between reads. For the scope of rendering status parameters, it is perfectly fine to assume that the document is fixed and cache it to avoid re-reading and parsing the file for every parameter requested.

flow/project.py
bdice: @b-butler @kidrahahjo I would prefer to avoid putting more complexity to the utility function `dotted_get` and instead just use signac's call operators like `job.statepoint()` and `job.document()` to facilitate caching with this extra check for `None`. This should be performant enough and avoids the need for a separate `cache = {}` object.

bdice: What we could do to improve `dotted_get` is adopt a pattern more like [this microbenchmark](https://github.com/bdice/python_microbenchmarks/blob/main/microbenchmarks/nested_access.py). I worked on this The best code is the one used in "recurse all" which avoids the re-joining (which probably incurs a memory allocation for the result string) and also avoids the recursion of `dotted_get`. Here is the timing I see for that microbenchmark:
```
      recurse_all: 0.000182753
one_direct_access: 0.000253401
       dotted_get: 0.000997744
```

bdice: I applied my suggested refactoring in https://github.com/glotzerlab/signac-flow/pull/667/commits/d77269787b4a991a6c1ed18ea8c5244024823cd7. Here's the result for `dotted_get2`, which is the refactored version.

```
      recurse_all: 0.000189232
one_direct_access: 0.00031354
      dotted_get2: 0.000369667
       dotted_get: 0.00102027
```

bdice:
@rosecers Can you update the changelog? See the PR checklist.

flow/project.py
```diff
@@ -3052,23 +3052,37 @@ def _add_parameters(status):
                 if aggregate_id.startswith("agg-"):
                     # TODO: Fill parameters with empty values (or shared values?)
                     raise ValueError("Cannot show parameters for aggregates.")
-                statepoint = self.open_job(id=aggregate_id).statepoint()
+                job = self.open_job(id=aggregate_id)
+                # Cache the job state point and document if used to render status parameters.
+                statepoint = None
+                document = None
 
                 def dotted_get(mapping, key):
```
kidrahahjo, line 3060: This implementation looks great. Thanks @bdice @rosecers!

I am a little confused about the difference between the use of a new `cache` object and this. Should we be relying on `signac`'s caching for internal flow functionality? I am just worried that if a concept other than document is brought in by `signac`, we'd have to add more conditions to keep things optimized.

bdice: Sort of. This "cache" behavior here is needed because we don't want to _fill_ the cache (read the state point / document) unless a key is required from the state point / document. This is operating at a slightly higher level than the signac state point cache and know more information about the keys being requested, so it is reasonable for signac-flow to handle this directly.

flow/project.py
b-butler: @bdice I wasn't suggesting to change it, just if we were I was providing my preference. I agree with your assessment.

cbkerr (changes requested):
Thanks Rosy!

Users would expect both "sp." and "doc." to work by analogy with the [query syntax](https://docs.signac.io/en/latest/query.html). I added the ability to parse the "sp." prefix if present and tested it with a signac-example project.

I would like to see unit tests for these. We basically need to add some lines like this to `generate_status_reference_data.py` [near line 33](https://github.com/glotzerlab/signac-flow/blob/master/tests/generate_status_reference_data.py#L33):
```
        {"parameters": ["sp.a"]}, # -p sp.a
        {"parameters": ["doc.a"]}, # -p doc.a
```
as well as write to the job documents of the test jobs.

I tried adding tests and regenerating the status reference data, but I think there may be a bug in this code because the status info about groups isn't showing up, causing errors.

This is part of what doesn't show up when I try to generate the status reference:
```
[GR]:group_registered [GI]:group_inactive [GS]:group_submitted [GH]:group_held [GQ]:group_queued [GA]:group_active [GE]:group_error
```

changelog.txt
```diff
@@ -5,6 +5,14 @@ Changelog
 The **signac-flow** package follows `semantic versioning <https://semver.org/>`_.
 The numbers in brackets denote the related GitHub issue and/or pull request.
 
+next -- 2022-xx-xx
```
bdice, line 8:
```suggestion
next
====

next -- 2022-xx-xx
```

changelog.txt
```diff
@@ -5,6 +5,17 @@ Changelog
...
+next
+====
+
+next -- 2022-xx-xx
+------------------
+
+Added
++++++
+
+- Added functionality to status call to show document parameters by using flag `-p doc.PARAM`
```
bdice, line 17:
```suggestion
- Added functionality to status call to show document parameters by using flag ``-p doc.PARAM`` (#667).
```

bdice:
I caught and fixed a bug in the status tests that was introduced in the original PR #562. Only the _last_ set of status parameters was actually being tested because the `assert` wasn't inside the `for` loop. https://github.com/glotzerlab/signac-flow/blob/b550ffaa3182160c3207d9927582026d737b488e/tests/test_status.py#L31

bdice (approved):
Thanks @rosecers!

bdice:
@cbkerr I re-requested your review since you requested changes. Please merge if/when you approve.

cbkerr:
I will document this change in the argument parser and then I will merge!

cbkerr (approved):
Looks good now. Thanks everyone!

PR #671, opened 2022-08-27
Allow users to pass with_job and cmd using FlowProject.operation

kidrahahjo:
## Description

To be added

## Motivation and Context

Closes #669 

## Checklist:

- [x] I am familiar with the [Contributing Guidelines](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [Contributor Agreement](https://github.com/glotzerlab/signac-flow/blob/master/ContributorAgreement.md).
- [x] My name is on the [list of contributors](https://github.com/glotzerlab/signac-flow/blob/master/contributors.yaml).
- [ ] The changes introduced by this pull request are covered by existing or newly introduced tests.
- [ ] The [package documentation](https://github.com/glotzerlab/signac-flow/tree/master/doc) and [framework documentation](https://docs.signac.io/) in [signac-docs](https://github.com/glotzerlab/signac-docs) are up to date with these changes.
- [ ] I have updated the [changelog](https://github.com/glotzerlab/signac-flow/blob/master/changelog.txt) and added any related issue and pull request numbers for future reference.

PR #679, opened 2022-10-05
Move cmd and with_job to FlowProject.operation

### LINKED ISSUE #669: Allow users to pass with_job and cmd through operation decorator
### Feature description

This issue is just used as a tracker for this feature.

See https://github.com/glotzerlab/signac-flow/issues/665 for more

b-butler:
## Description
Does not break current behavior.

- refactor: Deprecate cmd and with_job decorators.
- feat: Add cmd and with_job keyword only arguments for @operation
- test: Add tests for new keyword only arguments of FlowProject.operation

## Motivation and Context
Closes #669
and replaces #671.

I had large suggestions to the way the code should be changed. I think this is much simpler to get the desired behavior.

## Checklist:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -15,6 +15,13 @@ Added
 +++++
 
 - The CLI ``status`` command can show document parameters by using flag ``-p doc.PARAM`` (#667).
+- ``FlowProject.operation`` now has ``cmd``, ``with_job``, and ``directives`` keyword only arguments (#679, #655, #669).
+
+Changed
++++++++
+
+- ``@flow.cmd`` and ``flow.with_job`` are deprecated (#679, #669, #665).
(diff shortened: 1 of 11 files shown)


bdice (approved):
LGTM. In the future, it can be helpful to do something like this instead:

```python
def _func(...):
    # do stuff

def func(...):
    warnings.warn(FutureWarning, "func is deprecated")
    _func(...)
```

and then call `_func` internally, instead of using a `warnings.catch_warnings` + `warnings.simplefilter`.

b-butler:
@bdice I agree and it is generally what I do. Here, I think the solution is different in that I think we should remove the implementations in `flow/operations.py` and put them in operation register.

PR #683, opened 2022-10-20
Updated some usages of root_directory to path in flow/project.py

tabedzki:
## Description

Updated some usages of root_directory to path in flow/project.py
This follows the new deprecation warning from 0.22.0

## Motivation and Context

Change required due to deprecation

## Checklist:

- [ ] I am familiar with the [Contributing Guidelines](https://github.com/glotzerlab/signac-flow/blob/master/CONTRIBUTING.md).

### DIFF
--- a/flow/project.py
+++ b/flow/project.py
@@ -91,7 +91,7 @@ variable, for example in the project configuration file. The current template di
 {template_dir}
 
 All template variables can be placed within a template using the standard jinja2
-syntax, e.g., the project root directory can be written as: {{{{ project.root_directory() }}}}.
+syntax, e.g., the project root directory can be written as: {{{{ project.path }}}}.
 The available template variables are:
 {template_vars}
 
@@ -1027,7 +1027,7 @@ class FlowGroup:
         else:
             op_string = operation_name
 
-        root_directory = project.root_directory()
+        root_directory = project.path
         aggregate_id = get_aggregate_id(aggregate)
         full_name = f"{root_directory}%{aggregate_id}%{op_string}"
         # The job_op_id is a hash computed from the unique full name.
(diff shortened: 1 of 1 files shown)


bdice:
@tabedzki Thanks for this contribution! I will need to double check this, but I think we may need to keep the current code for `root_directory()` and silence the warnings for now. signac-flow currently supports signac back to version 1.0.0, which does not have the `path` attribute. Future versions of signac-flow will require `signac>=2.0.0` once that major version is released, which will mitigate this issue as well.

tabedzki:
OK, sounds good! As long as the team is aware of the warning (reproducing it below)
```
/home/tabedzki/miniconda3/envs/ovitos/lib/python3.10/site-packages/flow/project.py:1803: DeprecatedWarning: root_directory is deprecated as of 1.8 and will be removed in 2.0. Use Project.path instead.
  self.root_directory(), self._config.get("template_dir", "templates")
Using environment configuration: PennChestnutCluster
/home/tabedzki/miniconda3/envs/ovitos/lib/python3.10/site-packages/flow/project.py:1030: DeprecatedWarning: root_directory is deprecated as of 1.8 and will be removed in 2.0. Use Project.path instead.
  root_directory = project.root_directory()
```

tommy-waltmann:
@tabedzki Thanks for this! Getting rid of these warnings is a simple and necessary change. I'll defer to @bdice on whether the usage of `root_directory()` should be replaced with `path`, or the warnings should just be silenced.

bdice:
The two options I see are:
- Require signac>=1.8.0 and use the `path` attribute.
- Leave the current signac version pinning and silence the warnings in signac-flow.

I don’t want to make that decision alone - perhaps @b-butler or other @glotzerlab/signac-committers have thoughts.

joaander:
You could conditionally call `root_directory()` or `path` depending on the signac version if you want to keep support for older versions.

b-butler:
I don't have a strong preference. We can capture the warnings for now or require that flow 0.23.0 requires signac `1.8.0`. The version option that @joaander give is also a neat solution. Overall, my slight preferences would be
1. Require signac 1.8.0 and use path.
2. Check signac version
3. Filter warning

b-butler:
Closing due to the merge of #693.

PR #686, opened 2022-10-25
Refactor/remove deprecated code

b-butler:
## Description
Removes deprecated code.

## Motivation and Context
This needs to be done before 0.23.0, and there is no need to wait on the master branch.

## Checklist:


flow/project.py
```diff
@@ -1577,55 +1528,37 @@ def _internal_call(self, func, name, *, cmd, with_job, directives):
...
+                        if getattr(func, "_flow_cmd", False):
+                            return (
+                                f'trap "cd $(pwd)" EXIT && cd {job.ws} && {func(job)}'
+                            )
+                        else:
+                            return func(job)
 
-                return add_operation_with_directives
+                setattr(decorated, "_flow_with_job", True)
+                setattr(decorated, "_flow_cmd", getattr(func, "_flow_cmd", False))
```
bdice, line 1549:
```suggestion
                if not hasattr(decorated, "_flow_cmd"):
                    decorated._flow_cmd = False
```

flow/project.py
```diff
@@ -1577,55 +1528,37 @@ def _internal_call(self, func, name, *, cmd, with_job, directives):
...
+                    with jobs[0] as job:
+                        if getattr(func, "_flow_cmd", False):
+                            return (
+                                f'trap "cd $(pwd)" EXIT && cd {job.ws} && {func(job)}'
+                            )
+                        else:
+                            return func(job)
 
-                return add_operation_with_directives
+                setattr(decorated, "_flow_with_job", True)
```
bdice, line 1548:
```suggestion
                decorated._flow_with_job = True
```

tests/test_project.py
```diff
@@ -772,39 +653,6 @@ def b(job):
             not callable(value) for value in submit_job_operation.directives.values()
         )
 
-    @pytest.mark.filterwarnings("ignore:@FlowProject.operation.with_directives")
-    def test_operation_with_directives(self):
```
bdice, line 776: Does this test need to be updated instead of removed?

b-butler: `with_directives` is just removed so no, since we test directives otherwise.

flow/project.py
b-butler: decorated cannot have the attribute `_flow_cmd` by definition.
```suggestion
                decorated_flow_cmd = getattr(func, "_flow_cmd", False)
```

bdice: My bad. I didn't catch that one argument was `func` and the other was `decorated`.

bdice: Your suggestion looks correct here.

syjlee (approved):

PR #690, opened 2022-11-23
refactor/decorator ordering

b-butler:
## Description
Remove show-traceback completely from CLI and config. Deprecated assigning conditions before registering operations. Update test decorator ordering. Update documentation examples to new decorator ordering.

## Motivation and Context

Fixes a discussion on Slack about `with_job=True` invaliding conditions already specified as we store conditions in a `dict` where function hash is used for the key and `with_job=True` wraps the function. This is also desirable as it requires a more Pythonic order to the decorators and allows potential refactoring in the future to simplify the code.

## Checklist:


csadorf (approved):
Great! Thanks a lot!

shihkual (approved):
Looks good to me!

flow/project.py
```diff
@@ -1287,10 +1286,20 @@ def hi_all(*jobs):
...
             def __call__(self, func):
+                # Have to traverse the mro to ensure that func is already an operation and that
+                # self.condition is not.
                 operation_functions = [
-                    operation[1]
-                    for operation in self._parent_class._collect_operations()
+                    func for name, func in self._parent_class._collect_operations()
                 ]
+                if func not in operation_functions:
+                    _deprecated_warning(
```
bdice, line 1295: This will raise instead of warn in the future, right? In other words, we still have to do this check in the future.

bdice (approved):

flow/project.py
b-butler: Yes or something like it.

PR #698, opened 2022-12-04
Add GitHub Actions.

bdice:
## Description
Migrates signac-flow's CI to use GitHub Actions.

This PR drops coverage for a few things that I think are non-essential:
- Nightly/weekly testing of `pip install signac signac-flow` and `conda install signac signac-flow`
- Testing against the latest commit of `signac` (which is currently disabled because it's broken until we release 2.0)
- Checking Zenodo metadata on `release/.*` branches.

If another contributor wishes to add these to GitHub Actions, that would be fine.

Before merging, a repo administrator will need to update the CI checks used for branch protections.

## Checklist:


.github/workflows/deploy-test-pypi.yml
```diff
@@ -0,0 +1,18 @@
+name: Publish to TestPyPI
+
+on:
+  push:
+    branches:
+      - 'release/*.*.*'
```
joaander, line 6: `publish-packages.yml` is not tested in this PR because it does not match this branch name. You could remove this condition temporarily to test it, or remove it permanently. GitHub Actions provides many runners, so the additional testing time is not an issue.

.github/workflows/run-pytest.yml
```diff
@@ -0,0 +1,55 @@
...
+concurrency:
+  group: "${{ github.workflow }}-${{ github.ref }}"
+  cancel-in-progress: true
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
```
joaander, line 23: Why no testing on windows?

.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,72 @@
...
+        uses: actions/download-artifact@v2
+        with:
+            name: release
+            path: dist/
+      - name: Publish package to TestPyPI
+        if: ${{ inputs.upload_to_test }}
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          repository_url: https://test.pypi.org/legacy/
+          password: ${{ secrets.TEST_PYPI_API_TOKEN }}
```
joaander, line 67: Add the tokens to the repository secrets.

joaander (commented):
Thanks! I have some suggestions to improve the testing.

.github/workflows/deploy-test-pypi.yml
bdice: This script matches the testing workflows on signac-dashboard, which were copied from signac. I think I trust this workflow from those two repos, so I was going to leave it “untested” until 0.23.0 release time. I don’t think we want to publish test packages on every branch/commit.

.github/workflows/run-pytest.yml
bdice: Windows is not fully supported by signac-flow. We have a number of features like command line operations and “with job” that inject UNIX-specific things like trap commands. About 80% of our test suite passes on Windows, if I recall correctly, but tests of job submission and a few other categories fail.

bdice: This is an area for improvement — we should mark some tests as skip on Windows, and/or block those features with explicit errors on Windows. I don’t think we have received many complaints about support, but I know we have some users of the package’s partial functionality on Windows so I don’t want to explicitly disallow installation of the package on Windows as an unsupported platform.

.github/workflows/deploy-pypi.yml
```diff
@@ -0,0 +1,18 @@
...
+  group: "${{ github.workflow }}-${{ github.ref }}"
+  cancel-in-progress: true
+
+jobs:
+  build-and-test-sdist-and-wheels:
+    uses: ./.github/workflows/publish-packages.yml
+    with:
+      upload_to_test: false
+    secrets:
+      PYPI_API_TOKEN: ${{ secrets.PYPI_API_TOKEN }}
```
mikemhenry, line 18: I don't see this secret here: https://github.com/glotzerlab/signac-flow/settings/secrets/actions

.github/workflows/deploy-test-pypi.yml
```diff
@@ -0,0 +1,18 @@
...
+  group: "${{ github.workflow }}-${{ github.ref }}"
+  cancel-in-progress: true
+
+jobs:
+  build-and-test-sdist-and-wheels:
+    uses: ./.github/workflows/publish-packages.yml
+    with:
+      upload_to_test: true
+    secrets:
+      TEST_PYPI_API_TOKEN: ${{ secrets.TEST_PYPI_API_TOKEN }}
```
mikemhenry, line 18: I don't see this secret here: https://github.com/glotzerlab/signac-flow/settings/secrets/actions

.github/workflows/run-pytest.yml
mikemhenry: I agree that we can improve here, but I think we are okay to merge this without it.

.github/workflows/run-pytest.yml
```diff
@@ -0,0 +1,55 @@
...
+      run: |
+        pip install -r .github/workflows/ci-oldest-reqs.txt
+      if: ${{ matrix.config.dependencies == 'oldest' }}
+    - name: Install the package
+      run: |
+        pip install -e .
+    - name: Test with pytest
+      run: |
+        pytest --cov=flow --cov-config=setup.cfg --cov-report=xml tests/ -v
+    - uses: codecov/codecov-action@v3
```
mikemhenry, line 55: Do we want to have CI fail if codecov report upload fails? Sometimes it can be flaky, not sure if we have ran into that with this repo

mikemhenry (approved):
LGTM, just a few notes!

.github/workflows/deploy-pypi.yml
bdice: I added the token now.

Issue #566, opened 2021-08-20
Potential Template Errors


b-butler:
### Description
For the GreatLakes (and potentially others) template. Multi-node GPU submissions incorrectly set the `--ntasks-per-node` to be the total number of tasks disregarding individual node size.

## To Reproduce

Just request a multi-GPU node submission with `--pretend` and view the output. Here is an example

```shell
#SBATCH --job-name="TempProject/42b7b4f2921788ea14dac5566e6f06d0/foo/13ee8c7cb17a11b218fe41a3e31afab3"
#SBATCH --partition=gpu
#SBATCH --nodes=4
#SBATCH --ntasks-per-node=8
#SBATCH --gpus=8
```
Given the request that `nranks=4` and `ngpu=8`, this should be `--ntasks-per-node=2` as there are 2 GPUs per node for the GPU cluster of GreatLakes.

This problem may exist in other environments and was propagated to #561, so we should check other templates for this logical error.

joaander:
While #722 produces a proper resource request for sbatch, it fails to work correctly:
This request
```
#SBATCH --nodes=2
#SBATCH --ntasks=4
#SBATCH --gpus=4
...
# hard_disk_nvt_gpu(agg-88ab2b6305a8b249743764fc16e5c22e)
/home/joaander/miniforge3/bin/python /gpfs/accounts/sglotzer_root/sglotzer0/joaander/hoomd-validation/hoomd_validation/project.py run -o hard_disk_nvt_gpu -j agg-88ab2b6305a8b249743764fc16e5c22e
# Eligible to run:
# mpirun -n 4  /home/joaander/miniforge3/bin/python /gpfs/accounts/sglotzer_root/sglotzer0/joaander/hoomd-validation/hoomd_validation/project.py exec hard_disk_nvt_gpu agg-88ab2b6305a8b249743764fc16e5c22e
```
produces:
```
An ORTE daemon has unexpectedly failed after launch and before
communicating back to mpirun. This could be caused by a number
of factors, including an inability to create a connection back
to mpirun due to a lack of common network interfaces and/or no
route found between them. Please check network connectivity
(including firewalls and network routing requirements).
```

When I use this instead:
```
#SBATCH --ntasks=4                                                                                                                                                                                                                                                      
#SBATCH --gpus-per-task=1                                                                                                                                                                                                                                               
#SBATCH --mem-per-cpu="30g" 
```
[...]

PR #702, opened 2022-12-12
fix: Multi-node GPU summissions for greatlakes and picotte.

b-butler:
## Description
Fixes logic where the --ntasks-per-node would not normalize based on
number of nodes for GPU submissions where the number of tasks is often
the number of GPUs.

## Motivation and Context
Resolved: #566

## Checklist:


flow/templates/drexel-picotte.sh
```diff
@@ -12,7 +12,7 @@
     {% set nn = nn|default((nn_cpu, nn_gpu)|max, true) %}
     {% if partition == 'gpu' %}
 #SBATCH --nodes={{ nn|default(1, true) }}
-#SBATCH --ntasks-per-node={{ (gpu_tasks, cpu_tasks)|max }}
+#SBATCH --ntasks-per-node={{ ((gpu_tasks, cpu_tasks)|max / nn)|int }}
```
csadorf, line 15: This will only be correct if the number of nodes is evenly divisible by the number of tasks, right? Would this be an issue? If yes, should we protect against that?

csadorf (commented):
Looks good, just one quick question.

joaander:
I think there are more problems with this template. I ask for 4 ranks and 4 GPUs, but get 8 ranks spread across 4 GPUs:
```python
import flow

class Project(flow.FlowProject):
    pass

@Project.operation(directives=dict(nrank=4, ngpu=4))
def job(job):
    pass

if __name__ == "__main__":
    Project.get_project().main()
```

```shell
python project.py submit --pretend --partition gpu
Querying scheduler...
Submitting cluster job 'Project/99914b932bd37a50b983c5e7c90ae93b/job/86cd9023f6b91d9e9107a4de7a492fbe':                                                                           
 - Group: job(99914b932bd37a50b983c5e7c90ae93b)
WARNING:flow.project:Some of the keys provided as part of the directives were not used by the template script, including: nrank                                                   
# Submit command: sbatch
#!/bin/bash
#SBATCH --job-name="Project/99914b932bd37a50b983c5e7c90ae93b/job/86cd9023f6b91d9e9107a4de7a492fbe"                                                                                
#SBATCH --partition=gpu
#SBATCH --nodes=2
#SBATCH --ntasks-per-node=4
#SBATCH --gpus=4
```

I get the correct output (nodes=1, ntasks-per-node=N, gpus=N) for N=1, 2, I get incorrect output (nodes=2) for N=3,4.

I also tested N=8 and got:
```
#SBATCH --nodes=4
#SBATCH --ntasks-per-node=8
#SBATCH --gpus=8
```
!

Something is seriously wrong with the nodes calculation, not just the ntaks-per-node.

flow/templates/drexel-picotte.sh
joaander: If the tasks cannot be evenly distributed across an integer number of nodes (e.g. `nrank=13`), then there is no way to request this with `--ntasks-per-node` - if you rounded up you would be providing the user with more ranks than they requested. Use `--ntasks={nranks}` instead in these cases.

Preserve the `--nodes=`, `--ntasks-per-node=`, request when tasks can be evenly distributed across nodes. It will provide a more efficient communication pattern.

b-butler:
Thanks @joaander I will get working on this after the holidays.

flow/templates/drexel-picotte.sh
csadorf: Yes, so I think we should raise an error in case that the requested configuration cannot be provisioned.

b-butler: We currently round in other templates for GPU partitions e.g. `expanse`. We could change that. Generally that won't matter since we take the ceiling and charges for GPU nodes are usually just for GPUs I if I understand correctly.

joaander: Yes, they would likely be charged correctly. However, if the users launches their app with `srun` or `mpiexec` without arguments, then the autodetected number of tasks detected from the job configuration. If this is rounded up from what the user requested, then the user's script may fail (e.g. when it is coded to work with a specific number of ranks).

b-butler: Okay, in the case of GPU jobs does it make sense to have CPU tasks not be a multiple of GPUs? I feel that is something we should error at then.

joaander: It is possible for systems to support a different number of CPU tasks and GPUs. For example, NCSA Delta does:
```
$ srun --account=bbgw-delta-gpu --partition=gpuA40x4 --tasks=7 --mem=48g --gpus=5 --pty zsh
$ echo $SLURM_NTASKS                                                                                                       
7
$ echo $SLURM_TASKS_PER_NODE
4,3
```
In this test, it assigned all 4 GPUs on the first node and 1 GPU on the 2nd.

Just because it is possible doesn't need that signac-flow needs to support it. I can't think of any reasonable workflows that would need this. Also, you would need to check each system separately whether it has configured SLURM to allow for this uneven task distribution.

b-butler:
Given the shear complexities involved and problems in most templates given specific (corner case) resource requests. I am planning on moving much of this logic to Python in the environment classes. Much of this can be default, and just go into the base class so it doesn't offer much of a jump of complexity to users creating new environments.

joaander:
> Given the shear complexities involved and problems in most templates given specific (corner case) resource requests. I am planning on moving much of this logic to Python in the environment classes. Much of this can be default, and just go into the base class so it doesn't offer much of a jump of complexity to users creating new environments.

You used this approach on #708, right? That is very clear and appears easy to maintain.

bdice:
@b-butler Can we push this through or close it?

b-butler:
I believe #722 fixes this.

PR #709, opened 2023-01-21
Update pydocstyle.

bdice:
Small bump to pydocstyle version. Mirroring changes from https://github.com/glotzerlab/signac/pull/884.


b-butler (approved):
:+1:

PR #720, opened 2023-02-14
Make bundle_prefix consistent with bundle_id

### LINKED ISSUE #718: Bundling not working with FlowProject subclass
The FlowProject class is not using consistent logic for writing/reading bundle information, which causes problems when using certain subclasses of FlowProject. We should centralize the definition of a bundle prefix in a single place (probably a module constant) and use it as a string template for formatting so we don't have to keep these two code paths in sync.

This bug was identified by @bpuchala - thanks for reporting!

https://github.com/glotzerlab/signac-flow/blob/475b8e416edba5db1dab966335c0a5b8a1cfe108/flow/project.py#L2217

https://github.com/glotzerlab/signac-flow/blob/475b8e416edba5db1dab966335c0a5b8a1cfe108/flow/project.py#L2206

I'm not 100% sure how to test this or if it would require additional mocked logic in our testing framework. I'm inclined to centralize the definition and call that good enough. @bpuchala has confirmed that the proposed fix in https://github.com/bpuchala/signac-flow/commit/c7e8893d46f27b9ab59ea2a4c07208e4e3b7cf16 is sufficient to solve the problem.

bdice:
## Description
Fixes #718. Thanks @bpuchala for identifying the problem!

## Checklist:

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -21,6 +21,7 @@ Fixed
 +++++
 
 - Bumped ``isort`` to 5.12.0 in the pre-commit configuration file (#715).
+- Fixed bundle status information not being correctly read for subclasses of ``FlowProject`` (#718, #720).
 
 Removed
 +++++++
--- a/contributors.yaml
+++ b/contributors.yaml
@@ -171,4 +171,9 @@ contributors:
     given-names: Melody
     orcid: "https://orcid.org/0000-0001-9788-9958"
     affiliation: "University of Michigan"
+  -
+    family-names: Puchala
+    given-names: Brian
+    orcid: "https://orcid.org/0000-0002-2461-6614"
+    affiliation: "University of Michigan"
 ...
--- a/flow/project.py
+++ b/flow/project.py
@@ -2126,6 +2126,11 @@ class FlowProject(signac.contrib.Project, metaclass=_FlowProjectClass):
         """Return the canonical name to store bundle information."""
         return os.path.join(self.path, ".bundles", bundle_id)
 
+    @property
+    def _bundle_prefix(self):
+        sep = getattr(self._environment, "JOB_ID_SEPARATOR", "/")
+        return f"{self.__class__.__name__}{sep}bundle{sep}"
+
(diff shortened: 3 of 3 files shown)


bdice:
@bpuchala If you'd like, I'm happy to add your name to the [list of contributors](https://github.com/glotzerlab/signac-flow/blob/master/contributors.yaml). Let me know either way.

bpuchala:
@bdice Sure, thanks!

syjlee (approved):

vyasr (approved):
I am fine with just centralizing for now without adding new tests.

Out of curiosity, in what scenario was `self.__class__.__name__` giving a different result from `self`?

bdice:
> Out of curiosity, in what scenario was `self.__class__.__name__` giving a different result from `self`?

Not exactly sure, @bpuchala was using subclasses of FlowProject in some way.

bpuchala:
It was giving "None" instead of "FlowProject" when running from a derived class. Here's an example of the issue, when run in a directory with an existing signac project:

Example:
```
import signac
import flow

# Add to flow.FlowProject:
#
# def _get_bundle_prefix(self):
#         sep = getattr(self._environment, "JOB_ID_SEPARATOR", "/")
#         return f"{self}{sep}bundle{sep}"

class MyFlowProject(flow.FlowProject):
    def __init__(self):
        signac_project = signac.init_project()
        super().__init__(signac_project.config)
        pass

my_proj = MyFlowProject()
print(my_proj._get_bundle_prefix())
```

Output:
```
$ python demo.py 
None/bundle/
```

bdice:
Aha. This is pulling the signac.Project name from a defaulted config (we made project names optional in signac 1.8, and will be removing them in signac 2.0).

bdice:
@bpuchala I think the more canonical way to do this is to leverage the fact that MyFlowProject inherits from FlowProject, which inherits from signac’s Project class. You can call MyFlowProject.init_project() and get the same method rather than feeding the config through manually.

PR #722, opened 2023-02-27
Refactor submission templates

b-butler:
## Description
This PR moves much of the task and number of nodes logic to the environment classes where they can specify any needed parameters using the `resources` template context value now provided to the Jinja templates. This allows more complicated logic in a simpler to understand format compared to Jinja. This also promotes similar computation of resources, and explicit overrides through inheritance. For some more discussion see #702.

## Motivation and Context
This is attempting to solve some bugs in the existing template logic. For one, we sometimes fail on multi-nodes GPU submissions (see #702) and related issues. We also currently modify user resource requests (e.g. rounding CPU tasks to the same number per node) rather than submitting as is or erroring.

## Checklist:


b-butler:
TODO:
- [ ] Document `_get_scheduler_resources` more fully
- [ ] Get feedback on new choices on default settings (e.g. `--ntasks` over `--ntasks-per-node`)
- [ ] Port over all templates
- [ ] Test all templates
- [ ] Regenerate template test reference and validate

*Anyone working on this please feel free to modify this.*

bdice:
@b-butler Can we push this through or close it?

b-butler:
Only have Frontier to convert over and test now.

flow/project.py
```diff
@@ -663,6 +663,7 @@ def __init__(
         self._project = project
         self.submit_options = submit_options
         self.run_options = run_options
+        # TODO: This is no longer true.
```
tommy-waltmann, line 666: Should the comment below be updated?

flow/project.py
```diff
@@ -1534,6 +1535,7 @@ def _internal_call(
 
                 # Append the name and function to the class registry
                 self._parent_class._OPERATION_FUNCTIONS.append((name, func))
+                # TODO: This is no longer true and can likely be changed.
```
tommy-waltmann, line 1538: Should the comment below be changed?

tommy-waltmann (commented):
I have two small suggestions and a question:

This PR touches a lot of cluster templates. Does flow normally need to be validated on each of the clusters in addition to unit tests? If so, has the validation been done after the changes in this PR?

b-butler:
>This PR touches a lot of cluster templates. Does flow normally need to be validated on each of the clusters in addition to unit tests? If so, has the validation been done after the changes in this PR?

@tommy-waltmann that is part of the reason the PR took so long. All of the templates changed have been validated, I believe (need to double check Andes). That is also the reason some are not changed since I could not access them.

tommy-waltmann (approved):

flow/templates/andes.sh
```diff
@@ -1,37 +1,24 @@
...
     {% if 'gpu' in partition %}
-        {% set gpus_per_node = (gpu_tasks / nn)|round(0, 'ceil')|int %}
-        {% set cpus_per_node = (cpu_tasks / nn)|round(0, 'ceil')|int %}
-        {% if cpus_per_node > gpus_per_node * 14 and not force %}
+        {% if resources.ncpu_tasks > resources.ngpu_tasks * 14 and not force %}
             {% raise "Cannot request more than 14 CPUs per GPU." %}
         {% endif %}
     {% endif %}
+#SBATCH -N {{ resources.num_nodes }}
+#SBATCH --ntasks={{ resources.ncpus_tasks }}
```
b-butler, line 19:
```suggestion
#SBATCH --ntasks={{ resources.ncpu_tasks }}
```

PR #725, opened 2023-03-02
Feature/469 check status for specific operations

rayasare:
## Description
We removed the `-o` flag from `_add_print_status_args` and added the `-o` and `--operation` flags to the status subparser. 

## Motivation and Context
There was previously no way to check status for only specific operations/groups.

## Checklist:


rayasare:
We're not sure where to add the tests for these updates -- should we add the new operation flag to [generate_status_reference_data.py](https://github.com/glotzerlab/signac-flow/blob/master/tests/generate_status_reference_data.py) or add some additional logic to [test_project.py](https://github.com/glotzerlab/signac-flow/blob/master/tests/test_project.py)?

flow/project.py
```diff
@@ -2616,6 +2617,10 @@ def _fetch_status(
         status_parallelization : str
             Parallelization mode for fetching the status. Allowed values are
             "thread", "process", or "none". (Default value = "none")
+        names : iterable of :class:`str`
+            Only execute operations that match the provided set of names
```
bdice, line 2621:
```suggestion
            Only show status for operations that match the provided set of names
```

flow/project.py
```diff
@@ -2649,9 +2662,9 @@ def _fetch_status(
 
         def compute_status(data):
             scheduler_id, scheduler_status, aggregate_id, aggregate, group = data
-
             status = {}
             error_text = None
+            # if (operation is None) or (operation )
```
bdice, line 2667: What does this comment mean?

flow/project.py
```diff
@@ -4994,6 +5016,14 @@ class MyProject(FlowProject):
             "to show result for. Defaults to the main module. "
             "(requires pprofile)",
         )
+        parser_status.add_argument(
+            "-o",
+            "--operation",
+            type=str,
+            nargs="+",
+            help="Only select operation or groups that match the given "
```
bdice, line 5024:
```suggestion
            help="Select operation or groups that match the given "
```

flow/project.py
```diff
@@ -2859,6 +2876,10 @@ def print_status(
         output_format : str
             Status output format, supports:
             'terminal' (default), 'markdown' or 'html'.
+        operation : iterable of :class:`str`
+            Only execute operations that match the provided set of names
```
bdice, line 2880:
```suggestion
            Show status of operations that match the provided set of names
```

flow/project.py
```diff
@@ -2704,11 +2717,14 @@ def compute_status(data):
                     )
             return result
 
+        submission_groups = set(self._gather_flow_groups(names))
```
bdice, line 2720: Are these submission groups? Do you mean something else?

bdice (changes requested):
Looks like this is a draft - I left a couple comments for now. For testing, you should test the Python logic in `test_project.py` and also add tests for the status command line in `generate_status_reference_data.py`. I'm not sure why `status_reference_data.tar.gz` changed, but you probably don't want to edit that file manually -- please update the generator script, then run it and commit any changes to the generator scripts along with the updated tarball.

flow/project.py
rayasare: I forgot this was here -- deleting it.

flow/project.py
```diff
@@ -4994,6 +5015,14 @@ class MyProject(FlowProject):
             "to show result for. Defaults to the main module. "
             "(requires pprofile)",
         )
+        parser_status.add_argument(
```
bdice, line 5018: Why does this implement a new option and not use `self._add_operation_selection_arg_group(parser_status)`?

flow/project.py
```diff
@@ -2859,6 +2875,10 @@ def print_status(
         output_format : str
             Status output format, supports:
             'terminal' (default), 'markdown' or 'html'.
+        operation : iterable of :class:`str`
```
bdice, line 2878: In `run` and `submit`, this is called "names". Let's match the convention.

bdice (changes requested):

tommy-waltmann (approved):
Once Bradley's comments are addressed, this should be in good shape.

flow/project.py
rayasare: @bdice any other thoughts?

tests/test_project.py
```diff
@@ -122,6 +123,72 @@ def test_status_performance(self):
...
+            "group1_op3",
+            "group2_op1",
+            "op1_op3",
+        ],
+    )
+    def test_groups(self, groups, get_status):
+        stdout, _ = get_status({"operation": groups})
+        excluded_groups = {"group1", "group2", "op1", "op2", "op3"} - set(groups)
+        if len(excluded_groups) > 0:
+            # This needs to be modified if we test most than two groups
```
bdice, line 180:
```suggestion
            # This needs to be modified if we test more than two groups
```

tests/test_project.py
```diff
@@ -122,6 +123,72 @@ def test_status_performance(self):
...
+            for excluded_group in list(excluded_groups):
+                assert excluded_group not in "".join(stdout)
```
bdice, line 182: Define `operations_output` above, then use it here:
```suggestion
                assert excluded_group not in operations_output
```

tests/test_project.py
```diff
@@ -122,6 +123,72 @@ def test_status_performance(self):
...
+            "group2_op1",
+            "op1_op3",
+        ],
+    )
+    def test_groups(self, groups, get_status):
+        stdout, _ = get_status({"operation": groups})
+        excluded_groups = {"group1", "group2", "op1", "op2", "op3"} - set(groups)
+        if len(excluded_groups) > 0:
+            # This needs to be modified if we test most than two groups
+            for excluded_group in list(excluded_groups):
```
bdice, line 181: You should be able to iterate over the set:
```suggestion
            for excluded_group in excluded_groups:
```

tests/test_project.py
bdice: Can we write this test in a way that is resilient to the input structure? i.e. a test that doesn't have this limitation.

rayasare:
Originally we changed instances of `operation` to `names` to match the convention in `run` and `submit`, but had to revert because the status parser ([line 5004](https://github.com/glotzerlab/signac-flow/blob/main/flow/project.py)) adds the argument flag flag `--operation`, not `--names`. (The `-n` flag is already taken if we wanted to change the convention in the status parser.)

tests/generate_status_reference_data.py
```diff
@@ -46,8 +46,9 @@ def init_status_options(project):
         {"compact": True},  # -1, --one-line
         {"pretty": True},  # --pretty
         {"eligible_jobs_max_lines": 2},  # --eligible-jobs-max-lines 2
-        {"output_format": "markdown"},  # -o markdown, --output-format markdown
-        {"output_format": "html"},  # -o html, --output-format html
+        {"output_format": "markdown"},  # --output-format markdown
+        {"output_format": "html"},  # --output-format html
+        {"operation": ["op1"]},  # -o op1, --operation op1
```
bdice, line 51: Can we add a test for multiple operations?

tests/test_project.py
```diff
@@ -122,6 +123,70 @@ def test_status_performance(self):
...
+
+        return project
+
+    @pytest.fixture(scope="function")
+    def project(self):
+        return self.mock_project()
+
+    @pytest.fixture(scope="function")
+    def get_status(self, project):
+        def _get_status(kwargs):
```
bdice, line 147: Should this function accept `**kwargs` instead? Then we could refactor the call sites like `get_status(operation=groups)` instead of `get_status({"operation": groups})`. I think this is a good idea because it makes `get_status` have a similar signature to `project.print_status`.
```suggestion
        def _get_status(**kwargs):
```

tests/test_project.py
```diff
@@ -122,6 +123,70 @@ def test_status_performance(self):
...
+            "group2",
+            "group1_2",
+            "op1",
+            "group1_op3",
+            "group2_op1",
+            "op1_op3",
+        ],
+    )
+    def test_groups(self, groups, get_status):
+        stdout, _ = get_status({"operation": groups})
```
bdice, line 177: If we aren't using the `stderr` capture, why do we need to capture and return it in `get_status`?

bdice (changes requested):
A few suggestions for improvement. I fixed up the changelog.

bdice (approved):
Thanks for your persistence on this @rayasare. I think it's in good shape!

PR #775, opened 2023-10-27
Support Purdue Anvil.

joaander:
## Description

Add environment for Purdue Anvil.

## Motivation and Context

Anvil is a national HPC resources allocated though the NSF ACCESS program. Allow users to submit jobs to Anvil using flow.

## Checklist:

- [x] The [package documentation](https://github.com/glotzerlab/signac-flow/tree/main/doc) and [framework documentation](https://docs.signac.io/) in [signac-docs](https://github.com/glotzerlab/signac-docs) are up to date with these changes.
- [x] I have updated the [changelog](https://github.com/glotzerlab/signac-flow/blob/main/changelog.txt) and added any related issue and pull request numbers for future reference.


joaander:
I have successfully tested the partitions `debug`, `shared` (default) and `wholenode` in production scripts (hoomd-validation).

The `test-workflow` jobs produce results that appear correct for `wide`, `highmem`, `gpu-debug`, and `gpu`. I don't have a GPU allocation to use for testing GPU submissions.

I'm not sure about the hybrid OMP/MPI jobs generated by `generate_template_reference_data`. Without a 'cpus-per-task', SLURM may not correctly assign tasks and threads as the user intends. However, this is a general problem of the SLURM template and therefore fixing it is beyond the scope of this pull request.

flow/templates/anvil.sh
```diff
@@ -0,0 +1,33 @@
...
+            {% raise "Cannot request a gpu job with more than 4 GPUs." %}
+        {% endif %}
+    {% else %}
+        {% if 'gpu' in partition and not force %}
+            {% raise "Requesting gpu partition, but no GPUs requested!" %}
+        {% endif %}
+    {% endif %}
+    {% if partition in ['debug', 'shared', 'highmem'] and resources.ncpu_tasks > 128 %}
+        {% raise "Cannot request a shared partition with more than 128 CPU tasks." %}
+    {% endif %}
```
b-butler, line 18: This should be unnecessary given `env._shared_partitions`. The reason I didn't remove it from other templates is either some clusters only allow shared submissions of one socket not the entire node and I may have forgotten :sweat_smile: , but in this case this is more than one node so you should be good.
```suggestion
```

flow/templates/anvil.sh
```diff
@@ -0,0 +1,33 @@
+{# Templated in accordance with: https://www.rcac.purdue.edu/knowledge/anvil/ #}
+{% extends "slurm.sh" %}
+{% block tasks %}
+    {% if resources.ngpu_tasks %}
+        {% if not ('gpu' in partition or force) %}
+            {% raise "GPU operations require a gpu partition!" %}
+        {% endif %}
+        {% if resources.ngpu_tasks > 4 %}
```
b-butler, line 8: Is this true, they don't list a GPU node limit for `gpu` in https://www.rcac.purdue.edu/knowledge/anvil/run/partitions.

flow/templates/anvil.sh
```diff
@@ -0,0 +1,33 @@
...
+        {% endif %}
+    {% else %}
+        {% if 'gpu' in partition and not force %}
+            {% raise "Requesting gpu partition, but no GPUs requested!" %}
+        {% endif %}
+    {% endif %}
+    {% if partition in ['debug', 'shared', 'highmem'] and resources.ncpu_tasks > 128 %}
+        {% raise "Cannot request a shared partition with more than 128 CPU tasks." %}
+    {% endif %}
+    {% if resources.num_nodes > 1 or resources.ncpu_tasks >= 128 or resources.ngpu_tasks >= 4 %}
```
b-butler, line 19: This should be sufficient right?
```suggestion
    {% if resources.num_nodes > 1 %}
```

b-butler (changes requested):
Thanks for the environment @joaander!

flow/templates/anvil.sh
joaander: I don't know what you mean about `env._shared_partitions`. If I remove these lines, flow generates a 2 node shared partition script without error or warning:
```
#SBATCH --partition=shared
#SBATCH -N 2
#SBATCH --ntasks=256
#SBATCH -A dmr140129
```
Is your intention that the user receive this particular error from `sbatch` instead?

flow/templates/anvil.sh
joaander: Yes, this is true:
```
salloc -Admr140129 --nodes=2 --ntasks-per-node=4 --gpus-per-node=4  --partition shared -t 1:00:00
salloc: error: QOSMaxNodePerJobLimit
```
Would you prefer that we remove this error from flow and allow `sbatch` to notify the user?

flow/templates/anvil.sh
joaander: Probably. The line copies the pattern set by `bridges2.sh`.

flow/templates/anvil.sh
joaander: Although, they do provide a multi-node example script: https://www.rcac.purdue.edu/knowledge/anvil/run/examples/slurm/gpu#how_to_use_slurm_to_submit_a_multi-node_gpu_program

changelog.txt
```diff
@@ -11,6 +11,11 @@ Version 0.27
 [0.27.0] -- 202x-xx-xx
 ----------------------
 
+Added
```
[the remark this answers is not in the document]
joaander: Moved to #776.

flow/templates/anvil.sh
joaander: `-N` is not necessary. `sbatch` can compute it from `--ntasks`.

flow/templates/anvil.sh
joaander: I removed the checks so that `sbatch` provides the user with errors on invalid resource requests for a given partition.

b-butler (approved):
I think I misspoke in my previous review, so I went ahead and made some small changes to the template. Please check that they are correct, and if so feel free to (auto-)merge.

joaander:
I didn't misinterpret your earlier comments, I took them to the extreme and removed unneeded code. `sbatch` will error if the user requests GPUs in a non-gpu partition, just as it will error if a user requests more than the maximum number of cores in a specific partition (which we don't check for in the template).

Similarly, the `-N` is not needed and/or could be set unconditionally. However, it is not an error to include. These will likely be revised #777. I also noticed that other elements of the slurm scripts could be refactored - like moving the account to the base slurm script, or at least an intermediate common template.

Issue #777, opened 2023-10-30
Incorrect SLURM scripts produced with `omp_num_threads>1`.


joaander:
<!-- Please replace the text in the individual sections below. -->

### Description

flow incorrectly computes `np = nranks * omp_num_threads` and then sets `--ntasks=np` in the SLURM submission script.

### To reproduce

```
$ flow test-workflow -g 4 -c 128 -n 1
```

Then add:
```
@Project.post.true("hybrid_mpi_omp")                                                                                                                                                                                                                                    
@Project.operation(directives={"nranks": 4, "omp_num_threads": 127})                                                                                                                                                                                                    
    def hybrid_mpi_omp(job):  
        pass
```
to `project.py`

```
$ python project.py submit --pretend -o hybrid_mpi_omp --partition wholenode
```


### Error output

The resulting job script is:
```
#SBATCH --job-name="Project/b49788b9076312db79994d96758fdcf8/hybrid_mpi_o/6bb6c79496f9cc4858a0c6dd166318b6"
#SBATCH --partition=wholenode
#SBATCH --ntasks=508
...
export OMP_NUM_THREADS=127; mpirun -n 4 ...
```

A job definition like this does not provide the desired behavior. Tested interactively:
```
salloc -Admr140129 --ntasks=508 --partition wholenode
...
a937:~/devel/signac-flow ▶ export OMP_NUM_THREADS=127; mpirun -n 4 hostname
a937.anvil.rcac.purdue.edu
a937.anvil.rcac.purdue.edu
a937.anvil.rcac.purdue.edu
a937.anvil.rcac.purdue.edu
```
Notice how all ranks are run on the same node. Each rank should be on a separate node.

### Correct job definition:

```
salloc -Admr140129 --ntasks=4  --cpus-per-task=127 --partition wholenode -t 1:00:00
...
a937:~/devel/signac-flow ▶ mpirun -n 4 hostname
a937.anvil.rcac.purdue.edu
a965.anvil.rcac.purdue.edu
a966.anvil.rcac.purdue.edu
a971.anvil.rcac.purdue.edu
```
When ntasks is the number of MPI tasks and `--cpus-per-task` is set appropriately, the 4 MPI ranks are assigned to different nodes.

### System configuration

Please complete the following information:

 - Operating System [e.g. macOS]: Linux
 - Version of Python [e.g. 3.7]: 3.11
 - Version of signac [e.g. 1.0]: 2.1.0
[...]

PR #779, opened 2023-10-31
Fix srun invocation on Great Lakes.

joaander:
## Description

Switch back to `mpirun` on Great Lakes.

## Motivation and Context

This change was made in #722. My workflows (`hoomd-validation`) work correctly with the most current release of flow, but not the `main` branch. When `flow` forks to `srun`, somehow my environment variables are all lost. For example, one of errors I got testing this was `singularity not found` even though I checked in the shell and singularity was in the PATH and the filesystem where singularity was stored was accessible.

Strangely, everything works if I directly called the srun command from the shell:
```
srun -n 4  /home/joaander/miniforge3/bin/python /gpfs/accounts/sglotzer_root/sglotzer0/joaander/hoomd-validation/hoomd_validation/project.py exec hard_disk_nvt_gpu agg-88ab2b6305a8b249743764fc16e5c22e
```

I don't know what causes this behavior, but the expedient solution is to switch back to `mpirun` which we have used for a long time on Great Lakes with fewer issues.

## Checklist:

- [ ] The [package documentation](https://github.com/glotzerlab/signac-flow/tree/main/doc) and [framework documentation](https://docs.signac.io/) in [signac-docs](https://github.com/glotzerlab/signac-docs) are up to date with these changes.
- [ ] I have updated the [changelog](https://github.com/glotzerlab/signac-flow/blob/main/changelog.txt) and added any related issue and pull request numbers for future reference.


joaander:
Wait, this needs more testing on GPU nodes where srun configures the appropriate cgroups and mpirun does not. I may need to find a fix for the environment variable issue.

joaander:
Upon further testing with a 4 V100 GPU job (which slurm assigned 3 GPUs to 1 node and 1 GPU to another node), `mpirun` assigns multiple tasks to the same GPU while `srun` correctly distributes the tasks. Note: we still need `--gpus-per-task` which I will add separately when working on #777.

I found that `srun --export=ALL` solved the environment variable forwarding issue. I also found that `srun` buffered stdout/stderr until the job completed, so I also added `-u` to disable buffering.

`srun -u --export=ALL` correctly runs `hoomd-validation` jobs both when hoomd is compiled from source and from the glotzerlab-software container. I am waiting for one more test on the 4 GPU A100 node.

joaander:
This works on both V100 and A100 nodes on Great Lakes.

joaander:
@b-butler This PR now includes both fixes to the Great Lakes environment necessary to make the template functional for the next release. 
* Use cpus-per-task to ensure all processes requested via np are placed on one node.
* Use gpus-per-task to allow multi-GPU jobs to correctly distribute to the assigned GPUs.

This formulation works only with homogenous jobs. How do I call `homogeneous_openmp_mpi_config` to ensure this?

I did not fix `--cpus-per-task` for all environments, as we are planning a more general fix in #777.

b-butler (approved):
I made some changes to the implementation (moving to the template) since we are refactoring and `nranks` will be removed. The code is equivalent.

PR #784, opened 2023-11-02
Move environment configuration to its own class

b-butler:
## Description
Adds three classes
- `_PartitionConfig`: an environment's partition configuration
- `_Partition`: a specific partition's data
- `_NodeTypes`: what kind of nodes a partition supports
  - SHARED: 1 or fewer nodes
  - MIXED: any number of nodes/processes
  - WHOLENODE: submit in whole nodes only

## Motivation and Context
With recent changes to flow's template/submission logic (see #722), the environment classes were becoming loaded with class variables. This makes a logical change to storing the partition specific information in appropriate classes. This also removes some methods from the environments in favor of putting them in the partition classes.

## Checklist:
- [x] The [package documentation](https://github.com/glotzerlab/signac-flow/tree/main/doc) and [framework documentation](https://docs.signac.io/) in [signac-docs](https://github.com/glotzerlab/signac-docs) are up to date with these changes.
- [ ] I have updated the [changelog](https://github.com/glotzerlab/signac-flow/blob/main/changelog.txt) and added any related issue and pull request numbers for future reference.


b-butler:
Tested appropriate errors for GPU/non-GPU and node requests on Bridges2.

b-butler:
@cbkerr when you have a moment could you review?

b-butler:
@cbkerr will you have time to review or should I assign someone else?

cbkerr:
@b-butler Thanks for the ping. I can get to this during the next few days.

flow/environment.py
```diff
@@ -90,6 +91,169 @@ def template_filter(func):
...
+    - CPUs for a partition
+    - GPUs for a partition
+    - Node type of a partition
+
+    When querying a value for a specific partition the logic first searches the
+    provided mapping if any for the partition, if it is not found then the
+    mapping is searched for "default" if it exists, if not the class default is
+    used.
+
+    1. Partition specific -> 2. Provided default -> 3. _PartitionConfig default
```
cbkerr, line 122: What do the numbers represent? If search order, put them in a list instead

flow/environments/umich.py
```diff
@@ -13,10 +13,10 @@ class GreatLakesEnvironment(DefaultSlurmEnvironment):
 
     hostname_pattern = r"gl(-login)?[0-9]+\.arc-ts\.umich\.edu"
     template = "umich-greatlakes.sh"
-    _cpus_per_node = {"default": 36, "gpu": 40}
-    _gpus_per_node = {"default": 2}
-    _shared_partitions = {"standard", "gpu"}
```
cbkerr, line 18: What happened to the standard partition?

flow/environment.py
```diff
@@ -90,6 +91,169 @@ def template_filter(func):
...
+
+    - CPUs: ``None`` which represents any number. This will be interpretted as
+      an unlimited supply of CPUs practically.
+    - GPUs: 0.
+    - Node type: `_NodeTypes.MIXED`.
+
+    Parameters
+    ----------
+    cpus_per_node: dict[str, int], optional
+        Mapping between partitions and CPUs per node. Defaults to an empty `dict`.
```
cbkerr, line 134: I had to look at the environments to understand this docstring. Need to mention that keys in this dictionary are partition names.

I understand why "mapping" was used, but I couldn't figure out which direction it was pointing and how to populate the dictionary based on this info.

```suggestion
        A `dict` mapping from partition names (keys) to CPUs per node (values) on those partitions.
        Defaults to an empty `dict`.
```

Also applies to the other docstrings but we can discuss here first

b-butler: A mapping from a to b means a are keys and b are values.

flow/environments/umich.py
b-butler: The standard partition as well as the gpu partition don't need specifying since their behavior matches the default.

cbkerr (approved):
Concerning documentation:

Are there any user-facing changes to the template and environment documentation we would need to add?

What are the risks of default values changing and messing up inherited values?

b-butler:
@cbkerr Nope, this is purely a developer facing thing currently.
And the way the model works is that the defaults are inherited as class members from `_ComputeEnvironment` but you can't partially inherit meaning you must explicitly set the full specification for every new class.

PR #811, opened 2024-02-05
Add gpu_mig40 to Greatlakes.

joaander:
## Description

Add the `gpu_mig40` partition to UMich Great Lakes.

## Motivation and Context

Allow users to submit to this new GPU partition.

## Checklist:

- [ ] The [package documentation](https://github.com/glotzerlab/signac-flow/tree/main/doc) and [framework documentation](https://docs.signac.io/) in [signac-docs](https://github.com/glotzerlab/signac-docs) are up to date with these changes.
- [x] I have updated the [changelog](https://github.com/glotzerlab/signac-flow/blob/main/changelog.txt) and added any related issue and pull request numbers for future reference.


cbkerr (approved):
Thanks! Wondering how we would support submitting to both mig40 and gpu with fallback

joaander:
> Thanks! Wondering how we would support submitting to both mig40 and gpu with fallback

The two partitions have differing numbers of GPUs per node which flow uses to calculate the node occupancy. I'm also not certain how SLURM handles multiple GPU jobs with the two partitions - does it assign all to the same partition, or some ranks to one and some to another? Multiple partitions would cause serious problems for MPI jobs and load balancing issues for aggregates.

We could potentially enforce that the "gpu_mig40,gpu" partition is valid only for single GPU jobs. Feel free to implement this and test in a future PR. This might be best done after the ongoing work to refactor directives.

